PR #8, opened 2020-09-25
bugfix: fix prediction service and unit tests

dizcology:
Related issues:
https://github.com/googleapis/gapic-generator-python/issues/413
https://github.com/googleapis/gapic-generator-python/issues/414

### DIFF
--- a/google/cloud/aiplatform_v1beta1/services/prediction_service/client.py
+++ b/google/cloud/aiplatform_v1beta1/services/prediction_service/client.py
@@ -214,7 +214,7 @@ class PredictionServiceClient(metaclass=PredictionServiceClientMeta):
         if endpoint is not None:
             request.endpoint = endpoint
         if instances is not None:
-            request.instances = instances
+            request.instances.extend(instances)
         if parameters is not None:
             request.parameters = parameters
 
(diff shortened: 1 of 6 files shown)


google/cloud/aiplatform_v1beta1/services/prediction_service/client.py
```diff
@@ -328,7 +328,7 @@ def explain(
         if endpoint is not None:
             request.endpoint = endpoint
         if instances is not None:
-            request.instances = instances
```
busunkim96, line 331: Synth will try to revert changes like this in future Autosynth PRs. You can add this entire file to the `exclude` list, or a write a regex replacement in `synth.py`.

busunkim96 (approved):

google/cloud/aiplatform_v1beta1/services/prediction_service/client.py
dizcology: I had this in `synth.py` but failed to update the relevant line to v1beta1, it has been fixed: https://github.com/googleapis/python-aiplatform/pull/8/files#diff-3620890cb174a7f98c0722895e37bb79R72

software-dov: I can also reprioritize the issues that cause the need for synth hacks. The repeated field one should be trivial.

software-dov: Oh, my bad, it looks like that issue has actually been fixed in a more current generator release.

PR #11, opened 2020-09-25
feat: Migrate sample and sample tests - Batch 1 of 5

### LINKED ISSUE #9: Add samples and sample tests
Please add relevant code samples and tests.

vinnysenthil:

Fixes #9 (Batch 1 of 5) 🦕


samples/tests/cancel_training_pipeline_test.py
```diff
@@ -0,0 +1,78 @@
...
+def training_pipeline_id(capsys):
+    create_training_pipeline_sample.create_training_pipeline_sample(
+        project=PROJECT_ID,
+        display_name=DISPLAY_NAME,
+        training_task_definition=TRAINING_DEFINITION_GCS_PATH,
+        dataset_id=DATASET_ID,
+        model_display_name=f"Temp Model for {DISPLAY_NAME}",
+    )
+
+    out, _ = capsys.readouterr()
```
tmatsuo, line 44: Can we change `create_training_pipeline_sample()` to return what we need (in this case, it can return the pipeline_id)?

samples/tests/create_batch_prediction_job_video_classification_test.py
```diff
@@ -0,0 +1,75 @@
...
+        display_name=DISPLAY_NAME,
+        model_name=model_name,
+        gcs_source_uri=GCS_SOURCE_URI,
+        gcs_destination_output_uri_prefix=GCS_OUTPUT_URI,
+    )
+
+    out, _ = capsys.readouterr()
+
+    # Save resource name of the newly created batch prediction job
+    shared_state["batch_prediction_job_name"] = out.split("name:")[1].split("\n")[0]
```
tmatsuo, line 75: In general, please be little bit more careful when you slicing lists.

For example,
`out.split("name:")[1]`

This code will crash with an ambiguous error when there's no "name:" in the output.

samples/tests/create_batch_prediction_job_video_classification_test.py
```diff
@@ -0,0 +1,75 @@
...
+MODEL_ID = "667940119734386688"  # Permanent 5 class sports model
+DISPLAY_NAME = f"temp_create_batch_prediction_vcn_test_{uuid4()}"
+GCS_SOURCE_URI = (
+    "gs://ucaip-samples-test-output/inputs/vcn_40_batch_prediction_input.jsonl"
+)
+GCS_OUTPUT_URI = "gs://ucaip-samples-test-output/"
+
+
+@pytest.fixture
+def shared_state():
```
tmatsuo, line 37: Why do you have two separate fixture to achieve this? I'd only have one fixture. You'll encounter weird situation when those fixtures have different scope.

Maybe:

```python
@pytest.fixture(scope="function")
def deleter()
    state = {}
    yield state
    # any cleanup tasks

def test_something(capsys, deleter):
    # create something
    deleter["batch_prediction_job_name"] = out.split("name:")[1].split("\n")[0]

```

samples/tests/create_dataset_video_test.py
```diff
@@ -0,0 +1,55 @@
...
+)
+
+
+@pytest.fixture
+def shared_state():
+    state = {}
+    yield state
+
+
+@pytest.fixture(scope="function", autouse=True)
```
tmatsuo, line 36: If you are defining the same fixture in different files, consider using `conftest.py` where you can define common fixture which are automatically usable in your tests.

tmatsuo (changes requested):

samples/tests/create_dataset_video_test.py
vinnysenthil: Great point - I've captured it as issue https://github.com/googleapis/python-aiplatform/issues/36 and self-assigned this as a todo.

samples/tests/create_batch_prediction_job_video_classification_test.py
vinnysenthil: This logic has been moved to a helper method that uses regex, now if there is no "name:" in the output, the teardown will throw an error that `shared_state["batch_prediction_job_name"]` is not set.

samples/cancel_training_pipeline_sample.py
```diff
@@ -0,0 +1,33 @@
...
+    client_options = {"api_endpoint": "us-central1-aiplatform.googleapis.com"}
+    # Initialize client that will be used to create and send requests.
+    # This client only needs to be created once, and can be reused for multiple requests.
+    client = aiplatform.PipelineServiceClient(client_options=client_options)
+    location = "us-central1"
+    name = client.training_pipeline_path(
+        project=project, location=location, training_pipeline=training_pipeline_id
+    )
+    client.cancel_training_pipeline(name=name)
+    time.sleep(90)
```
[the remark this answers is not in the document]
vinnysenthil: Thanks for catching this, removed.

samples/tests/cancel_training_pipeline_test.py
vinnysenthil: cc @dizcology 

Currently all these samples are generated and we've opted to print the response, however allowing return values could be added at the generator level in the future. All our tests would need to be re-written to accommodate this change.

samples/create_dataset_image_sample.py
```diff
@@ -0,0 +1,38 @@
...
+    parent = "projects/{project}/locations/{location}".format(
+        project=project, location=location
+    )
+    dataset = {
+        "display_name": display_name,
+        "metadata_schema_uri": "gs://google-cloud-aiplatform/schema/dataset/metadata/image_1.0.0.yaml",
+    }
+    response = client.create_dataset(parent=parent, dataset=dataset)
+    print("Long running operation:", response.operation.name)
+    create_dataset_response = response.result(timeout=300)
```
tmatsuo, line 34: Re: timeout for long running operations

Can we make it a parameter to the sample function as follows?

```python
def create_dataset_image_sample(display_name: str, project: str, timeout: int = 300):
    # ...
    create_dataset_response = response.result(timeout=timeout)
```

This way, you can bump the timeout only by changing the test files (no need to update the devsite).

samples/tests/cancel_training_pipeline_test.py
```diff
@@ -0,0 +1,81 @@
...
+
+    training_pipeline_name = helpers.get_name(out)
+
+    training_pipeline_id = training_pipeline_name.split("/")[-1]
```
tmatsuo, line 48: Nit: I would assert to make sure there is a "/" in the training_pipeline_name.

What I'm worrying is that a situation where the test is failing, but we're not sure why.

samples/tests/create_batch_prediction_job_video_classification_test.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+@pytest.fixture(scope="function")
+def shared_state():
+
+    shared_state = {}
+
+    yield shared_state
+
+    batch_prediction_job = shared_state["batch_prediction_job_name"].split("/")[-1]
```
tmatsuo, line 44: ditto

samples/tests/create_batch_prediction_job_video_object_tracking_test.py
```diff
@@ -0,0 +1,76 @@
...
+def shared_state():
+    state = {}
+    yield state
+
+
+@pytest.fixture(scope="function", autouse=True)
+def teardown(shared_state):
+    yield
+
+    batch_prediction_job = shared_state["batch_prediction_job_name"].split("/")[-1]
```
tmatsuo, line 47: ditto

samples/tests/create_data_labeling_job_images_test.py
```diff
@@ -0,0 +1,94 @@
...
+def shared_state():
+    state = {}
+    yield state
+
+
+@pytest.fixture(scope="function", autouse=True)
+def teardown(capsys, shared_state):
+    yield
+
+    data_labeling_job_id = shared_state["data_labeling_job_name"].split("/")[-1]
```
tmatsuo, line 49: ditto

samples/tests/create_data_labeling_job_test.py
```diff
@@ -0,0 +1,96 @@
...
+def shared_state():
+    state = {}
+    yield state
+
+
+@pytest.fixture(scope="function", autouse=True)
+def teardown(capsys, shared_state):
+    yield
+
+    data_labeling_job_id = shared_state["data_labeling_job_name"].split("/")[-1]
```
tmatsuo, line 50: ditto

samples/tests/create_data_labeling_job_video_test.py
```diff
@@ -0,0 +1,91 @@
...
+def shared_state():
+    state = {}
+    yield state
+
+
+@pytest.fixture(scope="function", autouse=True)
+def teardown(shared_state):
+    yield
+
+    data_labeling_job_id = shared_state["data_labeling_job_name"].split("/")[-1]
```
tmatsuo, line 49: ditto

samples/tests/create_dataset_image_test.py
```diff
@@ -0,0 +1,54 @@
...
+def shared_state():
+    shared_state = {}
+    yield shared_state
+
+
+@pytest.fixture(scope="function", autouse=True)
+def teardown(shared_state):
+    yield
+
+    dataset_id = shared_state["dataset_name"].split("/")[-1]
```
tmatsuo, line 39: ditto

samples/tests/create_dataset_tabular_bigquery_test.py
```diff
@@ -0,0 +1,57 @@
...
+def shared_state():
+    state = {}
+    yield state
+
+
+@pytest.fixture(scope="function", autouse=True)
+def teardown(shared_state):
+    yield
+
+    dataset_id = shared_state["dataset_name"].split("/")[-1]
```
tmatsuo, line 40: ditto

samples/tests/create_dataset_tabular_gcs_test.py
```diff
@@ -0,0 +1,57 @@
...
+def shared_state():
+    state = {}
+    yield state
+
+
+@pytest.fixture(scope="function", autouse=True)
+def teardown(shared_state):
+    yield
+
+    dataset_id = shared_state["dataset_name"].split("/")[-1]
```
tmatsuo, line 40: ditto

samples/tests/create_dataset_test.py
```diff
@@ -0,0 +1,59 @@
...
+def shared_state():
+    shared_state = {}
+    yield shared_state
+
+
+@pytest.fixture(scope="function", autouse=True)
+def teardown(shared_state):
+    yield
+
+    dataset_id = shared_state["dataset_name"].split("/")[-1]
```
tmatsuo, line 42: ditto

samples/tests/create_dataset_video_test.py
```diff
@@ -0,0 +1,57 @@
...
+def teardown(shared_state):
+    yield
+
+    dataset_id = shared_state["dataset_name"].split("/")[-1]
```
tmatsuo, line 42: ditto

samples/create_dataset_image_sample.py
vinnysenthil: @dizcology has updated all our samples to take a timeout param instead of setting it inline

tmatsuo (commented):
I still see one `split` without asserting.

tmatsuo (commented):
Also tests are failing.
Maybe you can just put the test files in the same directory as the samples?

samples/tests/cancel_training_pipeline_test.py
vinnysenthil: Added

vinnysenthil:
@tmatsuo - Added that last assertion.

The test have dependencies on samples in other batch PRs. The tests will successfully run before we merge the `sample-final-stage` -> `master` PR.

tmatsuo (commented):
LGTM, I'll make the final approval when we see the tests are passing!

PR #13, opened 2020-09-25
feat: Migrate sample and sample tests - Batch 3 of 5

### LINKED ISSUE #9: Add samples and sample tests

vinnysenthil:

Fixes #9 (Batch 3 of 5) 🦕

Github issues raised by this PR: #51 , #52


samples/export_model_sample.py
```diff
@@ -0,0 +1,35 @@
...
+
+# [START aiplatform_export_model_sample]
+from google.cloud import aiplatform
+
+
+def export_model_sample(
+    gcs_destination_output_uri_prefix: str, project: str, model_id: str
+):
+    client_options = dict(api_endpoint="us-central1-aiplatform.googleapis.com")
+    client = aiplatform.ModelServiceClient(client_options=client_options)
```
kurtisvg, line 23: A comment here is useful to suggest that a user should reuse clients:
```suggestion
    # Initialize client that will be used to create and send requests. This client only needs to be created
    # once, and can be reused for multiple requests.
    client_options = dict(api_endpoint="us-central1-aiplatform.googleapis.com")
    client = aiplatform.ModelServiceClient(client_options=client_options)
```

samples/export_model_sample.py
```diff
@@ -0,0 +1,35 @@
...
+    location = "us-central1"
+    name = client.model_path(project=project, location=location, model=model_id)
+    output_config = {
+        "gcs_destination": {"output_uri_prefix": gcs_destination_output_uri_prefix}
+    }
+    response = client.export_model(name=name, output_config=output_config)
```
kurtisvg, line 29: nit: some spacing in this function might help with readability. Generally for samples we recommended to have at least:
* "arrange" - initialize the client or other resources
* "act" - create and send the request
* "assert" - verify the response is correct (print it out)

samples/get_batch_prediction_job_sample.py
```diff
@@ -0,0 +1,85 @@
...
+        print("  message:", partial_failure.message)
+        details = partial_failure.details
+    resources_consumed = response.resources_consumed
+    print(" resources_consumed")
+    print("  replica_hours:", resources_consumed.replica_hours)
+    completion_stats = response.completion_stats
+    print(" completion_stats")
+    print("  successful_count:", completion_stats.successful_count)
+    print("  failed_count:", completion_stats.failed_count)
+    print("  incomplete_count:", completion_stats.incomplete_count)
```
kurtisvg, line 82: This is a lot of prints. Do we need all of those? Can we print 1-2 and reference the API documentation or add a print for the rest of the objects attributes?

kurtisvg: (same in a bunch of other samples - more than 2-4 prints seems excessive).

samples/get_batch_prediction_job_sample.py
```diff
@@ -0,0 +1,85 @@
...
+def get_batch_prediction_job_sample(project: str, batch_prediction_job_id: str):
+    client_options = dict(api_endpoint="us-central1-aiplatform.googleapis.com")
+    client = aiplatform.JobServiceClient(client_options=client_options)
+    location = "us-central1"
+    name = client.batch_prediction_job_path(
+        project=project, location=location, batch_prediction_job=batch_prediction_job_id
+    )
+    response = client.get_batch_prediction_job(name=name)
+    print("response")
+    print(" name:", response.name)
```
kurtisvg, line 29: Where possible, f-strings are preferred: 
```suggestion
    print(f" {response.name}:")
```

samples/get_model_evaluation_sample.py
```diff
@@ -0,0 +1,36 @@
...
+# [START aiplatform_get_model_evaluation_sample]
+from google.cloud import aiplatform
+from google.protobuf import json_format
+
+
+def get_model_evaluation_sample(project: str, model_id: str, evaluation_id: str):
+    client_options = dict(api_endpoint="us-central1-aiplatform.googleapis.com")
+    client = aiplatform.ModelServiceClient(client_options=client_options)
+    location = "us-central1"
+    name = "projects/{project}/locations/{location}/models/{model}/evaluations/{evaluation}".format(
```
kurtisvg, line 24: Another great candidate for f-strings

samples/tests/export_model_tables_classification_test.py
```diff
@@ -0,0 +1,42 @@
+# Generated code sample for google.cloud.aiplatform.DatasetServiceClient.get_dataset
```
kurtisvg, line 1: Is this actually generated code?

I think the License is supposed to be the first thing, minus shebangs

samples/tests/export_model_tables_classification_test.py
```diff
@@ -0,0 +1,42 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import pytest
+
+from samples import export_model_tables_classification_sample
+from google.cloud import storage
+
+PROJECT_ID = "ucaip-sample-tests"
```
kurtisvg, line 21: These should be env vars: https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/AUTHORING_GUIDE.md#external-resources

samples/tests/export_model_tables_classification_test.py
```diff
@@ -0,0 +1,42 @@
...
+MODEL_ID = "5359002081594179584" # iris 1000
+GCS_URI = "gs://ucaip-samples-test-output/tmp/export_model_test"
+
+@pytest.fixture(scope="function", autouse=True)
+def teardown():
+    yield
+
+    storage_client = storage.Client()
+    bucket = storage_client.get_bucket("ucaip-samples-test-output")
+    blobs = bucket.list_blobs(prefix='tmp/export_model_test')
```
kurtisvg, line 31: we should use a UUID here, to make sure these tests running in parallel don't break anything: https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/AUTHORING_GUIDE.md#temporary-resources

kurtisvg (changes requested):

samples/export_model_sample.py
```diff
@@ -0,0 +1,35 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# [START aiplatform_export_model_sample]
+from google.cloud import aiplatform
+
+
+def export_model_sample(
+    gcs_destination_output_uri_prefix: str, project: str, model_id: str
+):
```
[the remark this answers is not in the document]
dizcology: For python, the no-arg sample function makes testing more difficult, since these need to be passed in for the tests to run successfully.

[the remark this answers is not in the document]
kurtisvg: You still leave the underlying function for testing. I've edited my original comment to better show what I meant.

samples/get_batch_prediction_job_sample.py
dizcology: The idea was to expose as much as possible of the nested structure of these messages to the reader of the sample.  The API reference doc get difficult to navigate and important fields sometimes get hidden.  This being said, I see what you mean about excessive print statements.

Technically it is a bit simpler for for to control the depth of the prints.  Do you think the following might serve as an acceptable compromise?

```
...
response = client.get_batch_prediction_job(name = name)
    print('response')
    print(' name:', response.name)
    print(' display_name:', response.display_name)
    print(' model:', response.model)
    print(' model_parameters:', json_format.MessageToDict(response._pb.model_parameters))
    print(' generate_explanation:', response.generate_explanation)
    print(' state:', response.state)
    print(' create_time:', response.create_time)
    print(' start_time:', response.start_time)
    print(' end_time:', response.end_time)
    print(' update_time:', response.update_time)
    print(' labels:', response.labels)

    # Additional nested fields:
    # input_config = response.input_config
    # output_config = response.output_config
    # output_info = response.output_info
    # error = response.error
    # partial_failures = response.partial_failures
    # resources_consumed = response.resources_consumed
    # completion_stats = response.completion_stats
```

busunkim96:
The usual flow is to merge to the main branch and then cut the release PR. Is it possible to change the branch you are merging into to `master`?

@dizcology This is likely also the the samples presubmit is missing. The kokoro configs added in #16 aren't present in the `release-v0.2.0` branch.

samples/tests/export_model_tables_classification_test.py
```diff
@@ -0,0 +1,43 @@
...
+
+import pytest
+import os
+
+from samples import export_model_tables_classification_sample
+from google.cloud import storage
+
+PROJECT_ID = os.getenv("BUILD_SPECIFIC_GCLOUD_PROJECT")
+MODEL_ID = "5359002081594179584"  # iris 1000
+GCS_URI = "gs://ucaip-samples-test-output/tmp/export_model_test"
```
kurtisvg, line 24: I thought I had a comment regarding this before as well, but may have missed it. 

Fixed resources should be passed in by an environment variables. See [External Resources](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/AUTHORING_GUIDE.md#external-resources) section of the sample style guide. 

Users and googlers without access to these resources should be able to run tests on these samples w/o having to modify them

samples/get_model_evaluation_sample.py
kurtisvg: This looks unresolved as well.

kurtisvg (changes requested):

samples/export_model_sample.py
vinnysenthil: Good call, this comment was added to all AI Platform samples 🙂

samples/export_model_sample.py
```diff
@@ -0,0 +1,35 @@
...
+    client_options = dict(api_endpoint="us-central1-aiplatform.googleapis.com")
```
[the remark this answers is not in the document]
vinnysenthil: Fixed in all samples

samples/tests/export_model_tables_classification_test.py
vinnysenthil: All AI Platform samples now use env vars for `PROJECT_ID`

samples/get_batch_prediction_job_sample.py
```diff
@@ -0,0 +1,51 @@
...
+    print(" start_time:", response.start_time)
+    print(" end_time:", response.end_time)
+    print(" labels:", response.labels)
+    input_config = response.input_config
+    output_config = response.output_config
+    output_info = response.output_info
+    error = response.error
+    partial_failures = response.partial_failures
+    resources_consumed = response.resources_consumed
+    completion_stats = response.completion_stats
```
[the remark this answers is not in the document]
vinnysenthil: Fixed.

samples/tests/export_model_tables_classification_test.py
vinnysenthil: Added UUID to this test

kurtisvg (changes requested):
I've copy/pasted several comments from previous reviews that weren't addressed.

samples/tests/export_model_tables_classification_test.py
aribray: I changed those two instances to `os.getenv("XYZ")`, but I don't think I'm resolving the root issue. From the style guide, resources

> should be identified and passed in through an environment variable

Does **identifying** == the ID of the resource in a comment, for the user to see? Thanks!

samples/tests/get_model_test.py
```diff
@@ -0,0 +1,28 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import pytest
+import os
+
+from samples import get_model_sample
+
+PROJECT_ID = os.getenv("BUILD_SPECIFIC_GCLOUD_PROJECT")
```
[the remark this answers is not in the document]
aribray: I dropped a response on another comment, but the long and short of it is that I want to be sure of the best way to identify the resource and pass it through the environment variables

samples/get_model_evaluation_sample.py
```diff
@@ -0,0 +1,33 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# [START aiplatform_get_model_evaluation_sample]
+from google.cloud import aiplatform
+
+
+def get_model_evaluation_sample(
+    project: str, model_id: str, evaluation_id: str, location: str = "us-central1"
+):
```
kurtisvg, line 21: Please add a wrapper function to explain the arguments for this sample

samples/create_custom_job_sample.py
```diff
@@ -0,0 +1,60 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# [START aiplatform_create_custom_job_sample]
+from google.cloud import aiplatform
+
+
+def create_custom_job_sample(display_name: str, container_image_uri: str, project: str):
```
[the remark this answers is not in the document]
kurtisvg: This sample looks great 👍 

However, there are several samples still missing one of these options

samples/tests/export_model_tables_classification_test.py
kurtisvg: No, you don't need to add a comment. The intended meaning is probably more literally "specify". 

Essentially what we want here is that I as a developer, should be able to run the tests on these samples without modifying them to confirm they still work. However unless I'm on the team maintaining the samples, I likely don't have access to the resources it's tested with. By using env variables means all I need to do when running locally is change the env var, and not worry about changing the model ids when there is a problem. 

So something like this is perfectly acceptable:
```python
resource_id = os.getenv("YOUR-RESOURCE-NAME")
```

For bonus points you can include a link to instructions for creating the resource, if you really want to make it accessible for other developers.

samples/tests/get_model_test.py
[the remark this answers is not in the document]
kurtisvg: responded to other comment

kurtisvg (changes requested):

samples/tests/export_model_tables_classification_test.py
aribray: TO-DO - replace hardcoded resource values with environment variables in all samples

kurtisvg (approved):

kurtisvg:
Please include the issue number for tracking the outstanding work before submitting.

dizcology (approved):

PR #16, opened 2020-09-28
setup sample testing configurations

dizcology:


vinnysenthil (approved):
As discussed on a call earlier today, LGTM.

PR #17, opened 2020-10-01
export aiplatform_v1beta1 as aiplatform.gapic_v1beta1

dizcology:
*DO NOT MERGE* until confirmed about the approach.

Current user code (which will be broken after this change):
```
dataset = aiplatform.Dataset(...)
```

Expected user code after this change:
```
from google.cloud.aiplatform import gapic_v1beta1
dataset = gapic_v1beta1.Dataset(...)
```
or 

```
dataset = aiplatform.gapic_v1beta1.Dataset(...)
```


vinnysenthil (approved):

vinnysenthil:
Namespace change has been LGTM'd

busunkim96:
I'd prefer you actually create and put the code in that namespace. (i.e. move the code into `google/cloud/aiplatform/gapic`. There's a generator option to customize namespaces. (https://github.com/googleapis/googleapis/blob/b821f320473c8ec05a1c7fb9a496c958b1ab9834/google/devtools/clouderrorreporting/v1beta1/BUILD.bazel#L161)

Is the goal to split the generated layer from handwritten bits?

vinnysenthil:
That is the goal, I'll let @dizcology follow up with a suggestion/decision.

dizcology:
@software-dov could you point us to an example of setting namespace and name (https://github.com/googleapis/gapic-generator-python/blob/master/gapic/generator/options.py#L35-L36) through the GAPICMicrogenerator (https://github.com/googleapis/synthtool/blob/0762e8ee2ec21cdfc4d82020b985a104feb0453b/synthtool/gcp/gapic_microgenerator.py#L34)?

PR #22, opened 2020-10-05
fix: re-add py sessions to noxfile

busunkim96:
It looks like these disappeared in a manual generation.

Unfortunately that likely means nox hasn't been running most of the sessions, so this PR may fail on existing code.


tmatsuo (approved):

vinnysenthil (approved):

dizcology (approved):

google/cloud/aiplatform/__init__.py
```diff
@@ -15,6 +15,4 @@
 # limitations under the License.
 #
 
-from google.cloud.aiplatform import gapic
```
busunkim96, line 18: This was raising an "imported but unused error". Let me know if I should re-add it.

noxfile.py
```diff
@@ -156,7 +156,7 @@ def docs(session):
     shutil.rmtree(os.path.join("docs", "_build"), ignore_errors=True)
     session.run(
         "sphinx-build",
-        "-W",  # warnings as errors
```
busunkim96, line 159: I remove this because there were warnings in the docs build. The session will only fail on sphinx errors.

busunkim96:
Not sure why cla-bot is unhappy, but all commits were made by Googlers with valid CLAs.

busunkim96:
@dizcology Please review the comments and merge if you're OK with the additional changes I made

PR #26, opened 2020-10-08
Add support for default project, credentials, and location to initali…

sasha-gitg:
…zer. Initializer tests. Moved initalizer based utils from utils to Init class. Formatting fixes.

Fixes b/169762813 🦕

### DIFF
--- a/google/cloud/aiplatform/initializer.py
+++ b/google/cloud/aiplatform/initializer.py
@@ -17,27 +17,13 @@
 import logging
 from typing import Dict, Optional
 
+from google.api_core import client_options
 import google.auth
 from google.auth import credentials as auth_credentials
 from google.cloud.aiplatform import utils
 
 
-def validate_region(region: str):
-    """Validates region against supported regions.
-
-    Args:
-        region: region to validate
-    Raises:
-        ValueError if region is not in supported regions.
-    """
-    region = region.lower()
-    if region not in SUPPORTED_REGIONS:
-        raise ValueError(
-            f"Unsupported region for AI Platform, select from {SUPPORTED_REGIONS}"
-        )
-
-
-class Init:
+class _Config:
     """Stores common parameters and options for API calls."""
 
     def __init__(self):
@@ -59,14 +45,15 @@ class Init:
         """Updates common initalization parameters with provided options.
 
         Args:
-            project: The default project to use when making API calls.
-            location: The default location to use when making API calls. If not set
-                defaults to us-central-1
-            experiment: The experiment to assign
-            staging_bucket: The default staging bucket to use to stage artifacts when
-                making API calls.
-            credentials: The default custom credentials to use when making API calls. If
-                not provided crendentials will be ascertained from the environment.
+            project (str): The default project to use when making API calls.
+            location (str): The default location to use when making API calls. If not
+                set defaults to us-central-1
+            experiment (str): The experiment to assign
+            staging_bucket (str): The default staging bucket to use to stage artifacts
+                when making API calls. In the form gs://...
+            credentials (google.auth.crendentials.Crendentials): The default custom
+                credentials to use when making API calls. If not provided crendentials
+                will be ascertained from the environment.
         """
         if project:
             self._project = project
@@ -113,8 +100,8 @@ class Init:
     def get_client_options(
         self,
         location_override: Optional[str] = None,
-        prediction_client: Optional[bool] = False,
-    ) -> Dict[str, str]:
+        prediction_client: bool = False,
+    ) -> client_options.ClientOptions:
         """Creates GAPIC client_options using location and type.
 
         Args:
@@ -124,7 +111,7 @@ class Init:
                 Platform (Unified).
 
             prediction_client (bool):
-                True if service client is a PredictionServiceClient, otherwise defaults 
+                True if service client is a PredictionServiceClient, otherwise defaults
                 to False. This is used to provide a prediction-specific API endpoint.
 
         Returns:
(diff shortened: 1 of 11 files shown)


vinnysenthil (approved):
LGTM, all tests pass. Could you just remove the extra `validate_region()`

google/cloud/aiplatform/initializer.py
```diff
@@ -14,28 +14,142 @@
...
+    def staging_bucket(self) -> Optional[str]:
+        """Default staging bucket, if provided."""
+        return self._staging_bucket
+
+    @property
+    def credentials(self) -> Optional[auth_credentials.Credentials]:
+        """Default credentials, if provided."""
+        return self._credentials
+
+    def get_client_options(
```
busunkim96, line 113: In the currently generated surface `api_endpoint` is the only relevant client option but there are others (mostly auth related).

https://googleapis.dev/python/google-api-core/latest/client_options.html

How is this expected to be used? Is the idea that someone would someone call this, add additional client options, and then pass it to the client?

If you're constructing a client options for the user I'd prefer you create the object (for better type checking).

```py
from google.api_core.client_options import ClientOptions

options = ClientOptions(api_endpoint="foo.googleapis.com")
```

vinnysenthil: This is great, `get_client_options()` is a helper method for us to construct the GAPIC client under-the-hood for the user. We should use the `api_core` object

google/cloud/aiplatform/initializer.py
```diff
@@ -14,28 +14,142 @@
...
-			self.project=project
-		if location:
-			self.location=location
-		if experiment:
-			self.experiment=experiment
-		if staging_bucket:
-			self.staging_bucket=staging_bucket
 
 # singleton to store init parameters: ie, aiplatform.init(project=..., location=...)
 singleton = Init()
```
dizcology, line 155: Consider using some class-level patterns to create the singleton (for example, to avoid user incorrectly using the `Init` class): https://python-3-patterns-idioms-test.readthedocs.io/en/latest/Singleton.html.

google/cloud/aiplatform/initializer.py
```diff
@@ -14,28 +14,142 @@
...
+
+    @property
+    def credentials(self) -> Optional[auth_credentials.Credentials]:
+        """Default credentials, if provided."""
+        return self._credentials
+
+    def get_client_options(
+        self,
+        location_override: Optional[str] = None,
+        prediction_client: Optional[bool] = False,
```
dizcology, line 116: I think in this case the typing should be just `bool`.

google/cloud/aiplatform/initializer.py
vinnysenthil: This method will be used by the _developers_ of Model Builder SDK when instantiating a GAPIC client. We will create GAPIC clients under the hood repeatedly throughout the MB SDK Classes and need to be able to use different endpoints based on the region or client type for every method call.

Here is a simple example:

**User 1 Code**
```
from google.cloud import aiplatform as aip

my_dataset = aip.Dataset(
  dataset_name="projects/123456/locations/europe-west4/datasets/987654"
)
```

**User 2 Code**
```
from google.cloud import aiplatform as aip

aip.init(location="asia-east1")
my_dataset = aip.Dataset(dataset_name="34627", location="us-central1")
```


**MB SDK Source Code**
```
from google.cloud.aiplatform import gapic
from google.cloud.aiplatform.utils import get_client_options

class Dataset:
    def __init__(self, dataset_name: str, location: Optional[str] = None):
        """Retrieves dataset resource."""

        self._client = gapic.DatasetServiceClient(
            client_options=get_client_options(location_override=location)
		)

        # 
        # More logic here
        #

        self._dataset = self._client.get_dataset(name=dataset_name)
```

google/cloud/aiplatform/initializer.py
```diff
@@ -14,28 +14,142 @@
...
+
+        Args:
+            location_override (str):
+                Set this parameter to get client options for a location different from
+                location set by initializer. Must be a GCP region supported by AI
+                Platform (Unified).
+
+            prediction_client (bool):
+                True if service client is a PredictionServiceClient, otherwise defaults 
+                to False. This is used to provide a prediction-specific API endpoint.
```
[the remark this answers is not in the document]
vinnysenthil: The MB SDK developer will set this to true if they want to create a PredictionServiceClient as the prediction servers use a different endpoint.

google/cloud/aiplatform/initializer.py
vinnysenthil, line 116:
```suggestion
        prediction_client: bool = False,
```

google/cloud/aiplatform/initializer.py
```diff
@@ -14,28 +14,142 @@
...
+        Args:
+            project: The default project to use when making API calls.
+            location: The default location to use when making API calls. If not set
+                defaults to us-central-1
+            experiment: The experiment to assign
+            staging_bucket: The default staging bucket to use to stage artifacts when
+                making API calls.
+            credentials: The default custom credentials to use when making API calls. If
+                not provided crendentials will be ascertained from the environment.
+        """
```
vinnysenthil, line 70: Adding arg types to doc string

```suggestion
        """Updates common initialization parameters with provided options.

        Args:
            project (str): The default project to use when making API calls.
            location (str): The default location to use when making API calls. If not set
                defaults to us-central-1
            experiment (str): The experiment to assign
            staging_bucket (str): The default GCS staging bucket to use to stage artifacts
                when making API calls. For example, "gs://my-staging-bucket".
            credentials (google.auth.credentials.Credentials): The default custom 
                credentials to use when making API calls. If not provided credentials will
                be ascertained from the environment.
        """
```

google/cloud/aiplatform/initializer.py
busunkim96: Thank you for the explanation @vinnysenthil.  I guess `client_options.api_endpoint` doesn't make much sense if the library is going to be using different endpoints. I still think there needs to be *some way* to pass user defined client options to the generated clients, maybe with the caveat that some will be ignored...

sasha-gitg: Captured in this ticket [b/170489302](b/170489302)

google/cloud/aiplatform/initializer.py
```diff
@@ -14,28 +14,142 @@
...
+    ):
+        """Updates common initalization parameters with provided options.
+
+        Args:
+            project: The default project to use when making API calls.
+            location: The default location to use when making API calls. If not set
+                defaults to us-central-1
+            experiment: The experiment to assign
+            staging_bucket: The default staging bucket to use to stage artifacts when
+                making API calls.
```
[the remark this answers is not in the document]
sasha-gitg: gs://

google/cloud/aiplatform/initializer.py
sasha-gitg: The expectation is for users to not use this class directly and it should be exposed on the SDKs surface. I changed this to a private class and renamed them _Config and global_config to reduce the confusion.

PR #29, opened 2020-10-16
Add initial implementation of Model class.

sasha-gitg:
Includes AI Platform resource noun base class.

Fixes [b/169782649](b/169782649) 🦕

### DIFF
--- a/google/cloud/aiplatform/__init__.py
+++ b/google/cloud/aiplatform/__init__.py
@@ -25,7 +25,6 @@ from google.cloud import aiplatform
 
 aiplatform.init(project='my_project')
 """
-
 init = initializer.global_config.init
 
 
--- a/google/cloud/aiplatform/__init__.py
+++ b/google/cloud/aiplatform/__init__.py
@@ -18,6 +18,7 @@
 from google.cloud.aiplatform import gapic
 
 from google.cloud.aiplatform import initializer
+from google.cloud.aiplatform.models import Model
 
 """
 Usage:
--- a/google/cloud/aiplatform/base.py
+++ b/google/cloud/aiplatform/base.py
@@ -1,0 +1,104 @@
+# -*- coding: utf-8 -*-
+
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import abc
+from typing import Optional
+
+from google.auth import credentials as auth_credentials
+from google.cloud.aiplatform import utils
+from google.cloud.aiplatform import initializer
+
+class AiPlatformResourceNoun(metaclass=abc.ABCMeta):
+    """Base class the AI Platform resource nouns.
+
+    Subclasses require two class attributes:
+
+    client_class: The client to instantiate to interact with this resource noun.
+    is_prediction_client: Flag to indicate if the client requires a prediction endpoint.
+
+    Subclass is required to populate private attribute _gca_resource which is the
+    service representation of the resource noun.
+    """
+
+    @property
+    @classmethod
+    @abc.abstractmethod
+    def client_class(cls) -> utils.AiPlatformServiceClient:
+        """Client class required to interact with resource."""
+        pass
+
+    @property
+    @classmethod
+    @abc.abstractmethod
+    def is_prediction_client(cls) -> bool:
+        """Flag to indicate whether to use prediction endpoint with client."""
+        pass
+    
+    
+    def __init__(self, project: Optional[str]=None, location: Optional[str]=None,
+        credentials: Optional[auth_credentials.Credentials] = None):
+        """Initializes class with project, location, and api_client.
+        
+        Args:
+            project(str): Project of the resource noun.
+            location(str): The location of the resource noun.
+            credentials(google.auth.crendentials.Crendentials): Optional custom
+                credentials to use when accessing interacting with resource noun.
+        """
+
+        self.project = project or initializer.global_config.project
+        self.location = location or initializer.global_config.location
+        
+        self.api_client = self._instantiate_client(self.location, credentials)
+    
+    @classmethod
+    def _instantiate_client(cls, location: Optional[str]=None,
+        credentials: Optional[auth_credentials.Credentials] = None
+        ) -> utils.AiPlatformServiceClient:
+        """Helper method to instantiate service client for resource noun.
+        
+        Args:
+            project (str): Project of the resource noun.
+            location (str): The location of the resource noun.
+            credentials (google.auth.crendentials.Crendentials):
+                Optional custom credentials to use when accessing interacting with
+                resource noun.
+        Returns:
+            client (utils.AiPlatofrmServiceClient):
+                Initialized service client for this service noun.
+        """
+        return initializer.global_config.create_client(
+            client_class=cls.client_class,
+            credentials=credentials,
+            location_override=location,
+            prediction_client=cls.is_prediction_client)
+        
+        
+    @property
+    def name(self) -> str:
+        """Name of this resource."""
+        return self._gca_resource.name.split('/')[-1]
+    
+    @property
+    def resource_name(self) -> str:
+        """Full qualified resource name."""
+        return self._gca_resource.name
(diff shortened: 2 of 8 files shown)


tests/unit/aiplatform/test_models.py
```diff
@@ -0,0 +1,362 @@
...
+_TEST_SERVING_CONTAINER_ARGS = ["--test", "arg"]
+_TEST_SERVING_CONTAINER_ENVIRONMENT_VARIABLES = {
+    "learning_rate": 0.01,
+    "loss_fn": "mse",
+}
+_TEST_SERVING_CONTAINER_PORTS = [8888, 10000]
+_TEST_MODEL_ID = 1234
+
+
+class TestModel:
```
vinnysenthil, line 57: Verified all 10 tests pass ✔️

vinnysenthil (approved):
Looks great, the base class is elegant. Just some small suggested changes and questions

google/cloud/aiplatform/base.py
```diff
@@ -0,0 +1,110 @@
...
+
+import abc
+from typing import Optional
+
+from google.auth import credentials as auth_credentials
+from google.cloud.aiplatform import utils
+from google.cloud.aiplatform import initializer
+
+
+class AiPlatformResourceNoun(metaclass=abc.ABCMeta):
```
vinnysenthil, line 26: Should we have unit tests for this class?

google/cloud/aiplatform/initializer.py
```diff
@@ -135,6 +137,53 @@ def get_client_options(
...
+            location (str): Location. If not provided will use the current location.
+        Returns:
+            resource_parent: Formatted parent resource string.
+        """
+        return "/".join(
+            [
+                "projects",
+                project or self.project,
+                "locations",
+                location or self.location,
```
dizcology, line 156: The parent can be more complex than just `projects/*/locations/*`, e.g. model_evaluation's parents also include `models/*`.

google/cloud/aiplatform/models.py
```diff
@@ -15,9 +15,213 @@
...
+    ModelServiceClient,
+)
+from google.cloud.aiplatform_v1beta1.types import model as gca_model
+from google.cloud.aiplatform_v1beta1.types import env_var
+
+
+class Model(base.AiPlatformResourceNoun):
+
+    client_class = ModelServiceClient
+    is_prediction_client = False
```
dizcology, line 38: Semantically this can be a little confusing, as if we are saying the Model class is prediction client: `Model.is_prediction_client`, while we mean `Model.client.is_prediction_client`.

google/cloud/aiplatform/models.py
```diff
@@ -15,9 +15,213 @@
...
+
+    @property
+    def uri(self):
+        """Uri of the model."""
+        return self._gca_resource.artifact_uri
+
+    @property
+    def description(self):
```
dizcology, line 46: Is this the general pattern we will follow, where every top level attribute of the protobuf message object gets promoted to a python object property?  if so we might do this just once in the resource base class (and carefully add correct docstrings).

google/cloud/aiplatform/models.py
```diff
@@ -15,9 +15,213 @@
...
+        )
+
+        managed_model = gca_model.Model(
+            display_name=display_name,
+            description=description,
+            artifact_uri=artifact_uri,
+            container_spec=container_spec,
+        )
+
+        lro = api_client.upload_model(
```
dizcology, line 211: When the `LRO` class is in place, we will use an instance of that class instead?

google/cloud/aiplatform/initializer.py
```diff
@@ -135,6 +137,53 @@ def get_client_options(
...
+                (Required)An AI Platform Service Client.
+            credentials (auth_credentials.Credentials):
+                Custom auth credentials. If not provided will use the current config.
+            location_override (str): Optional location override.
+            prediction_client (str): Optional flag to use a prediction endpoint.
+        Returns:
+            client: Instantiated AI Platform Service client
+        """
+
+        return client_class(
```
dizcology, line 180: add TODO: add user agent.

google/cloud/aiplatform/models.py
```diff
@@ -15,9 +15,213 @@
...
+        )
+
+        managed_model = lro.result()
+        fields = utils.extract_fields_from_resource_name(managed_model.model)
+        return cls(
+            model_name=fields.id, project=fields.project, location=fields.location
+        )
+
+    # TODO(b/169782716) add support for deployment when Endpoint class complete
+    def deploy(self):
```
dizcology, line 223: this is one place where the resource -> client mapping fails, since we need an EndpointServiceClient to deploy a model.

google/cloud/aiplatform/models.py
dizcology: @sirtorry FYI.

google/cloud/aiplatform/base.py
```diff
@@ -0,0 +1,110 @@
...
+            client_class=cls.client_class,
+            credentials=credentials,
+            location_override=location,
+            prediction_client=cls.is_prediction_client,
+        )
+
+    @property
+    def name(self) -> str:
+        """Name of this resource."""
+        return self._gca_resource.name.split("/")[-1]
```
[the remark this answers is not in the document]
sasha-gitg: Correct, it's not possible to instantiate Model without _gca_resource.

tests/unit/aiplatform/test_models.py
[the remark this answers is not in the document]
sasha-gitg: Yes, https://docs.pytest.org/en/stable/getting-started.html#group-multiple-tests-in-a-class

```
pytest discovers all tests following its Conventions for Python test discovery, so it finds both test_ prefixed functions. There is no need to subclass anything, but make sure to prefix your class with Test otherwise the class will be skipped. We can simply run the module by passing its filename:```

File: google/cloud/aiplatform/base.py (hunk 2 above)
Comment (sasha-gitg) [reply google/cloud/aiplatform/base.py:26]: We cannot instantiate this class so we would need to subclass it and would be duplicating implementation and testing done by the Model class.

File: google/cloud/aiplatform/initializer.py (hunk 3 above)
Comment (sasha-gitg) [reply google/cloud/aiplatform/initializer.py:159]: Renaming to common location path.

File: google/cloud/aiplatform/base.py (hunk 2 above)
Comment (vinnysenthil) [reply google/cloud/aiplatform/base.py:26]: Got it, thanks

File: google/cloud/aiplatform/models.py (hunk 4 above)
Comment (sasha-gitg) [reply google/cloud/aiplatform/models.py:38]: Renaming to _is_client_prediction_client

File: google/cloud/aiplatform/models.py (hunk 5 above)
Comment (sasha-gitg) [reply google/cloud/aiplatform/models.py:46]: It's a general pattern. I didn't add `description` to the ABC because it's not shared by all Nouns. For example, TrainingPipeline.

File: google/cloud/aiplatform/models.py (hunk 6 above)
Comment (sasha-gitg) [reply google/cloud/aiplatform/models.py:211]: Yes.

**Comment by vinnysenthil** [comment]
@googlebot I fixed it.

File: google/cloud/aiplatform/models.py (hunk 8 above)
Comment (sasha-gitg) [reply google/cloud/aiplatform/models.py:223]: That client is not mapped to the Model resource because this method returns an Endpoint class with EndpointServiceClient. That Endpoint class will either be passed in or instantiated.

File: google/cloud/aiplatform/initializer.py (hunk 7 above)
Comment (sasha-gitg) [reply google/cloud/aiplatform/initializer.py:184]: Done.

File: tests/unit/aiplatform/test_initializer.py [hunk 10]
```diff
@@ -19,22 +19,25 @@
...
-_TEST_PROJECT='test-project'
-_TEST_LOCATION='us-central1'
-_TEST_INVALIED_LOCATION='test-invalid-location'
-_TEST_EXPERIMENT='test-experiment'
-_TEST_STAGING_BUCKET='test-bucket'
+_TEST_PROJECT = "test-project"
+_TEST_PROJECT_2 = "test-project-2"
+_TEST_LOCATION = "us-central1"
+_TEST_LOCATION_2 = "europe-west4"
+_TEST_INVALIED_LOCATION = "test-invalid-location"
```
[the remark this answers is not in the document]
sasha-gitg: Done.

google/cloud/aiplatform/__init__.py
```diff
@@ -25,7 +25,6 @@
 
 aiplatform.init(project='my_project')
 """
```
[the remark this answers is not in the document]
sasha-gitg: Done.

google/cloud/aiplatform/initializer.py
sasha-gitg: Done.

google/cloud/aiplatform/models.py
sasha-gitg: Done.

PR #30, opened 2020-10-19
feat: LRO handler

sirtorry:
- sets refresh, cancel and result_type to default
- automatically attaches callback

fixes b/169779287


google/cloud/aiplatform/lro.py
```diff
@@ -15,6 +15,44 @@
...
-class LRO:
-    pass
+from google.api_core import operation as ga_operation
+from google.cloud.aiplatform_v1beta1.services.endpoint_service.transports.base import (
+    EndpointServiceTransport
+)
+from google.protobuf import empty_pb2 as empty
+
+
+class LRO(ga_operation):
```
dizcology, line 27: `ga_operation` is not a class.  Perhaps you meant `ga_operation.Operation`, but i don't think we want to inherit from it here.

google/cloud/aiplatform/lro.py
```diff
@@ -15,6 +15,44 @@
...
+    """A handler for Operation futures.
```
dizcology, line 28: this docstring is incomplete, and should go under the `__init__` method.

google/cloud/aiplatform/lro.py
```diff
@@ -15,6 +15,44 @@
...
+        operation (google.longrunning.operations_pb2.Operation): The
+            initial operation.
+        result_type (func:`type`): The protobuf type for the operation's
+            result.
+    """
+
+    def __init__(
+        self, operation, refresh=None, cancel=None, result_type=empty.Empty, **kwargs
+    ):
+        operations_client = EndpointServiceTransport().operations_client
```
dizcology, line 40: I don't think this works.  Need to create an operations_client in a similar way as the other resource noun classes.

google/cloud/aiplatform/lro.py
```diff
@@ -15,6 +15,44 @@
...
+
+    Args:
+        operation (google.longrunning.operations_pb2.Operation): The
+            initial operation.
+        result_type (func:`type`): The protobuf type for the operation's
+            result.
+    """
+
+    def __init__(
+        self, operation, refresh=None, cancel=None, result_type=empty.Empty, **kwargs
```
dizcology, line 38: need typing annotations.

google/cloud/aiplatform/lro.py
```diff
@@ -15,6 +15,44 @@
...
+        if refresh is None:
+            refresh = functools.partial(operations_client.get_operation, operation.name)
+        if cancel is None:
+            cancel = functools.partial(
+                operations_client.cancel_operation, operation.name
+            )
+
+        super().__init__(operation, refresh, cancel, result_type, **kwargs)
+        self.operation_name = operation.name
+        self.add_done_callback(callback)
```
dizcology, line 50: looks like `callback` is not defined at this point.

google/cloud/aiplatform/lro.py
```diff
@@ -15,6 +15,44 @@
...
+
+    def callback(operation_future):
+        """Callback used to set results of operation future.
+
+        Args:
+            operation_future (ga_operation.Operation): operation to set result of
```
dizcology, line 56: of?

tests/unit/aiplatform/test_lro.py
```diff
@@ -0,0 +1,54 @@
...
+from google.api_core import exceptions
+from gopgle.api_core.tests.unit import test_operation
+from google.cloud.aiplatform import lro
+from google.protobuf import empty_pb2 as empty
+
+TEST_OPERATION_NAME = "test/operation"
+
+
+def make_lro():
+    operation = make_operation_proto()
```
dizcology, line 27: what is `make_operation_proto`?

dizcology (changes requested):

tests/unit/aiplatform/test_lro.py
```diff
@@ -0,0 +1,54 @@
...
+    assert len(lro._done_callbacks) is 1
+    assert lro._result_type is empty.Empty
+    assert lro.metadata() is None
+    assert lro.done() is False
+    assert lro.running()
+
+
+def test_callback():
+    lro = make_lro()
+    lro._operation.done = True
```
dizcology, line 46: the lro object of the LRO class in this PR does not have an `_operation` attribute.

dizcology (changes requested):

sasha-gitg (changes requested):
High level comment, taking this LRO as an example: https://github.com/googleapis/python-aiplatform/blob/dev/google/cloud/aiplatform/models.py#L211

When we receive an LRO from the GAPIC API call, it's already initialized and polling so we may not want to initialize a separate LRO and have two LROs polling for the same result. Unless we want to delete the initial LRO which I think adds too much complexity for what we want.

Instead, I think we want a Class that takes the LRO initialized by the GAPIC library as an attribute, exposes methods that are relevant to the MB usage (result, add_done_callback).

Additionally, we have a requirement to be able to update a MB object with resource properties once the LRO is complete. In the example linked above that would look something like:

```python
lro = api_client.upload_model(...)
mb_lro = LRO(lro)
empty_model = cls.init_from_lro(api_client, mb_lro)
return empty_model
```

The `init_from_lro` on Model would look something like:

``` python
@classmethod
def init_from_lro(cls, api_client, lro):
  self = cls.__new__(cls)
  self.api_client = api_client
  self.lro = lro
  self.lro.add_update_resource_callback(
     self, lro_result_key='model, api_get=lambda name: api_client.get_model(name=name) )
  return self
```

So the LRO class would look something like:

```python
class LRO:

  def __init__(self, gapic_lro: operation.Operation):
    self._lro = gapic_lro

 def add_update_resource_callback(self,
   resource_noun_obj: base.AiPlatformResourceNoun,
   result_key: str, api_get: Callable):
   def callback(result):
     result_value = getattr(result, result_key)
     resource = api_get(result_value)
     resource_noun_obj._gca_resource = resource
   self._lro.add_done_callback(callback)
```

google/cloud/aiplatform/lro.py
dizcology: I think it does throw a NotImplementedError.  Should use the implementation from transports/grpc.py instead, *if* we are going to create the operations_client by borrowing from one of the services.  I think this is not the right way to do it, but we should just create an operations_client directly by providing the appropriate grpc channel.

tests/unit/aiplatform/test_lro.py
```diff
@@ -0,0 +1,59 @@
...
+from gopgle.api_core.tests.unit import test_operation
+from google.cloud.aiplatform import base
+from google.cloud.aiplatform import lro
+from google.protobuf import empty_pb2 as empty
+
+TEST_OPERATION_NAME = "test/operation"
+
+
+def make_lro():
+    operation = test_operation.make_operation_future()
```
dizcology, line 24: probably not a good idea to import a helper function from api_core's unit tests.

dizcology (changes requested):

google/cloud/aiplatform/lro.py
```diff
@@ -17,4 +17,31 @@
...
+
+    def __init__(self, operation: google.api_core.operation.Operation):
+        """Initialises class with operation.
+
+        Args:
+            operation (google.api_core.operation.Operation): operation to handle
+        """
+        self._operation_future = operation
+
+    def add_update_resource_callback(
```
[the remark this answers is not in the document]
sirtorry: @sasha-gitg thoughts

tests/unit/aiplatform/test_lro.py
sirtorry: how so?

google/cloud/aiplatform/lro.py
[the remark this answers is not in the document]
sirtorry: made both possible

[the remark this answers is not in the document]
sasha-gitg: I like it being optional.

google/cloud/aiplatform/lro.py
```diff
@@ -15,6 +15,46 @@
...
+        """Updates resource with result of operation.
+
+        Args:
+            resource_noun_obj (base.AiPlatformResourceNoun):
+                resource to be updated upon operation completion
+        """
+
+        def callback(operation_future):
+            result_obj = operation_future.result()
+            resource_noun_obj._gca_resource = result_obj
```
sasha-gitg, line 53: The result of the operation_future may not be the one we want to assign to _gca_resource. For example operation_future.result for upload_model returns a UploadModelResponse: https://github.com/googleapis/python-aiplatform/blob/master/google/cloud/aiplatform_v1beta1/services/model_service/client.py#L192

That response has the resource name for the model but it's not the Model resource proto we need to assign to _gca_resource which is here: https://github.com/googleapis/python-aiplatform/blob/master/google/cloud/aiplatform_v1beta1/types/model.py#L34

We have to make an additional API call to get that proto. That's why this method should look something like this so we can pass in the get resource proto API call to be executed when the operation is complete:

```python
def add_update_resource_callback(self,
   resource_noun_obj: base.AiPlatformResourceNoun,
   result_key: str, api_get: Callable):
   def callback(result):
     result_value = getattr(operation_future.result(), result_key)
     resource = api_get(name=result_value)
     resource_noun_obj._gca_resource = resource
  self._operation_future.add_done_callback(callback)
```

Usage from Model would look like this:

```python
lro.add_update_resource_callback(self, 'model', api_client.get_model)
```

sasha-gitg (changes requested):

google/cloud/aiplatform/lro.py
sirtorry: ah i didn't account for this scenario. this makes sense.

tests/unit/aiplatform/test_lro.py
sirtorry: figured out why

google/cloud/aiplatform/lro.py
```diff
@@ -15,6 +15,71 @@
...
 class LRO:
-    pass
+    """A handler for operation futures."""
+
+    def __init__(
+        self,
+        operation_future: ga_operation.Operation,
+        resource_noun_obj: Optional[base.AiPlatformResourceNoun] = None,
+        result_key: Optional[str] = None,
+        api_get: Optional[Callable[[proto.Message], proto.Message]] = None,
```
sasha-gitg, line 33: I think the signature for `api_get` should be `Optional[Callable[..., proto.Message]]` because the GAPIC calls take a variable number of key word arguments.

google/cloud/aiplatform/lro.py
```diff
@@ -15,6 +15,71 @@
...
+                result_key and api_get also need to be passed.
+            result_key (str):
+                Optional. Attribute to retrieve from result.
+                resource_noun_obj and api_get also need to be passed.
+            api_get (Callable[[proto.Message],proto.Message]):
+                Optional. Callable that takes resource name and returns object.
+                resource_noun_obj and result_key also need to be passed.
+        """
+        self._operation_future = operation_future
+        if (
```
sasha-gitg, line 51: ```all([resource_noun_obj, result_key, api_get])``` could work here too

google/cloud/aiplatform/lro.py
```diff
@@ -15,6 +15,71 @@
...
+
+        def callback(operation_future):
+            result_value = getattr(operation_future.result(), result_key)
+            resource = api_get(name=result_value)
+            resource_noun_obj._gca_resource = resource
+
+        self._operation_future.add_done_callback(callback)
+
+    @property
+    def operation_future(self):
```
sasha-gitg, line 83: Return type hint: `-> ga_operation.Operation`

tests/unit/aiplatform/test_lro.py
```diff
@@ -0,0 +1,111 @@
...
+    test_lro = lro.LRO(operation_future)
+    resource_noun_obj = AiPlatformResourceNounImpl()
+    result_key = "name"
+    api_get = mock.Mock(spec=["__call__"])
+    test_lro.add_update_resource_callback(resource_noun_obj, result_key, api_get)
+
+    assert len(test_lro._operation_future._done_callbacks) is 1
+
+
+def test_operation_future():
```
sasha-gitg, line 103: This test looks like `test_constructor` with one less assertion. Can it be removed?

sasha-gitg (approved):

tests/unit/aiplatform/test_lro.py
```diff
@@ -0,0 +1,97 @@
...
+    if response is not None:
+        operation_proto.response.Pack(response)
+
+    if error is not None:
+        operation_proto.error.CopyFrom(error)
+
+    return operation_proto
+
+
+def make_operation_future(client_operations_responses=None):
```
dizcology, line 51: For later: move this into test fixtures.

tests/unit/aiplatform/test_lro.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+def test_constructor_with_update():
+    operation_future = make_operation_future()
+    resource_noun_obj = AiPlatformResourceNounImpl()
+    result_key = "name"
+    api_get = mock.Mock(spec=["__call__"])
+    test_lro = lro.LRO(operation_future, resource_noun_obj, result_key, api_get)
+
+    assert len(test_lro.operation_future._done_callbacks) is 1
```
dizcology, line 86: Need a test that verifies that the callback is actually called and that `resource_noun_obj._gca_resource` is correctly populated.

sirtorry:
adding a blocking version of resource updating

tests/unit/aiplatform/test_lro.py
sirtorry: we can verify that the callback is actually called but wouldn't it make more sense as unit test of ga_operation.Operation or google.api_core.future.polling.PollingFuture?

dizcology: This `LRO` class is depending on how ga_operation handles callbacks, for example we are assuming that it would be called with the correct context (with `resource_noun_obj` in scope, for example).  This depends on how ga_operation implements callback, and we should make sure it works in the way we need it to.  For instance, python concurrent.future's callback includes a `context` parameter, implying that some variables would not be in scope.

dizcology (approved):
Please make sure the kokoro build passes before merging.

google/cloud/aiplatform/lro.py
```diff
@@ -15,6 +15,90 @@
...
+                result_key and api_get also need to be passed.
+            result_key (str):
+                Optional. Attribute to retrieve from result.
+                resource_noun_obj and api_get also need to be passed.
+            api_get (Callable[...,proto.Message]):
+                Optional. Callable that takes resource name and returns object.
+                resource_noun_obj and result_key also need to be passed.
+        """
+        self._operation_future = operation_future
+        if all([resource_noun_obj, result_key, api_get]):
```
ivanmkc, line 51: Should an exception be thrown if api_get if passed without `resource_noun_obj` and `result_key` being passed as well.

Otherwise, could lead to some hard to debug situations if the user misses that requirement.

ivanmkc: I'm trying to think of a more explicit way to show required arguments.

In Java, multiple constructors could solve this. Swift could solve this with multiple constructors or associated value enums.

If there's a more Pythonic way to express this (rather than just hoping the user reads the documentation), I'd love to know.

busunkim96: +1 for raising if an unsupported combination is passed. 

Here's an example in the generated client constructor:
https://github.com/googleapis/python-automl/blob/0233804885846295508a5fc98929dba598172244/google/cloud/automl_v1/services/prediction_service/client.py#L233-L248

tests/unit/aiplatform/test_lro.py
```diff
@@ -0,0 +1,115 @@
...
+    resource_noun_obj = AiPlatformResourceNounImpl()
+    result_key = "name"
+    api_get = mock.Mock(spec=["__call__"])
+    test_lro = lro.LRO(operation_future, resource_noun_obj, result_key, api_get)
+
+    assert len(test_lro.operation_future._done_callbacks) == 1
+
+
+def test_update_resource():
+    expected_result = UploadModelResponse(model="tardigrade")
```
sirtorry, line 85: looks like Any.Pack() errors because this proto.Message does not have descriptor. does proto plus not handle this? I'm a little confused.

google/cloud/aiplatform/lro.py
ivanmkc: One way would be to use a class method:

```
    def __init__(
        self,
        operation_future: ga_operation.Operation):
        self._operation_future = operation_future

    @classmethod
    def make_with_callback(
        self,
        operation_future: ga_operation.Operation,
        resource_noun_obj: base.AiPlatformResourceNoun,
        result_key: str,
        api_get: Callable[..., proto.Message],
    ):
        obj = cls.__new__(cls)  # Does not call __init__
        super(LRO, obj).__init__(operation_future)
        if all([resource_noun_obj, result_key, api_get]):
            self.add_update_resource_callback(resource_noun_obj, result_key, api_get)
        else:
            # throw exception
```

This is my attempt at wrapping my head around Python so I'm open to comments.

tests/unit/aiplatform/test_lro.py
sirtorry: or i guess a better question is how do I set operations_pb2.Operation.response.

busunkim96: [Proto-plus](https://proto-plus-python.readthedocs.io/en/latest/reference/message.html) wraps the raw python protobuf messages so protobuf Python methods won't work on it directly.

I'm not completely following what you're trying to do. Could I have more context? [This api-core](https://github.com/googleapis/python-api-core/blob/85e14e3f3a6b3e3c3ae9626daf60183f4c05fb6a/google/api_core/operation.py#L116) code sets the response based on the response from the API.

sirtorry: ah yes. I'm trying to run [operation.py#L133](https://github.com/googleapis/python-api-core/blob/85e14e3f3a6b3e3c3ae9626daf60183f4c05fb6a/google/api_core/operation.py#L133) during init. In [test_lro#L89](https://github.com/googleapis/python-aiplatform/blob/c9d33e98cc66b10f4f966ea84eb8b4494bca16b7/tests/unit/aiplatform/test_lro.py#L89), I'm trying to do this by:
1. setting google.longrunning.operations_pb2.Operation.done to True
2. assigning something to google.longrunning.operations_pb2.Operation.response

The problem I'm facing is part 2 as shown in [test_lro#L41](https://github.com/googleapis/python-aiplatform/blob/c9d33e98cc66b10f4f966ea84eb8b4494bca16b7/tests/unit/aiplatform/test_lro.py#L41).

sirtorry: I've tried:
- `operation_proto.response.Pack(response)`
- `operation_proto.response = response`

I'm probably missing something obvious.

google/cloud/aiplatform/lro.py
sasha-gitg: Depends on what you mean by show.

https://docs.python.org/3/library/typing.html#typing.overload
You still have to do the same song and dance in a single method and it's up to the introspection to respect the overloaded signatures.

tests/unit/aiplatform/test_lro.py
busunkim96: Ah makes sense 👍. Since `operation_proto` is a protobuf python object, the underlying `pb`  needs to be assigned. You can convert the `proto.Message` into python protobuf through the `.pb()` class method. 

https://proto-plus-python.readthedocs.io/en/latest/reference/message.html#proto.message.Message.pb


```py
from google.longrunning import operations_pb2
import proto

class Row(proto.Message):
    value = proto.Field(proto.STRING, number=1)

row = Row(value='abc')
operation_proto = operations_pb2.Operation(name="test")
operation_proto.response.Pack(Row.pb(row))
```

```py
>>> operation_proto
name: "test"
response {
  type_url: "type.googleapis.com/Row"
  value: "\n\003abc"
}
```

google/cloud/aiplatform/lro.py
sirtorry: I've come up with a quick solution for this for now. if we think users will use LRO directly, we can improve it in a future PR.

PR #31, opened 2020-10-20
Implement initial Dataset class

vinnysenthil:
### Summary of changes:

- Implement `Dataset` class with following methods
  - `__init__`
  - `_create`
  - `_import`
  - `create`
  - `import_data`
  - `export_data`
- Ten unit tests covering 98% of `Dataset` class
- Add two helper functions to `utils.py` — `validate_id` and `validate_string_list`
- Add an error when `project` is not initialized or set in credentials

**Fixes [b/169781839](http://b/169781839)**


google/cloud/aiplatform/datasets.py
```diff
@@ -15,6 +15,358 @@
...
+        super().__init__(project=project, location=location, credentials=credentials)
+        self._gca_resource = self.api_client.get_dataset(name=dataset_name)
+
+    @classmethod
+    def create(
+        cls,
+        display_name: str,
+        metadata_schema_uri: str,
+        metadata: Sequence[Tuple[str, str]] = (),
+        source: Optional[Union[str, List[str]]] = None,
```
sasha-gitg, line 102: Elevate source and import_schema_uri in the argument order:

ie.

```python
def create(cls, display_name, source, import_schema_uri...)
```

google/cloud/aiplatform/datasets.py
```diff
@@ -15,6 +15,358 @@
...
+        labels: Optional[Dict] = None,
+        data_items_labels: Optional[Dict] = None,
+        import_schema_uri: Optional[str] = None,
+        project: Optional[str] = None,
+        location: Optional[str] = None,
+        credentials: Optional[auth_credentials.Credentials] = None,
+    ) -> "Dataset":
+        """Creates new dataset.
```
sasha-gitg, line 110: Add that importing dataset is optional.

google/cloud/aiplatform/datasets.py
```diff
@@ -15,6 +15,358 @@
...
+    def _create(
+        cls,
+        api_client: DatasetServiceClient,
+        parent: str,
+        display_name: str,
+        metadata_schema_uri: str,
+        labels: Optional[Dict] = {},
+        request_metadata: Sequence[Tuple[str, str]] = (),
+    ) -> operation.Operation:
+
```
sasha-gitg, line 226: Docstring.

google/cloud/aiplatform/datasets.py
```diff
@@ -15,6 +15,358 @@
...
+            labels=labels,
+        )
+
+        return api_client.create_dataset(
+            parent=parent, dataset=gapic_dataset, metadata=request_metadata
+        )
+
+    def _import(
+        self,
+        source: Union[str, List[str]],
```
sasha-gitg, line 239: I think for source we should accept Sequence over List, throughout the code, because it matches the type or GcsSource.uris.

For type checking, to determine whether to envelope the string in a list, check for a string instead of a list.

google/cloud/aiplatform/datasets.py
```diff
@@ -15,6 +15,358 @@
...
+                labels specified inside index file refenced by
+                [import_schema_uri][google.cloud.aiplatform.v1beta1.ImportDataConfig.import_schema_uri],
+                e.g. jsonl file.
+        Returns:
+            operation (Operation):
+                An object representing a long-running operation.
+        """
+
+        import_config = ImportDataConfig(
+            gcs_source=GcsSource(uris=source if type(source) == list else [source]),
```
sasha-gitg, line 278: Either add BQ support or capture it as a bug and add TODO please.

google/cloud/aiplatform/datasets.py
```diff
@@ -15,6 +15,358 @@
...
+                If the uri doesn't end with '/', a '/' will be automatically
+                appended. The directory is created if it doesn't exist.
+
+        Returns:
+            operation (Operation):
+                An object representing a long-running operation.
+        """
+
+        export_data_config = ExportDataConfig(
+            gcs_destination=GcsDestination(output_uri_prefix=output_dir)
```
sasha-gitg, line 362: Same comment for BQ support.

google/cloud/aiplatform/datasets.py
```diff
@@ -15,6 +15,358 @@
...
+        )
+
+        export_lro = self.api_client.export_data(
+            name=self.resource_name, export_config=export_data_config
+        )
+
+        return export_lro
```
sasha-gitg, line 369: To be symmetric with import_data, I think we should block here and return the output_dir uri upon completion. We should add async support for both as a single work item.

google/cloud/aiplatform/datasets.py
```diff
@@ -15,6 +15,358 @@
...
-class Dataset:
-    pass
+from google.api_core import operation
+from google.auth import credentials as auth_credentials
+
+from google.cloud.aiplatform import base
+from google.cloud.aiplatform import utils
+from google.cloud.aiplatform import initializer
+from google.cloud.aiplatform.utils import validate_id
+from google.cloud.aiplatform.utils import validate_string_list
```
sasha-gitg, line 27: Is this unused?

google/cloud/aiplatform/utils.py
```diff
@@ -68,6 +69,16 @@ def _match_to_fields(match: re.Match) -> Optional[Fields]:
     )
 
 
+def validate_string_list(obj: List[str]):
```
sasha-gitg, line 72: If this is unused, remove please.

sasha-gitg (approved):
Looks great! I really like how you used pytest fixtures for patching.

google/cloud/aiplatform/datasets.py
```diff
@@ -15,6 +15,358 @@
...
+                credentials set in aiplatform.init.
+        """
+
+        # Fully qualified dataset name, i.e. "projects/.../locations/.../datasets/12345"
+        valid_name = extract_fields_from_resource_name(
+            resource_name=dataset_name, resource_noun="datasets"
+        )
+
+        # Partial dataset name (i.e. "12345") with known project and location
+        if (
```
dizcology, line 77: will something similar to this block be needed for every resource?  if so let's refactor it into utils or the base class later.

google/cloud/aiplatform/datasets.py
```diff
@@ -15,6 +15,358 @@
...
+                credentials set in aiplatform.init.
+        Returns:
+            dataset (Dataset):
+                Instantiated representation of the managed dataset resource.
+        """
+
+        # Validate that source and import schema are passed together or not at all
+        if bool(source) ^ bool(import_schema_uri):
+            raise ValueError(
+                "Please provide both source and import_schema_uri to import data"
```
dizcology, line 181: Period.  And maybe add "or omit both...".

google/cloud/aiplatform/datasets.py
```diff
@@ -15,6 +15,358 @@
...
+            parent=initializer.global_config.common_location_path(
+                project=project, location=location
+            ),
+            metadata_schema_uri=metadata_schema_uri,
+            request_metadata=metadata,
+            labels=labels,
+            api_client=api_client,
+        )
+
+        created_dataset = create_dataset_lro.result()
```
dizcology, line 197: What's the plan for streaming logs or show polling updates?

google/cloud/aiplatform/datasets.py
```diff
@@ -15,6 +15,358 @@
...
+            project=project,
+            location=location,
+            credentials=credentials,
+        )
+
+        # If an import source was not provided, return empty created Dataset.
+        if not source:
+            return dataset_obj
+
+        return dataset_obj.import_data(
```
dizcology, line 210: This would return an LRO, while in the "no data import" case, we wait until the create_dataset LRO completes.  We need to document clearly (e.g. in the docstring) this inconsistent behavior.

google/cloud/aiplatform/initializer.py
```diff
@@ -76,13 +77,21 @@ def project(self) -> str:
         if self._project:
             return self._project
 
-        _, project_id = google.auth.default()
+        try:
+            _, project_id = google.auth.default()
+        except GoogleAuthError:
+            raise ValueError(
```
dizcology, line 83: Why do we not simply re-raise the GoogleAuthError?

tests/unit/aiplatform/test_datasets.py
```diff
@@ -0,0 +1,256 @@
...
+
+class TestDataset:
+    def setup_method(self):
+        reload(initializer)
+        reload(aip)
+
+    @pytest.fixture
+    def get_dataset_mock(self):
+        with patch.object(DatasetServiceClient, "get_dataset") as get_dataset_mock:
+            get_dataset_mock.return_value = GapicDataset(
```
dizcology, line 77: this can become difficult to maintain to make sure the mocks return correct responses.  we should look into replays.

tests/unit/aiplatform/test_datasets.py
```diff
@@ -0,0 +1,256 @@
...
+_TEST_OUTPUT_DIR = "gs://my-output-bucket"
+
+
+class TestDataset:
+    def setup_method(self):
+        reload(initializer)
+        reload(aip)
+
+    @pytest.fixture
+    def get_dataset_mock(self):
```
dizcology, line 75: perhaps move the mock fixtures into conftest.py so they could be more easily reused.

google/cloud/aiplatform/datasets.py
vinnysenthil: Replacing `Union[str, List[str]]` with `Sequence[str]`

google/cloud/aiplatform/initializer.py
vinnysenthil: In place of the `ValueError`? Good point, I can do that.

google/cloud/aiplatform/datasets.py
vinnysenthil: Yeah these multiple LROs under-the-hood can be misleading to the user. I'm adding this discussion to [b/171275584](http://b/171275584)

google/cloud/aiplatform/datasets.py
vinnysenthil: Added, [b/171311614](http://b/171311614)

google/cloud/aiplatform/datasets.py
vinnysenthil: Unfortunately, that elevates default/optional arguments above `metadata_schema_uri` which does not have a default. I'll elevate source and import_schema_uri as such:

```
    def create(
        cls,
        display_name: str,
        metadata_schema_uri: str,
        source: Optional[Sequence[str]] = None,
        import_schema_uri: Optional[str] = None,
        ...
    )
```

google/cloud/aiplatform/datasets.py
vinnysenthil: This will be accomplished with [b/171275584](http://b/171275584)

tests/unit/aiplatform/test_datasets.py
vinnysenthil: Captured in [b/171333554](http://b/171333554)

PR #34, opened 2020-10-22
feat: Python Packaging class and helpers for Packaging Python Script

sasha-gitg:
Helpers for packaging a Python script into a Python package that is suitable for AI Platform training. 

Also includes helpers to stage the script in GCS for Custom Training.

Relevant Documentation: https://cloud.google.com/ai-platform/training/docs/packaging-trainer

First part of [b/169782649](b/169782649) 🦕

### DIFF
--- a/google/cloud/aiplatform/training_jobs.py
+++ b/google/cloud/aiplatform/training_jobs.py
@@ -17,9 +17,10 @@
 
 import datetime
 import functools
-import os
+import pathlib
 import shutil
 import subprocess
+import sys
 import time
 import tempfile
 from typing import Callable, Dict, Optional, Sequence
@@ -58,7 +59,7 @@ def _timestamped_copy_to_gcs(
         gcs_path (str): The path of the copied file in gcs.
     """
 
-    local_file_name = os.path.basename(local_file_path)
+    local_file_name = pathlib.Path(local_file_path).name
 
     timestamp = datetime.datetime.now().isoformat(sep="-", timespec="milliseconds")
     blob_path = "-".join(["aiplatform", timestamp, local_file_name])
@@ -73,34 +74,20 @@ def _timestamped_copy_to_gcs(
     return gcs_path
 
 
-def _get_python3_alias():
-    """Finds Python 3 alias which is used to execute setuptools packaging.
+def _get_python_executable() -> str:
+    """Returns Python executable.
 
     Raises:
-        RunTimeError if Python 3 alias is not found.
+        EnvironmentError if Python executable is not found.
     Returns:
-        Python 3 command alias to use for setuptools packaging.
+        Python executable to use for setuptools packaging.
     """
 
-    def is_python3(python_cmd):
-        try:
-            check_python = subprocess.check_output([python_cmd, "--version"]).decode()
-        except FileNotFoundError:
-            return False
-        return check_python.startswith("Python 3")
+    python_executable = sys.executable
 
-    python_cmd = "python"
-    if is_python3(python_cmd):
-        return python_cmd
-
-    python_cmd = "python3"
-    if is_python3(python_cmd):
-        return python_cmd
-
-    raise EnvironmentError(
-        "Cannot find Python 3 alias for packaging."
-        'Please alias Python 3 to "python" or "python3".'
-    )
+    if not python_executable:
+        raise EnvironmentError("Cannot find Python executable for packaging.")
+    return python_executable
 
 
 class _TrainingScriptPythonPackager:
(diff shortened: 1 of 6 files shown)


google/cloud/aiplatform/training_jobs.py
```diff
@@ -15,9 +15,309 @@
...
+        script_name = os.path.basename(self.script_path)
+
+        with tempfile.TemporaryDirectory() as tmpdirname:
+            source_distribution_path = self.make_package(tmpdirname)
+            return copy_method(source_distribution_path)
+
+    def package_and_copy_to_gcs(
+        self,
+        staging_bucket: str,
+        project: str = None,
```
busunkim96, line 264:
```suggestion
        project_id: str = None,
```
If this accepts both project numbers and ids please disregard.

busunkim96:
Artifact Registry may be a better fit for packages than GCS down the line. It looks like they are in beta right now. https://cloud.google.com/artifact-registry

setup.py
```diff
@@ -42,6 +42,7 @@
     include_package_data=True,
     install_requires=(
         "google-api-core[grpc] >= 1.22.2, < 2.0.0dev",
+        "google-cloud-storage >= 1.32.0",
```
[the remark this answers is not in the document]
sasha-gitg: 1. Done
2. This is hard dependency because it's required for both python package and script based custom training.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -15,9 +15,309 @@
...
+
+
+def _get_python3_alias():
+    """Finds Python 3 alias which is used to execute setuptools packaging.
+
+    Raises:
+        RunTimeError if Python 3 alias is not found.
+    Returns:
+        Python 3 command alias to use for setuptools packaging.
+    """
```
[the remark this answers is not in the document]
sasha-gitg: Done. Thanks!

google/cloud/aiplatform/training_jobs.py
```diff
@@ -15,9 +15,309 @@
...
+
+        Args:
+            package_directory (str): Directory to build package in.
+        Returns:
+            source_distribution_path (str): Path to built package.
+        Raises:
+            RunTimeError if package creation fails.
+        """
+        # The root folder to builder the package in
+        trainer_root_path = os.path.join(package_directory, self._TRAINER_FOLDER)
```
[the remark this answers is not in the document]

google/cloud/aiplatform/training_jobs.py
sasha-gitg: Ack.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -15,9 +15,297 @@
...
+) -> str:
+    """Copies a local file to a GCS bucket.
+
+    The file copied to GCS is the name of the local file prepended with an
+    "aiplatform-{timestamp}-" string.
+
+    Args:
+        local_file_path (str): Required. Local file to copy to GCS.
+        staging_bucket (str):
+            Required. The staging bucket to copy to. (should not include gs://)
```
dizcology, line 50: I think in some other parts of the uCAIP API `gs://` is expected, perhaps we should keep this consistent.

Also the name `staging_bucket` suggests it should be a bucket, but in reality a GCS prefix will do just fine?

(example: https://github.com/googleapis/python-aiplatform/pull/13/files#diff-c3ab7043e19fe58fca6c927aad405f57839d6c40b651039d881afb8a7eba9040R29)

google/cloud/aiplatform/training_jobs.py
```diff
@@ -15,9 +15,297 @@
...
+        Returns:
+            source_distribution_path (str): Path to built package.
+        Raises:
+            RunTimeError if package creation fails.
+        """
+        # The root folder to builder the package in
+        package_path = pathlib.Path(package_directory)
+
+        # Root directory of the package
+        trainer_root_path = package_path / self._TRAINER_FOLDER
```
dizcology, line 170: This would be writing into the user's file system and could potentially overwrite existing files.  Perhaps build the package in a tmp folder instead?

google/cloud/aiplatform/training_jobs.py
```diff
@@ -15,9 +15,297 @@
...
+
+    @property
+    def module_name(self) -> str:
+        """Module name that can be executed during training. ie. python -m"""
+        return f"{self._ROOT_MODULE}.{self._TASK_MODULE_NAME}"
+
+
+class TrainingJob(base.AiPlatformResourceNoun):
+
+    client_class = PipelineServiceClient
```
dizcology, line 282: Should be `JobServiceClient` instead?  Or perhaps the plan is to not expose CustomJob, requiring its access only through TrainingPipeline?

google/cloud/aiplatform/training_jobs.py
sasha-gitg: This method is invoked through `package_and_copy` which passes in a tmp folder for this reason.

This is a private class so we don't intend for the user to use this class directly.

google/cloud/aiplatform/training_jobs.py
sasha-gitg: The guidance we received was to use the PipelineService instead of the CustomJob service during review because it has stronger integration with the upstream and downstream resources. ie.  InputDataConfig takes a managed dataset, model_to_upload produces a managed model.

vinnysenthil (approved):
Looks great! Logic is clear, thanks for the comments on `make_package()`

google/cloud/aiplatform/training_jobs.py

PR #35, opened 2020-10-28
feat: adds endpoint class and model deploy/undeploy

sirtorry:
Fixes b/171074487.
Partially resolves b/169782716. Will add prediction in separate PR.


google/cloud/aiplatform/models.py
```diff
@@ -215,10 +244,548 @@ def upload(
...
+            deployed_model_display_name (str):
+                Optional. The display name of the DeployedModel. If not provided
+                upon creation, the Model's display_name is used.
+            traffic_percentage (int):
+                Optional. Desired traffic to newly deployed model. Defaults to
+                100. Traffic of previously deployed models at the endpoint  will
+                be scaled down to accommodate new deployed model's traffic.
+            machine_type (str):
+                Optional. The type of machine. Following machine types are
+                supported:
```
dizcology, line 274: is this copied from GAPIC library's docstring?  this can quickly go out of date and become inaccurate.

google/cloud/aiplatform/models.py
```diff
@@ -215,10 +244,548 @@ def upload(
...
+                Managed endpoint resource.
+        """
+
+        # Fully qualified endpoint name, i.e. "projects/.../locations/.../endpoints/12345"
+        valid_name = utils.extract_fields_from_resource_name(
+            resource_name=endpoint_name, resource_noun="endpoints"
+        )
+
+        # Partial endpoint name (i.e. "12345") with known project and location
+        if (
```
dizcology, line 394: if this logic is the same as the other resource nouns, refactor it.

google/cloud/aiplatform/models.py
```diff
@@ -215,10 +244,548 @@ def upload(
...
+        return endpoint_obj
+
+    @classmethod
+    def _create(
+        cls,
+        api_client: EndpointServiceClient,
+        parent: str,
+        display_name: str,
+        description: Optional[str] = None,
+        traffic_split: Optional[Dict] = {},
```
dizcology, line 508: is `traffic_split` needed when creating an endpoint?

also the default value should be `None`, not `{}`.

google/cloud/aiplatform/models.py
```diff
@@ -215,10 +244,548 @@ def upload(
...
+        Returns:
+            traffic_split (Dict):
+                Traffic split to use.
+        """
+        old_models_traffic = 100 - traffic_percentage
+        if old_models_traffic:
+            unallocated_traffic = old_models_traffic
+            for deployed_model in traffic_split:
+                current_traffic = traffic_split[deployed_model]
+                new_traffic = int(current_traffic / 100 * old_models_traffic)
```
dizcology, line 592: does this guarantee that the sum of all percentages equals 100? it is required.

google/cloud/aiplatform/models.py
```diff
@@ -215,10 +244,548 @@ def upload(
...
         )
 
-    # TODO(b/169782716) add support for deployment when Endpoint class complete
-    def deploy(self):
-        raise NotImplementedError("Deployment not implemented.")
+    def deploy(
+        self,
+        endpoint_obj: Optional["Endpoint"] = None,
+        deployed_model_display_name: Optional[str] = None,
+        traffic_percentage: Optional[int] = 100,
```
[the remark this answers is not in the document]
sirtorry: traffic_split is different from traffic_percentage in this file.

google/cloud/aiplatform/models.py
sirtorry: I'll remove the supported machine types. Does that sound good?

google/cloud/aiplatform/models.py
sirtorry: the addition after the loop should guarantee it

google/cloud/aiplatform/models.py
```diff
@@ -215,10 +244,548 @@ def upload(
...
+            machine_type=machine_type,
+            min_replica_count=min_replica_count,
+            max_replica_count=max_replica_count,
+            metadata=metadata,
+        )
+
+        return endpoint_obj
+
+
+class Endpoint(base.AiPlatformResourceNoun):
```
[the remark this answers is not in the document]
sirtorry: @sasha-gitg I recall you wanted endpoint to be in the model file. does this PR change your mind?

google/cloud/aiplatform/models.py
sirtorry: traffic_split should be optional when creating an endpoint.

google/cloud/aiplatform/models.py
```diff
@@ -215,10 +244,522 @@ def upload(
...
+
+        endpoint_client = cls._instantiate_client(
+            location=location, credentials=credentials
+        )
+
+        client_class = PredictionServiceClient
+        _is_client_prediction_client = True
+        prediction_client = cls._instantiate_client(
+            location=location, credentials=credentials
+        )
```
sirtorry, line 462: thoughts on this?

google/cloud/aiplatform/models.py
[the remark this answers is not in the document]
sasha-gitg: The argument to keep these in the same file was to avoid a circular imports because Model.deploy takes Endpoint and Endpoint.deploy takes Model and there is an ask to take their respective resource names and instantiate the classes in the methods.

google/cloud/aiplatform/models.py
```diff
@@ -50,18 +57,22 @@ def __init__(
         location: Optional[str] = None,
         credentials: Optional[auth_credentials.Credentials] = None,
     ):
-        """Retrieves the model resource and instanties it's representation.
+        """Retrieves the model resource and instantiates it's representation.
```
ivanmkc, line 60: nit: The apostrophe is unnecessary. 'Retrieves the model resource and instantiates its representation'

google/cloud/aiplatform/models.py
```diff
@@ -75,12 +86,30 @@ def _get_model(self, model_name: str) -> gca_model.Model:
...
+            resource_name=model_name, resource_noun="models"
         )
 
-        # TODO(b/170954330) add optional instantiation if resource path given
-        model = self.api_client.get_model(name=resource_name)
+        # Partial model name (i.e. "12345") with known project and location
+        if (
+            not valid_name
+            and utils.validate_id(model_name)
+            and (self.project or initializer.global_config.project)
```
ivanmkc, line 99: Perhaps break up this expression since:
1. If both (self.project or initializer.global_config.project) are unset, then the user sees the error:
`ValueError("Please provide a valid model name or ID")` which is misleading.

2. If both (self.location or initializer.global_config.location) are unset, then the user sees the error:
`ValueError("Please provide a valid model name or ID")` which is misleading.

google/cloud/aiplatform/models.py
```diff
@@ -50,18 +57,22 @@ def __init__(
...
+                set in aiplatform.init will be used.
             location (str):
-                Optional location to retrieve model from. If not set, location set in
-                aiplatform.init will be used.
-            credentials (auth_credentials.Credentials):
-                Optional credentials to use to retrieve the model.
+                Optional location to retrieve model from. If not set, location
+                set in aiplatform.init will be used.
+            credentials: Optional[auth_credentials.Credentials]=None,
+                Custom credentials to use to upload this model. Overrides
```
ivanmkc, line 74: nit: Should this read `If not set, credentials set in aiplatform.init will be used.` for consistency?

google/cloud/aiplatform/models.py
```diff
@@ -215,10 +244,522 @@ def upload(
...
+        min_replica_count: Optional[int] = 0,
+        max_replica_count: Optional[int] = None,
+        metadata: Optional[Sequence[Tuple[str, str]]] = (),
+        timeout: Optional[int] = None,
+    ) -> "Endpoint":
+        """
+        Deploys model to endpoint. Endpoint will be created if unspecified.
+
+        Args:
+            endpoint ("Endpoint"):
```
ivanmkc, line 262: May be obvious to you all, but what happens in the case of a naming conflict?

google/cloud/aiplatform/models.py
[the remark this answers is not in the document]
dizcology: Are these two mutually imported only in typing annotations? (in which case we might be able to still split them into separate files)

google/cloud/aiplatform/models.py
```diff
@@ -215,10 +244,522 @@ def upload(
...
+        # Fully qualified endpoint name, i.e. "projects/.../locations/.../endpoints/12345"
+        valid_name = utils.extract_fields_from_resource_name(
+            resource_name=endpoint_name, resource_noun="endpoints"
+        )
+
+        # Partial endpoint name (i.e. "12345") with known project and location
+        if (
+            not valid_name
+            and utils.validate_id(endpoint_name)
+            and (self.project or initializer.global_config.project)
```
ivanmkc, line 380: ditto wrt to the misleading exception

google/cloud/aiplatform/models.py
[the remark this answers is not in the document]
dizcology: I see - so model.deploy takes traffic_percentage, while endpoint.deploy takes traffic_split?  Can these be consolidated?

google/cloud/aiplatform/models.py
dizcology: If parts of the docstring are copied from GAPIC (which may be a very regular thing to do for this library) we might want to explore ways to automatically grab relevant pieces of the GAPIC docstring.

google/cloud/aiplatform/models.py
dizcology: The GAPIC create_endpoint method probably allows omitting the traffic_split field, and if the user passes in anything non empty the service will likely ignore it or raise an error. It should be just omitted in this method.

google/cloud/aiplatform/models.py
dizcology: I see - please be sure to include a test for this.

google/cloud/aiplatform/models.py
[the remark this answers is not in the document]
sasha-gitg: I agree if it was just type annotations but there was an ask during review to support resource names so we would need to instantiate the classes within the methods. ie:

```python

class Model:
    def deploy(endpoint: Union[Endpoint, str], ....):
        if isinstance(endpoint, str):
            endpoint = Endpoint(endpoint)
        ...
```

This isn't P0 but we should take it into account to avoid a future refactor.

google/cloud/aiplatform/models.py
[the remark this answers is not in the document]
sirtorry: endpoint.deploy also takes traffic_percentage.

google/cloud/aiplatform/models.py
sirtorry: i'll look to add this to utils unless there's objection

google/cloud/aiplatform/models.py
sirtorry: I've been copying from the gapic docstring but it's not 1:1 as the mbsdk might change the behaviour prior to calling the gapic function.

google/cloud/aiplatform/models.py
sirtorry: naming conflicts are unlikely if not impossible. they're long computer generated strings.

google/cloud/aiplatform/models.py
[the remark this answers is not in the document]
sirtorry: we'll table this for the future

google/cloud/aiplatform/models.py
sirtorry: allow use of 2 clients, in a separate PR.

google/cloud/aiplatform/models.py
sasha-gitg: `initializer.global_config.location` will always return the default "us-central1" if not set so we don't need to add that condition here.

Additionally, `initializer.global_config.project` will throw an exception if project is not found when being invoked and will be handled below with creating model path so that condition can also be removed here.

* Note, that there is an additional failure mode for `initializer.global_config.project` that is not currently caught: Default project_id has not been set in the environment. I will have changes to raise in that situation in the upcoming CustomTrainingJob PR.

google/cloud/aiplatform/models.py
```diff
@@ -215,10 +229,525 @@ def upload(
...
+        Returns:
+            endpoint (Endpoint):
+                Instantiated representation of the endpoint resource.
+        """
+
+        endpoint_client = cls._instantiate_client(
+            location=location, credentials=credentials
+        )
+
+        client_class = PredictionServiceClient
```
sasha-gitg, line 437: To handle the case of a second client we can create another class method that instantiates the prediction client.

Like so:

```python

@classmethod
def _instantiate_prediction_client(cls, location=None, credentials=None):
  return initializer.global_config.create_client(
            client_class=PredictionServiceClient,
            credentials=credentials,
            location_override=location,
            prediction_client=True,
        )
```

And in this _create method you would create the client which you would use for predict calls.:

```python
prediction_client = cls._instantiate_prediction_client(location=location, credentials=credentials)
```

Similarly, I assume, you would instantiate the prediction client in the constructor:

```python
self.prediction_client = cls._instantiate_prediction_client(location=location, credentials=credentials)
```

google/cloud/aiplatform/models.py
```diff
@@ -215,10 +229,525 @@ def upload(
...
+            location=location,
+            credentials=credentials,
+        )
+
+        create_endpoint_operation.add_update_resource_callback(
+            resource_noun_obj=endpoint,
+            result_key="name",
+            api_get=lambda name: endpoint_client.get_endpoint(name=name),
+        )
+
```
sasha-gitg, line 468: Preference to block here and add async support later.

sirtorry: makes sense and will be cleaner for operations that require endpoint.

google/cloud/aiplatform/models.py
sirtorry: if we do this, is there still a reason to have `_is_client_prediction_client`?

sasha-gitg: I agree that we most likely do not need that flag.

google/cloud/aiplatform/models.py
```diff
@@ -215,10 +228,483 @@ def upload(
...
+                max_replica_count=max_replica_count,
+            )
+            deployed_model = gca_endpoint.DeployedModel(
+                automatic_resources=automatic_resources,
+                model=model.resource_name,
+                display_name=deployed_model_display_name,
+            )
+
+        if traffic_split is None:
+            if bool(self._gca_resource.traffic_split) is False:
```
sasha-gitg, line 638: Will self._gca_resource.traffic_split not automatically evaluate to False without the need of type casting?

Also this is equivalent to `if not bool(self._gca_resource.traffic_split)`

google/cloud/aiplatform/models.py
```diff
@@ -215,10 +228,483 @@ def upload(
...
+                raise ValueError(
+                    """There are currently no deployed models so the traffic percentage
```
sasha-gitg, line 640: Shouldn't we need to check that traffic percentage is not 100 first?

google/cloud/aiplatform/models.py
```diff
@@ -215,10 +228,483 @@ def upload(
...
+        #     TODO: check traffic split provided is valid
+
+        operation_future = self.endpoint_client.deploy_model(
+            endpoint=self.resource_name,
+            deployed_model=deployed_model,
+            traffic_split=traffic_split,
+            metadata=metadata,
+        )
+
+        return lro.LRO(operation_future)
```
sasha-gitg, line 657: Block before returning. I also think we don't need to pass back the LRO. This can just return None.

google/cloud/aiplatform/models.py
```diff
@@ -215,10 +228,483 @@ def upload(
...
+            )
+
+        operation_future = self.endpoint_client.undeploy_model(
+            endpoint=self.resource_name,
+            deployed_model_id=deployed_model_id,
+            traffic_split=traffic_split,
+            metadata=metadata,
+        )
+
+        return lro.LRO(operation_future)
```
sasha-gitg, line 702: Block and probably no need to return the LRO.

google/cloud/aiplatform/models.py
```diff
@@ -215,10 +228,483 @@ def upload(
...
 
+    def predict(self, instances: List[Dict], parameters: Optional[Dict]) -> List[Dict]:
```
sasha-gitg, line 704: Please capture as bug and TODO with ticket number.

tests/unit/aiplatform/test_models.py
```diff
@@ -357,3 +358,19 @@ def test_upload_uploads_and_gets_model_with_custom_location(self):
...
+            initializer.global_config, "create_client"
+        ) as create_client_mock:
+            api_client_mock = mock.Mock(spec=EndpointServiceClient)
+            create_client_mock.return_value = api_client_mock
+
+            test_model = models.Model(_TEST_MODEL_NAME)
+            mock_endpoint = mock.Mock(autospec=models.Endpoint)
+
+            assert test_model.deploy(mock_endpoint) == mock_endpoint
+            mock_endpoint.deploy.assert_called_once_with()
```
sasha-gitg, line 377: Why isn't this called with arguments?

tests/unit/aiplatform/test_models.py
```diff
@@ -357,3 +358,19 @@ def test_upload_uploads_and_gets_model_with_custom_location(self):
             api_client_mock.get_model.assert_called_once_with(
                 name=test_model_resource_name
             )
+
+    def test_deploy():
```
sasha-gitg, line 364: self

sasha-gitg (changes requested):

google/cloud/aiplatform/models.py
```diff
@@ -215,10 +228,483 @@ def upload(
...
         )
 
-    # TODO(b/169782716) add support for deployment when Endpoint class complete
-    def deploy(self):
-        raise NotImplementedError("Deployment not implemented.")
+    def deploy(
+        self,
+        endpoint: Optional["Endpoint"] = None,
+        deployed_model_display_name: Optional[str] = None,
+        traffic_percentage: Optional[int] = 100,
```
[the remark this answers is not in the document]
sirtorry: what should be done when there's no pre-existing deployed model? my understanding is traffic split should sum to 0 or not exist.

google/cloud/aiplatform/models.py
```diff
@@ -215,10 +228,483 @@ def upload(
...
+        traffic_split: Optional[Dict] = None,
+        machine_type: Optional[str] = None,
+        min_replica_count: Optional[int] = 0,
```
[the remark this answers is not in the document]
sirtorry: i believe they might. i'll make this 1 for now.

google/cloud/aiplatform/models.py
```diff
@@ -215,10 +228,483 @@ def upload(
...
+
+        created_endpoint = None
+        endpoint = cls(
+            endpoint=created_endpoint,
+            project=project,
+            location=location,
+            credentials=credentials,
+        )
+
+        endpoint_client.update_resource(
```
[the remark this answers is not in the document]
sirtorry: was supposed to be create_endpoint_operation (lro.LRO)

google/cloud/aiplatform/models.py
```diff
@@ -215,10 +228,483 @@ def upload(
...
+            ),
+            display_name=display_name,
+            description=description,
+            labels=labels,
+            metadata=metadata,
+        )
+
+        created_endpoint = None
+        endpoint = cls(
+            endpoint=created_endpoint,
```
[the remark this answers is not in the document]
sirtorry: redid constructor.

google/cloud/aiplatform/models.py
```diff
@@ -215,10 +228,483 @@ def upload(
...
+            traffic_split (Dict):
+                Traffic split to use.
+        """
+        old_models_traffic = 100 - traffic_percentage
+        if old_models_traffic:
+            unallocated_traffic = old_models_traffic
+            for deployed_model in traffic_split:
+                current_traffic = traffic_split[deployed_model]
+                new_traffic = int(current_traffic / 100 * old_models_traffic)
+                traffic_split[deployed_model] = new_traffic
```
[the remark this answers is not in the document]
sirtorry: ack. will return new dict.

google/cloud/aiplatform/models.py
```diff
@@ -215,10 +228,483 @@ def upload(
...
+                unallocated_traffic -= new_traffic
+            # will likely under-allocate. make total 100.
+            traffic_split[deployed_model] += unallocated_traffic
```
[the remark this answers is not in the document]
sirtorry: reimplemented as suggested.

google/cloud/aiplatform/models.py
sirtorry: why block? will we change this in the future?

google/cloud/aiplatform/models.py
sirtorry: why block? will we change this in the future?

google/cloud/aiplatform/models.py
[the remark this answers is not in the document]
sasha-gitg: Good point. If the endpoint has no deployed model, it's safe to provide send 100% traffic to the model.

google/cloud/aiplatform/models.py
sasha-gitg: We will change it in the future. We should keep the experience consistent across resource nouns for now and add async consistently across the SDK.

google/cloud/aiplatform/models.py
sasha-gitg: Acked and reasoned above.

tests/unit/aiplatform/test_models.py
sirtorry: ack. it shouldn't have been.

google/cloud/aiplatform/models.py
sirtorry: will block.

google/cloud/aiplatform/models.py
sirtorry: address along with predict PR.

google/cloud/aiplatform/models.py
```diff
@@ -215,10 +228,534 @@ def upload(
...
+            ),
+            display_name=display_name,
+            description=description,
+            labels=labels,
+            metadata=metadata,
+        )
+
+        endpoint = cls(project=project, location=location, credentials=credentials)
+
+        lro.LRO.update_resource(
```
sasha-gitg, line 427: To me it seems like there is bug here because this is not a staticmethod:
https://github.com/googleapis/python-aiplatform/blob/dev/google/cloud/aiplatform/lro.py#L59

I think `LRO.update_resource` needs to be changed to take self and not an operation future.

And then this should be `create_endpoint_operation.update_resource` without operation future passed in.

google/cloud/aiplatform/models.py
```diff
@@ -215,10 +228,534 @@ def upload(
...
+            operation_future=create_endpoint_operation.operation_future,
+            resource_noun_obj=endpoint,
+            result_key="name",
+            api_get=lambda name: api_client.get_endpoint(name=name),
+        )
+
```
sasha-gitg, line 433: Please block here. `create_endpoint_operation.operation_future.result()` should work.

google/cloud/aiplatform/models.py
```diff
@@ -215,10 +228,534 @@ def upload(
...
+        old_models_traffic = 100 - traffic_percentage
+        if old_models_traffic:
+            unallocated_traffic = old_models_traffic
+            for deployed_model in traffic_split:
+                current_traffic = traffic_split[deployed_model]
+                new_traffic = int(current_traffic / 100 * old_models_traffic)
+                new_traffic_split[deployed_model] = new_traffic
+                unallocated_traffic -= new_traffic
+            # will likely under-allocate. make total 100.
+            for deployed_model in new_traffic_split:
```
sasha-gitg, line 517: Will `unallocated_traffic` <= len(deployed_models) always be True?

google/cloud/aiplatform/models.py
```diff
@@ -215,10 +228,534 @@ def upload(
...
+
+
+class Endpoint(base.AiPlatformResourceNoun):
+
+    client_class = EndpointServiceClient
+    _is_client_prediction_client = False
+
+    def __init__(
+        self,
+        endpoint_name: Optional[gca_endpoint.Endpoint] = None,
```
sasha-gitg, line 317: I understand this was set as Optional to be able to create the class while the operations future was still executing. The issue I see is that the user will see this and think that they can get an endpoint with no resource name.

1. I think it's safer in this PR to require this argument and block during create and instantiate the class after the LRO is complete (so no need to pass in None).

2. In the future when we provide async support we should rely on an alternative constructor to initialize the endpoint to None so we don't surface that implementation detail on our APIs surface to users. (doesn't need to be added to this PR).

google/cloud/aiplatform/models.py
```diff
@@ -215,10 +228,534 @@ def upload(
...
+                        percentage for this deployed model needs to be 100."""
+                    )
+            traffic_split = self._allocate_traffic(
+                traffic_split=self._gca_resource.traffic_split,
+                traffic_percentage=traffic_percentage,
+            )
+        else:
+            traffic_sum = 0
+            for deployed_model in traffic_split:
+                # TODO verify every referenced deployed model exists
```
sasha-gitg, line 678: Please add a buganizer ticket.

google/cloud/aiplatform/models.py
```diff
@@ -215,10 +228,534 @@ def upload(
...
+            traffic_split = self._unallocate_traffic(
+                traffic_split=self._gca_resource.traffic_split,
+                deployed_model_id=deployed_model_id,
+            )
+        else:
+            if traffic_split[deployed_model_id]:
+                raise ValueError("Model being undeployed should have 0 traffic.")
+            traffic_sum = 0
+            for deployed_model in traffic_split:
+                # TODO verify every referenced deployed model exists
```
sasha-gitg, line 734: Ticket here as well please.

google/cloud/aiplatform/models.py
sirtorry: it should. int/floor/truncating a number in division should result in <1 difference per number.

google/cloud/aiplatform/models.py
sirtorry: will block for now

sasha-gitg: Also, this should be Optional[str].

google/cloud/aiplatform/training_jobs.py
```diff
@@ -254,7 +309,9 @@ def package_and_copy(self, copy_method: Callable[[str], str]) -> str:
 
         with tempfile.TemporaryDirectory() as tmpdirname:
             source_distribution_path = self.make_package(tmpdirname)
-            return copy_method(source_distribution_path)
+            output_location = copy_method(source_distribution_path)
+            _LOGGER.info("Training script copied to:\n%s." % output_location)
```
ivanmkc, line 313: Nice touch, that's useful

sirtorry:
@sasha-gitg all remaining test issues are now related the predict.

sasha-gitg (approved):

PR #39, opened 2020-11-03
feat: python-aiplatform beta library

dizcology:
Generated library version 0.3.0 for google-cloud-aiplatform.


setup.py
```diff
@@ -15,39 +15,31 @@
...
+    url="https://github.com/googleapis/python-aiplatform",
     platforms="Posix; MacOS X; Windows",
     include_package_data=True,
     install_requires=(
         "google-api-core[grpc] >= 1.22.2, < 2.0.0dev",
         "libcst >= 0.2.5",
-        "proto-plus >= 1.4.0",
+        "proto-plus >= 1.10.1",
+        "mock >= 4.0.2",
+        "google-cloud-storage >= 1.26.0",
```
busunkim96, line 37: Please keep an upper bound  `<2.0.0dev`

setup.py
```diff
@@ -15,39 +15,31 @@
...
     ),
     python_requires=">=3.6",
+    scripts=["scripts/fixup_aiplatform_v1beta1_keywords.py",],
```
busunkim96, line 40: This is unneeded since this library was never released with the older generator. Please can add this file to the `excludes` and remove this line. You can also drop the dependency on `libcst`.

setup.py
```diff
@@ -15,39 +15,31 @@
...
     classifiers=[
-        "Development Status :: 4 - Beta",
+        "Development Status :: 3 - Alpha",
```
busunkim96, line 42: Quick check to make sure this is intentional. If it is, please also lower the release level in `.repo-metadata.json`.

dizcology: Should be Beta - corrected.

busunkim96 (approved):

PR #42, opened 2020-11-04
feat: Add Custom Training by Python Script

sasha-gitg:
Adds CustomTrainingJob Class.

The current implementation splits the constructing of the CustomTrainingJob and running of the Job. Note that no Custom Training service resource is created when the object is constructed. It is created at the `run` call. Which can lead to issues if the correct args were not passed in during construction. This mainly is an issue for model serving args as we require the model serving environment definitions at construction time, which should make sense because it's tightly coupled to the script provided at construction time. An alternative is to move this to the run call.

An additional alternative is to add all constructor and run args to the constructor. And only have the model accessible through the get_model API:

```python
job = CustomTrainingJob('test_script.py', ...)  # launch job here
my_model = job.get_model()
```  

Instead of the current flow:

```python
job = CustomTrainingJob('test_script.py', ...)
model = job.run(my_dataset, ...)
```

Fixes [b/169779290](b/169779290) 🦕

### DIFF
--- a/google/cloud/aiplatform/training_jobs.py
+++ b/google/cloud/aiplatform/training_jobs.py
@@ -17,20 +17,69 @@
 
 import datetime
 import functools
+import logging
 import pathlib
 import shutil
 import subprocess
 import sys
 import tempfile
-from typing import Callable, Dict, Optional, Sequence
+import time
+from typing import Callable, Dict, List, Optional, Sequence
 
 
 from google.auth import credentials as auth_credentials
 from google.cloud.aiplatform import base
 from google.cloud.aiplatform import datasets
+from google.cloud.aiplatform import initializer
 from google.cloud.aiplatform import models
+from google.cloud.aiplatform import schema
+from google.cloud.aiplatform import utils
+from google.cloud.aiplatform_v1beta1 import AcceleratorType
+from google.cloud.aiplatform_v1beta1 import CustomJobSpec
+from google.cloud.aiplatform_v1beta1 import FractionSplit
+from google.cloud.aiplatform_v1beta1 import GcsDestination
+from google.cloud.aiplatform_v1beta1 import InputDataConfig
+from google.cloud.aiplatform_v1beta1 import JobServiceClient
+from google.cloud.aiplatform_v1beta1 import Model
+from google.cloud.aiplatform_v1beta1 import ModelContainerSpec
 from google.cloud.aiplatform_v1beta1 import PipelineServiceClient
+from google.cloud.aiplatform_v1beta1 import PipelineState
+from google.cloud.aiplatform_v1beta1 import TrainingPipeline
 from google.cloud import storage
+from google.protobuf import json_format
+from google.protobuf.struct_pb2 import Value
+from google.rpc import code_pb2
+
+logging.basicConfig(level=logging.INFO, stream=sys.stdout)
+_LOGGER = logging.getLogger(__name__)
+
+_PIPELINE_COMPLETE_STATES = set(
+    [
+        PipelineState.PIPELINE_STATE_SUCCEEDED,
+        PipelineState.PIPELINE_STATE_FAILED,
+        PipelineState.PIPELINE_STATE_CANCELLED,
+        PipelineState.PIPELINE_STATE_PAUSED,
+    ]
+)
+
+
+def _timestamped_gcs_dir(root_gcs_path: str, dir_name_prefix: str) -> str:
+    """Composes a timestamped GCS directory.
+
+    Args:
+        root_gcs_path: GCS path to put the timestamped directory.
+        dir_name_prefix: Prefix to add the timestamped directory.
+    Returns:
+        Timestamped gcs directory path in root_gcs_path.
+    """
+    timestamp = datetime.datetime.now().isoformat(sep="-", timespec="milliseconds")
+    dir_name = "-".join([dir_name_prefix, timestamp])
+    if root_gcs_path.endswith("/"):
+        root_gcs_path = root_gcs_path[:-1]
+    gcs_path = "/".join([root_gcs_path, dir_name])
+    if not gcs_path.startswith("gs://"):
+        return "gs://" + gcs_path
+    return gcs_path
 
 
 def _timestamped_copy_to_gcs(
@@ -151,6 +200,9 @@ setup(
         "{python_executable} setup.py sdist --formats=gztar"
     )
 
+    # Module name that can be executed during training. ie. python -m
+    module_name = f"{_ROOT_MODULE}.{_TASK_MODULE_NAME}"
+
     def __init__(self, script_path: str, requirements: Optional[Sequence[str]] = None):
         """Initializes packager.
 
@@ -254,7 +306,9 @@ setup(
 
         with tempfile.TemporaryDirectory() as tmpdirname:
             source_distribution_path = self.make_package(tmpdirname)
-            return copy_method(source_distribution_path)
+            output_location = copy_method(source_distribution_path)
+            _LOGGER.info("Training script copied to:\n%s." % output_location)
+            return output_location
 
     def package_and_copy_to_gcs(
         self,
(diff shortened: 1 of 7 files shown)


google/cloud/aiplatform/training_jobs.py
```diff
@@ -17,20 +17,67 @@
...
-from typing import Callable, Dict, Optional, Sequence
+import time
+from typing import Callable, Dict, List, Optional, Sequence
 
 
 from google.auth import credentials as auth_credentials
 from google.cloud.aiplatform import base
 from google.cloud.aiplatform import datasets
+from google.cloud.aiplatform import initializer
 from google.cloud.aiplatform import models
```
dizcology, line 34: Do we have a general guideline for when to import modules (e.g. `from google.cloud.aiplatform import models` here) versus when to import classes (e.g. `from google.cloud.aiplatform_v1beta1 import Model` below)?  

if there are simple rules of deciding this, please add to the contributing guide file for reference.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -281,46 +333,474 @@ def package_and_copy_to_gcs(
...
+        base_output_dir: Optional[str] = None,
         args: Optional[Dict[str, str]] = None,
-        replica_count=0,
-        machine_type="n1-standard-4",
-        accelerator_type=None,
-        accelerator_count=0,
-        spec_yaml_path: Optional[str] = None,
-    ) -> models.Model:
-        pass
+        replica_count: str = 0,
```
dizcology, line 457: the type perhaps should be `int` here.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -281,46 +333,474 @@ def package_and_copy_to_gcs(
...
+        machine_type: str = "n1-standard-4",
+        accelerator_type: str = "ACCELERATOR_TYPE_UNSPECIFIED",
```
dizcology, line 459: This is fine, but perhaps consider directly using the existing GAPIC Enums for this?

google/cloud/aiplatform/training_jobs.py
```diff
@@ -281,46 +333,474 @@ def package_and_copy_to_gcs(
...
+                os.environ["AIP_DATA_FORMAT"]
+            model_display_name (str):
+                If the script produces a managed AI Platform Model. The display name of
+                the Model. The name can be up to 128 characters long and can be consist
+                of any UTF-8 characters.
+            base_output_dir (str):
+                GCS output directory of job. If not provided a
+                timestamped directory in the staging directory will be used.
+            args (Dict[str, str]):
+                Key-value pair or command line arguments to be passed to the Python
```
dizcology, line 494: Maybe some more explanations here, since `format_args` does something implicitly.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -281,46 +333,474 @@ def package_and_copy_to_gcs(
...
+        def format_args(args: Dict[str, str]) -> List[str]:
+            """Formats key-value arguments.
+
+            Args:
+                args (Dict[str, str]):
+                    Required: key-value pair or arguments.
+            Returns:
+                arg_string: List of formated arguments.
+            """
+            return [f"--{key}={value}" for key, value in args.items()]
```
dizcology, line 535: In particular, the user-provided package should not accept positional arguments?

google/cloud/aiplatform/training_jobs.py
```diff
@@ -281,46 +333,474 @@ def package_and_copy_to_gcs(
...
+            )
+        return model
+
+    def _sync_gca_resource(self):
+        """Helper method to sync the local gca_source against the service."""
+        self._gca_resource = self.api_client.get_training_pipeline(
+            name=self.resource_name
+        )
+
+    def _block_until_complete(self):
```
dizcology, line 689: for later: This (and maybe other) helper methods might be abstracted to deal with other kinds of `Job`s e.g. batch prediction job.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -281,46 +333,474 @@ def package_and_copy_to_gcs(
...
+            wait = min(wait * multiplier, max_wait)
+
+        self._raise_failure()
+
+        if self._gca_resource.model_to_upload and not self.is_failed:
+            _LOGGER.info(
+                "Model available at %s" % self._gca_resource.model_to_upload.name
+            )
+
+    def _raise_failure(self):
```
dizcology, line 713: The service does not gracefully do this, but it would be nice if the library informs the user of what/where intermediate things were created (service resources, GCS objects, etc) so they know what needs to be cleaned up.  This is perhaps useful for both success and failure cases, but seems more important in the case of failure cases.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -281,46 +333,474 @@ def package_and_copy_to_gcs(
...
+            )
+
+    @property
+    def state(self) -> PipelineState:
+        """Current training state."""
+        self._assert_has_run()
+        return self._gca_resource.state
+
+    @property
+    def is_failed(self) -> bool:
```
dizcology, line 793: maybe `has_failed`.

tests/unit/aiplatform/test_lro.py
```diff
@@ -26,6 +30,18 @@
 
 
 TEST_OPERATION_NAME = "test/operation"
+_TEST_PROJECT = "test-project"
+
+
+def setup_module(module):
```
dizcology, line 36: is this used?

google/cloud/aiplatform/training_jobs.py
sasha-gitg: If there is no standard defined then we should go with the go/pystyle#imports.

I'll make changes in this PR and we should ensure future PRs follow that style.

I also have a preference to reduce levels of indirection of imports as well so we should be importing directly from the module that contains the implementation.

@vinnysenthil 
@sirtorry 
@ivanmkc

tests/unit/aiplatform/test_lro.py
sasha-gitg: https://docs.pytest.org/en/stable/xunit_setup.html#module-level-setup-teardown

Why don't you think so?

google/cloud/aiplatform/training_jobs.py
sasha-gitg: Done.

google/cloud/aiplatform/training_jobs.py
sasha-gitg: I went with this because we don't seem to have an enum for machine type so it seems inconsistent to require a string for one hardware type and an enum for another. Though, as we receive feedback we can switch.

google/cloud/aiplatform/training_jobs.py
sasha-gitg: Agree.

google/cloud/aiplatform/training_jobs.py

google/cloud/aiplatform/training_jobs.py
sasha-gitg: Changed to args and accept a list to support positional args and flags.

google/cloud/aiplatform/training_jobs.py
sasha-gitg: Acked, resolved, and explained above.

google/cloud/aiplatform/training_jobs.py
sasha-gitg: We log all these items during the run.

tests/unit/aiplatform/test_lro.py
dizcology: I see - thanks.  I have not used this before.

tests/unit/aiplatform/test_training_jobs.py
```diff
@@ -31,6 +62,33 @@
...
+_TEST_MODEL_DISPLAY_NAME = "model-display-name"
+_TEST_TRAINING_FRACTION_SPLIT = 0.6
+_TEST_VALIDATION_FRACTION_SPLIT = 0.2
+_TEST_TEST_FRACTION_SPLIT = 0.2
+
+_TEST_OUTPUT_PYTHON_PACKAGE_PATH = "gs://test/ouput/python/trainer.tar.gz"
+
+_TEST_MODEL_NAME = "projects/my-project/locations/us-central1/models/12345"
+
+_TEST_PIPELINE_RESOURCE_NAME = (
```
dizcology, line 88: for later: all of these are getting a little heavy and difficult to manage.  we should at least put all of these pieces in a separate file so they are shared across tests.  further down the road we might need to modify the approach.

dizcology (approved):

PR #45, opened 2020-11-06
feat: Refactor Dataset class to decouple from GCS and BQ logic

ivanmkc:
Design proposal here: [Model Builder SDK Dataset Source Refactor Proposal](https://docs.google.com/document/d/1_KoOm42mdjHp-X_QSgMWWCAYji5Tdl_9SYQ6RakJoas/edit?usp=sharing)

The text below is pretty much the same as the proposal, with some more details on things that are WIP.

## Problem
Due to idiosyncrasies in how the underlying Gapic API and services treat BQ and GCS inputs, there are many pitfalls for creating a Dataset using `Dataset.create`.

```
    @classmethod
    def create(
        cls,
        display_name: str,
        metadata_schema_uri: str,
        gcs_source: Optional[Sequence[str]] = None,
        bq_source: Optional[str] = None,
        import_schema_uri: Optional[str] = None,
        metadata: Sequence[Tuple[str, str]] = (),
        labels: Optional[Dict] = None,
        data_items_labels: Optional[Dict] = None,
    ) -> "Dataset":
```

Here is a taste of the current valid/invalid parameter combinations for creating a Dataset:
| metadata_schema_uri | gcs_source | bq_source | import_schema_uri | data_items_labels | Is this valid?                                                  |
|---------------------|------------|-----------|-------------------|-------------------|-----------------------------------------------------------------|
| tabular             | present    | None      | present           | present           | No, tabular doesn't use import_schema_uri and data_items_labels |
| tabular             | None       | present   | present           | present           | No, tabular doesn't use import_schema_uri and data_item_labels  |
| non-tabular         | present    | None      | present           | present           | Yes                                                             |
| tabular             | None       | present   | None              | None              | Yes                                                             |
| non-tabular         | *          | present   | *                 | *                 | No, BQ only works with tabular                                  |
| *                   | present    | present   | *                 | *                 | No, BQ and GCS cannot both be set at the same time              |
| non-tabular         | present    | None      | None              | None              | No, import_schema_uri is required for GCS source                |
| non-tabular         | None       | None      | None              | None              | Yes, nothing is imported to this newly created dataset          |
| text         | present       | None      | None              | image_classification_multi_label              | No, text is not compatible with image_classification_multi_label         

Sure, we can include documentation to tell the user but IMO it's not a great user experience when there are so many potential pitfalls. Many of the invalid states also fail silently or fail later on during the model training step. This ambiguity doesn't seem to be what the EAFP principle was intended for.

We could also write  a large amount of initialization-time checks, but they'd be tougher to maintain and debug then the alternative way proposed below. That also wouldn't solve problems with coupling and extensibility.

## Proposal
Instead of having a bunch of loosely-typed and sometimes optional parameters, I propose that we encapsulate this logic into a few classes.
The 4 classes (`EmptyNonTabularSourceConfig`, `BQTabularSourceConfig`, `GCSTabularSourceConfig` and `GCSNonTabularSourceConfig`) encapsulate the GCS/BQ specific logic and **guarantees** that if a user passes in the right types to one of these, a valid Dataset can be created.

Note: It's possible to combine `GCSTabularSourceConfig` and `GCSNonTabularSourceConfig` if people prefer that. It'll just mean that there needs to be validation to make sure `import_schema_uri = None` and `data_items_labels = None` when `metadata_schema_uri = tabular`

For example, `BQTabularSourceConfig` for BigQuery tabular datasets has this signature, only taking in the parameter it needs:

```
class BQTabularSourceConfig(SourceConfig):
    def __init__(self, source_uri: str):
        ... validation for URI done here ...
```

The new `Dataset.create` signature would thus look like:
```
@classmethod
    def create(
        cls,
        display_name: str,
        source: SourceConfig,
        request_metadata: Sequence[Tuple[str, str]] = (),
        labels: Optional[Dict] = None,
    ) -> "Dataset":
```

It may seem like we're adding a 4 new user-facing classes and increasing complexity but the fact is that this complexity in parameter combinations already exists in Dataset.create(...) as can be seen in the above table. It is just currently hidden behind Optional's.

Since this is Python, the type system will only check that the number of parameters are correct. However, this is still an improvement since the function signature tells the user exactly what are valid combinations. If users use something like MyPy for static type checking, they will get additional type safety. 

## Additional benefits
- Decoupling: This PR removes GCS and BQ implementation details from Dataset's `create` and `import_data` functions, resulting in cleaner logic, since it all gets encapsulated into the SourceConfig implementation details. One example is the logic to check that GCS URI's start with "gs://".
- Extensibility: Inheriting the `SourceConfig` abstract class makes it extensible to other source types besides GCS and BQ.

## How would this work with subclassing Dataset to create datatype specific subclasses?
- Without this proposal, just creating subclasses of Dataset makes things easier for users but leaves a few problems:
  - Unless we make `Dataset.create` completely private, users will still be able to call the problematic `Dataset.create` function.
  - Even if we make it private, the SDK devs will still have to deal with the problematic `Dataset.create` function.
  - Without decoupling GCS and BQ logic from Datasets, that logic is scattered everywhere in the Datasets.py codebase, making things hard to change. In the future, it'll also be tough to extend to other datatypes that don't directly leverage GCS and BQ.

With this proposal, it's still easy to create subclasses. Here's an example:

```
class ImageDataset(Dataset):
    @classmethod
    def create(
        cls,
        display_name: str,
        gcs_source_uris: [str],
        import_schema_uri: str, 
        data_items_labels: Optional[Dict] = None,
        request_metadata: Sequence[Tuple[str, str]] = (),
        labels: Optional[Dict] = None,
    ) -> "Dataset":
        # TODO: Add validation to check that the `import_schema_uri` is valid for an image dataset
        source = GCSNonTabularSourceConfig(source_uris=gcs_source_uris, metadata_schema_uri=schema.dataset.metadata.image, import_schema_uri=import_schema_uri, data_items_labels=data_items_labels)
        GCSSourceConfig.__init__(self, display_name=display_name, source=source)
```

## Further recommended refactoring
To completely decouple `Dataset` from GCS, we'd have to remove the dependency on GCSDestination. This can be done as part of `(b/171311614): Add support for BiqQuery export path`.

## WIP
- Add all doc strings.
- Move SourceConfig implementations to their own files and folder structure.
- Add unit-tests for new classes.
- Find a better name than `SourceConfig`

### DIFF
--- a/google/cloud/aiplatform/data_source.py
+++ b/google/cloud/aiplatform/data_source.py
@@ -3,17 +3,8 @@ from abc import ABC, abstractmethod
 from google.cloud.aiplatform import schema
 from google.cloud.aiplatform_v1beta1 import ImportDataConfig
 from google.cloud.aiplatform_v1beta1 import GcsSource
-
-class DataImportable(ABC):
-    """
-        An abstract class that provides import_data_config for importing data to an existing database
-    """
-
-    @property 
-    def import_data_config(self) -> ImportDataConfig:
-        raise NotImplementedError
         
-class SourceConfig(ABC):
+class DataSource(ABC):
     @property
     def metadata_schema_uri(self) -> str:
         raise NotImplementedError
@@ -22,15 +13,14 @@ class SourceConfig(ABC):
     def metadata(self) -> Dict:
         raise NotImplementedError
 
-## TODO: Move to a EmptySourceConfig file
-class EmptyNonTabularSourceConfig(SourceConfig):
+class EmptyNonTabularDataSource(DataSource):
     """Class for a empty, non-tabular dataset schema that provides Dataset metadata
-    Used with Dataset.create(...) to create an empty, non-tabular dataset.
+    
+    To be used with the Dataset.create function to create an empty, non-tabular dataset.
     """
 
     def __init__(self, metadata_schema_uri: str):
-        """TODO
-
+        """
         Args:
             metadata_schema_uri (str):
                 Required. Points to a YAML file stored on Google Cloud Storage
@@ -50,16 +40,22 @@ class EmptyNonTabularSourceConfig(SourceConfig):
     def metadata(self) -> Dict:
         return {}
 
-## TODO: Move to a BQSourceConfig file
-class BQTabularSourceConfig(SourceConfig):
+class BQTabularDataSource(DataSource):
     """Class for a tabular dataset schema that provides Dataset metadata
     
-    Used for data storaged on BigQuery.
-    When used with Dataset.create(...), the data will be imported to the dataset at creation time.
+    Used for data stored on BigQuery.
+    To be used with the Dataset.create function to import the dataset at creation time.
     """
 
     def __init__(self, source_uri: str):
-        # TODO: Add doc strings
+        """
+        Args:
+            source_uris (Sequence[str]):
+                Required. Google Cloud Storage URI(-s) to the
+                input file(s). May contain wildcards. For more
+                information on wildcards, see
+                https://cloud.google.com/storage/docs/gsutil/addlhelp/WildcardNames.
+        """
 
         # Perform validation of URI
         if not (source_uri.startswith('bq://') or source_uri.startswith('bigquery://')):
@@ -78,12 +74,20 @@ class BQTabularSourceConfig(SourceConfig):
     def metadata(self) -> Dict:
         return {"input_config": {"bigquery_source": {"uri": self._source_uri}}}
 
-## TODO: Move to a GCSSourceConfig file
 # Do not instantiate this abstract class
-class GCSSourceValidating(SourceConfig):
+class GCSSourceValidating(DataSource):
     """Abstract class that takes in GCS source_uri's and performs validation"""
 
-    def __init__(self, source_uris: [str]):        
+    def __init__(self, source_uris: [str]):
+        """
+        Args:
+            source_uris (Sequence[str]):
+                Required. Google Cloud Storage URI(-s) to the
+                input file(s). May contain wildcards. For more
+                information on wildcards, see
+                https://cloud.google.com/storage/docs/gsutil/addlhelp/WildcardNames.
+        """
+
         # Perform validation of URI's
         if not all([uri.startswith('gs://') for uri in source_uris]):
             raise ValueError(
@@ -93,18 +97,18 @@ class GCSSourceValidating(SourceConfig):
         self._source_uris = source_uris
         
 # Do not instantiate this abstract class
-class GCSSourceConfig(GCSSourceValidating, SourceConfig):
+class GCSDataSource(GCSSourceValidating, DataSource):
     """Abstract class for GCS source_uri's that provides Dataset metadata"""
 
     @property
     def metadata(self) -> Dict:
         return {"input_config": {"gcs_source": {"uri": self._source_uris}}}
 
-class GCSTabularSourceConfig(GCSSourceConfig):
+class GCSTabularDataSource(GCSDataSource):
     """Class for a tabular dataset schema that provides Dataset metadata. 
     
     Used for CSV files on Google Cloud Storage.
-    When used with Dataset.create(...), the data will be imported to the dataset at creation time.
+    To be used with the Dataset.create function to import the dataset at creation time.
     """
 
     @property
@@ -112,14 +116,22 @@ class GCSTabularSourceConfig(GCSSourceConfig):
         # Always return tabular as this config only supports tabular
         return schema.dataset.metadata.tabular
 
-class GCSNonTabularImportConfig(GCSSourceValidating, DataImportable):
-    """Class for a non-tabular dataset schema that provides import metadata. 
-    Used with Dataset.import_data(...) to import files from Google Cloud Storage to an existing dataset.
+class DataImportable(ABC):
+    """An abstract class that provides import_data_config for importing data to an existing database"""
+
+    @property 
+    def import_data_config(self) -> ImportDataConfig:
+        raise NotImplementedError
+    
+class GCSNonTabularImportDataSource(GCSSourceValidating, DataImportable):
+    """Class for a non-tabular dataset schema that provides import metadata.
+
+    Used for files on Google Cloud Storage.
+    To be used with the Dataset.import_data function to import files from Google Cloud Storage to an existing dataset.
     """
 
     def __init__(self, source_uris: [str], import_schema_uri: str, data_items_labels: Optional[Dict] = None):
-        """TODO
-
+        """
         Args:
             source_uris (Sequence[str]):
                 Required. Google Cloud Storage URI(-s) to the
@@ -161,14 +173,15 @@ class GCSNonTabularImportConfig(GCSSourceValidating, DataImportable):
                 data_item_labels=self._data_items_labels,
             )
 
-class GCSNonTabularSourceConfig(GCSSourceConfig, GCSNonTabularImportConfig):
+class GCSNonTabularDataSource(GCSDataSource, GCSNonTabularImportDataSource):
     """Class for a non-tabular dataset schema that provides Dataset metadata. 
-    Used with Dataset.create(...) to import files from Google Cloud Storage to the dataset at creation time.
+    
+    Used for files on Google Cloud Storage.
+    To be used with the Dataset.create function to import data at creation time.
     """
 
     def __init__(self, source_uris: [str], metadata_schema_uri: str, import_schema_uri: str, data_items_labels: Optional[Dict] = None):
-        """TODO
-
+        """
         Args:
             source_uris (Sequence[str]):
                 Required. Google Cloud Storage URI(-s) to the
@@ -206,11 +219,11 @@ class GCSNonTabularSourceConfig(GCSSourceConfig, GCSNonTabularImportConfig):
 
         if metadata_schema_uri == schema.dataset.metadata.tabular:
             raise ValueError(
-                "Expected non-tabular metadata schema uri. For tabular metadata schema, use GCSTabularSourceConfig"
+                "Expected non-tabular metadata schema uri. For tabular metadata schema, use GCSTabularDataSource"
             )
         
-        GCSSourceConfig.__init__(self, source_uris)
-        GCSNonTabularImportConfig.__init__(self, source_uris, import_schema_uri, data_items_labels)
+        GCSDataSource.__init__(self, source_uris)
+        GCSNonTabularImportDataSource.__init__(self, source_uris, import_schema_uri, data_items_labels)
         self._metadata_schema_uri = metadata_schema_uri
 
     @property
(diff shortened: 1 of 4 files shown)


google/cloud/aiplatform/datasets.py
```diff
@@ -96,17 +94,9 @@ def create(
                 Required. The user-defined name of the Dataset.
                 The name can be up to 128 characters long and can be consist
                 of any UTF-8 characters.
-            metadata_schema_uri (str):
```
ivanmkc, line 99: Replace `metadata_schema`, `source`, `source (GCS)`, `bq_source` and `data_items_labels` with `source (SourceConfig interface)`

google/cloud/aiplatform/datasets.py
```diff
@@ -157,21 +132,16 @@ def create(
                 Instantiated representation of the managed dataset resource.
         """
 
-        # Validate that source and import schema are passed together or not at all
```
ivanmkc, line 160: This logic is moved to its relevant `SourceConfig`

google/cloud/aiplatform/datasets.py
```diff
@@ -185,23 +155,20 @@ def create(
             credentials=credentials,
         )
 
-        # If an import source was not provided, return empty created Dataset.
-        if not source:
+        # Import if source is an importable type
```
ivanmkc, line 158: `SourceConfig's` that inherit from `DataImportable` can be imported.

google/cloud/aiplatform/datasets.py
```diff
@@ -248,107 +217,29 @@ def _create(
...
             metadata_schema_uri=metadata_schema_uri,
+            metadata=dataset_metadata,
             labels=labels,
         )
 
         return api_client.create_dataset(
             parent=parent, dataset=gapic_dataset, metadata=request_metadata
         )
 
-    def _import(
```
ivanmkc, line 258: Unnecessary intermediate function

google/cloud/aiplatform/source_config.py
```diff
@@ -0,0 +1,160 @@
...
+class DataImportable(ABC):
+    """
+        An abstract class that provides import_data_config for importing data to an existing database
+    """
+
+    @property 
+    def import_data_config(self) -> ImportDataConfig:
+        raise NotImplementedError
+        
+class SourceConfig(ABC):
```
ivanmkc, line 16: TODO: Doc strings

google/cloud/aiplatform/source_config.py
```diff
@@ -0,0 +1,160 @@
...
+    @property
+    def metadata_schema_uri(self) -> str:
+        raise NotImplementedError
+
+    @property
+    def metadata(self) -> Dict:
+        raise NotImplementedError
+
+## TODO: Move to a EmptySourceConfig file
```
ivanmkc, line 25: What folder structure is suitable?

tests/unit/aiplatform/test_datasets.py
```diff
@@ -30,6 +30,7 @@
 from google.cloud.aiplatform import Dataset
 from google.cloud.aiplatform import initializer
 from google.cloud.aiplatform import schema
+from google.cloud.aiplatform.source_config import BQTabularSourceConfig, GCSNonTabularSourceConfig, EmptyNonTabularSourceConfig
```
ivanmkc, line 33: TODO: Remove import of class

tests/unit/aiplatform/test_datasets.py
```diff
@@ -241,30 +238,14 @@ def test_create_and_import_dataset(self, create_dataset_mock, import_data_mock):
...
     def test_import_data(self, import_data_mock):
         aiplatform.init(project=_TEST_PROJECT)
 
         my_dataset = Dataset(dataset_name=_TEST_NAME)
 
         my_dataset.import_data(
-            gcs_source=_TEST_SOURCE_URI_GCS,
-            import_schema_uri=_TEST_IMPORT_SCHEMA_URI,
-            data_items_labels=_TEST_DATA_LABEL_ITEMS,
+            source=GCSNonTabularSourceConfig(source_uris=[_TEST_SOURCE_URI_GCS], metadata_schema_uri=_TEST_METADATA_SCHEMA_URI_NONTABULAR, import_schema_uri=_TEST_IMPORT_SCHEMA_URI, data_items_labels=_TEST_DATA_LABEL_ITEMS) # TODO: Remove metadata_schema_uri as import step doesn't need it
```
ivanmkc, line 248: Working on this atm.

google/cloud/aiplatform/source_config.py
```diff
@@ -0,0 +1,160 @@
...
+                if their content bytes are identical (e.g. image bytes or
+                pdf bytes). These labels will be overridden by Annotation
+                labels specified inside index file refenced by
+                [import_schema_uri][google.cloud.aiplatform.v1beta1.ImportDataConfig.import_schema_uri],
+                e.g. jsonl file.
+        """
+
+        if metadata_schema_uri == schema.dataset.metadata.tabular:
+            raise ValueError(
+                "Expected non-tabular metadata schema uri. For tabular metadata schema, use GCSTabularSourceConfig"
```
ivanmkc, line 136: Ew, the code smell is strong here. The cleaner way would be to create an interface to differentiate tabular schema types and non-tabular at the type level.

However, since Python won't enforce the types for us anyways, this code might be a better trade-off.

google/cloud/aiplatform/source_config.py
```diff
@@ -0,0 +1,160 @@
...
+            )
+
```
ivanmkc, line 138: We can add checks here to make sure the `import_schema_uri` is valid for the given `metadata_schema_uri`.

E.g. `import_schema_uri = image_classification_multi_label_io_format_1.0.0.yaml` would not be compatible with `metadata_schema_uri = text_1.0.0.yaml`

sirtorry:
Can we get some before and after examples?

ivanmkc:
@sirtorry sorry for late reply, the before/after examples are listed in the attached design doc.

google/cloud/aiplatform/data_source.py
```diff
@@ -0,0 +1,264 @@
...
+
+class DataImportable(abc.ABC):
+    """An abstract class that provides import_data_config for importing data to an existing database"""
+
+    @property
+    def import_data_config(self) -> ImportDataConfig:
+        raise NotImplementedError
+
+
+class GCSNonTabularImportDataSource(GCSSourceValidating, DataImportable):
```
ivanmkc, line 132: Importing a dataset after creation takes different parameters so ImportDataSources are separate classes that conform to `DataImportable`.

ivanmkc:
Renamed xxxSourceConfig to xxxDataSource as that made more sense to me. Feel free to discuss.

google/cloud/aiplatform/data_source.py
```diff
@@ -0,0 +1,264 @@
...
+    @property
+    def metadata(self) -> Dict:
+        return {"input_config": {"bigquery_source": {"uri": self._source_uri}}}
+
+
+# Do not instantiate this abstract class
+class GCSSourceValidating(DataSource):
+    """Abstract class that takes in GCS source_uri's and performs validation"""
+
+    def __init__(self, source_uris: [str]):
```
sasha-gitg, line 85: `Sequence` type hint.

google/cloud/aiplatform/data_source.py
```diff
@@ -0,0 +1,264 @@
...
+
+    def __init__(self, source_uri: str):
+        """
+        Args:
+            source_uris (Sequence[str]):
+                Required. Google Cloud Storage URI(-s) to the
+                input file(s). May contain wildcards. For more
+                information on wildcards, see
+                https://cloud.google.com/storage/docs/gsutil/addlhelp/WildcardNames.
+        """
```
sasha-gitg, line 61: Raises section.

google/cloud/aiplatform/data_source.py
```diff
@@ -0,0 +1,264 @@
...
+
+    def __init__(self, source_uris: [str]):
+        """
+        Args:
+            source_uris (Sequence[str]):
+                Required. Google Cloud Storage URI(-s) to the
+                input file(s). May contain wildcards. For more
+                information on wildcards, see
+                https://cloud.google.com/storage/docs/gsutil/addlhelp/WildcardNames.
+        """
```
sasha-gitg, line 93: Raises section.

google/cloud/aiplatform/data_source.py
```diff
@@ -0,0 +1,264 @@
...
+class GCSNonTabularDataSource(GCSDataSource, GCSNonTabularImportDataSource):
+    """Class for a non-tabular dataset schema that provides Dataset metadata.
+
+    Used for files on Google Cloud Storage.
+    To be used with the Dataset.create function to import data at creation time.
+    """
+
+    def __init__(
+        self,
+        source_uris: [str],
```
sasha-gitg, line 197: Sequence type hint.

google/cloud/aiplatform/data_source.py
```diff
@@ -0,0 +1,264 @@
...
+    @property
+    def metadata_schema_uri(self) -> str:
+        raise NotImplementedError
+
+    @property
+    def metadata(self) -> Dict:
+        raise NotImplementedError
+
+
+class EmptyNonTabularDataSource(DataSource):
```
sasha-gitg, line 18: Can this be the initial class instead of DataSource?

google/cloud/aiplatform/data_source.py
```diff
@@ -0,0 +1,264 @@
+from typing import Optional, Dict
+import abc
+from google.cloud.aiplatform import schema
+from google.cloud.aiplatform_v1beta1 import ImportDataConfig
+from google.cloud.aiplatform_v1beta1 import GcsSource
+
+
+class DataSource(abc.ABC):
+    @property
+    def metadata_schema_uri(self) -> str:
```
sasha-gitg, line 10: `@abc.abstractmethod` on these properties if this is intended to be an ABC.

google/cloud/aiplatform/data_source.py
```diff
@@ -0,0 +1,264 @@
...
+        # Always return tabular as this config only supports tabular
+        return schema.dataset.metadata.tabular
+
+
+class DataImportable(abc.ABC):
+    """An abstract class that provides import_data_config for importing data to an existing dataset"""
+
+    @property
+    def import_data_config(self) -> ImportDataConfig:
```
sasha-gitg, line 128: @abc.abstractmethod is this is intended to be an ABC.

google/cloud/aiplatform/data_source.py
sasha-gitg: This is a strange class because it seems to contain `metadata_schema_uri` and `metadata` as properties but they are not implemented.

sasha-gitg (changes requested):
I wonder if this could be simpler. I think 9 classes is a bit much to represent 4 configurations.

ivanmkc:
Thanks @sasha-gitg, I will fix all the issues and ponder a bit more about consolidating the classes. Will come back to this in a few days after I generate samples for Forecasting.

PR #46, opened 2020-11-09
bug: remove requirement for import_schema_uri when passing in gcs_source

sasha-gitg:


vinnysenthil (approved):
LGTM, we should enforce this restriction for non-Tabular AutoML objectives

ivanmkc (commented):
LGTM as long as the failure case that @vinnysenthil pointed out is accounted for somehow.

If GCS source is provided, is metadata_schema_uri is non-tabular and is not custom training case.

PR #49, opened 2020-11-11
refactor: Pull out reusable code in CustomTrainingJob to use in other training jobs

ivanmkc:
## Summary
There is reusable code in `CustomTrainingJob`, specifically in the `init` and `run` methods. Other training jobs such as `AutoMLTablesTrainingJob` can use it as well.

The code mainly has to do with creating and running a training pipeline.
The leftover code in `CustomTrainingJob` will thus only be related to custom training and not generic pipeline code.

In preparation for: https://b.corp.google.com/issues/172282518

## Remaining questions
~~1. In the `TrainingJob.run` method, `training_fraction_split`, `validation_fraction_split`, `test_fraction_split: float` are all only needed when the `dataset` parameter is also provided. I've grouped this together into `DatasetWithSplits`. I'd want to replace the `CustomTrainingJob.run` parameters with this as well. What do you all think?~~
~~2. Should we put `TrainingJob` into its own file and separate out `CustomTrainingJob` and `AutoMLTablesTrainingJob` into their own? If not, how should we organize things as to not have massive files?~~
~~3. Should I make `TrainingJob` an abstract base class to prevent initialization? I can also do this by overriding `new`:~~
```
class TrainingJob(object):

    def __new__(cls, *args, **kwargs):
        if cls is TrainingJob:
            raise TypeError("base class may not be instantiated")
        return object.__new__(cls, *args, **kwargs)
```
~~4. Docstrings: On which functions are they needed? I'll try to look up the Google guidelines.~~

## TODO
~~- [ ] Write unit tests for the generic `TrainingJob` class.~~
- [x] Run linters

## Testing
Passes all existing unit tests

### DIFF
--- a/google/cloud/aiplatform/training_jobs.py
+++ b/google/cloud/aiplatform/training_jobs.py
@@ -26,6 +26,7 @@ import tempfile
 import time
 from typing import Callable, Dict, List, Optional, NamedTuple, Sequence, Union
 
+import abc
 
 from google.auth import credentials as auth_credentials
 from google.cloud.aiplatform import base
@@ -64,6 +65,312 @@ _PIPELINE_COMPLETE_STATES = set(
         gca_pipeline_state.PipelineState.PIPELINE_STATE_PAUSED,
     ]
 )
+
+
+class _TrainingJob(base.AiPlatformResourceNoun):
+    client_class = pipeline_service_client.PipelineServiceClient
+    _is_client_prediction_client = False
+
+    def __init__(
+        self,
+        display_name: str,
+        project: Optional[str] = None,
+        location: Optional[str] = None,
+        credentials: Optional[auth_credentials.Credentials] = None,
+    ):
+        """Constructs a Training Job.
+
+        Args:
+            display_name (str):
+                Required. The user-defined name of this TrainingPipeline.
+            project (str):
+                Optional project to retrieve model from. If not set, project set in
+                aiplatform.init will be used.
+            location (str):
+                Optional location to retrieve model from. If not set, location set in
+                aiplatform.init will be used.
+            credentials (auth_credentials.Credentials):
+                Optional credentials to use to retrieve the model.
+        """
+        utils.validate_display_name(display_name)
+        
+        super().__init__(project=project, location=location, credentials=credentials)
+        self._display_name = display_name
+        self._project = project
+        self._credentials = credentials
+        self._gca_resource = None
+
+    @property
+    @abc.abstractmethod
+    def _model_upload_fail_string(self) -> str:
+        """Helper property for model upload failure."""
+
+        pass
+
+    @abc.abstractmethod
+    def run(self) -> Optional[models.Model]:
+        """Runs the training job. Should call _run_job internally"""
+        pass
+
+    def _create_input_data_config(
+        self, 
+        dataset: Optional[datasets.Dataset],
+        training_fraction_split: float, 
+        validation_fraction_split: float, 
+        test_fraction_split: float) -> gca_training_pipeline.InputDataConfig:
+
+        """Constructs a input data config to pass to the training pipeline. 
+        Override this to create a custom config
+
+        Args:
+            training_fraction_split (float):
+                The fraction of the input data that is to be
+                used to train the Model. This is ignored if Dataset is not provided.
+            training_fraction_split (float):
+                The fraction of the input data that is to be
+                used to train the Model. This is ignored if Dataset is not provided.
+            validation_fraction_split (float):
+                The fraction of the input data that is to be
+                used to validate the Model. This is ignored if Dataset is not provided.
+            test_fraction_split (float):
+                The fraction of the input data that is to be
+                used to evaluate the Model. This is ignored if Dataset is not provided.
+        """
+        
+        input_data_config = None
+        if dataset:
+            # Create fraction split spec
+            fraction_split = gca_training_pipeline.FractionSplit(
+                training_fraction=training_fraction_split,
+                validation_fraction=validation_fraction_split,
+                test_fraction=test_fraction_split,
+            )
+
+            # create input data config
+            input_data_config = gca_training_pipeline.InputDataConfig(
+                fraction_split=fraction_split,
+                dataset_id=dataset.name,
+            )
+        
+        return input_data_config
+
+    def _run_job(
+        self,
+        training_task_definition: str,
+        training_task_inputs: dict,
+        dataset: Optional[datasets.Dataset],
+        training_fraction_split: float,
+        validation_fraction_split: float,
+        test_fraction_split: float,
+        model: Optional[gca_model.Model] = None,
+    ) -> Optional[models.Model]:
+        """Runs the training job.
+
+        Args:
+            training_task_definition (str):
+                Required. A Google Cloud Storage path to the
+                YAML file that defines the training task which
+                is responsible for producing the model artifact,
+                and may also include additional auxiliary work.
+                The definition files that can be used here are
+                found in gs://google-cloud-
+                aiplatform/schema/trainingjob/definition/. Note:
+                The URI given on output will be immutable and
+                probably different, including the URI scheme,
+                than the one given on input. The output URI will
+                point to a location where the user only has a
+                read access.
+            training_task_inputs (~.struct.Value):
+                Required. The training task's parameter(s), as specified in
+                the
+                ``training_task_definition``'s
+                ``inputs``.
+            dataset (aiplatform.Dataset):
+                Optional. The dataset within the same Project from which data will be used to train the Model. The
+                Dataset must use schema compatible with Model being trained,
+                and what is compatible should be described in the used
+                TrainingPipeline's [training_task_definition]
+                [google.cloud.aiplatform.v1beta1.TrainingPipeline.training_task_definition].
+                For tabular Datasets, all their data is exported to
+                training, to pick and choose from.
+            training_fraction_split (float):
+                The fraction of the input data that is to be
+                used to train the Model. This is ignored if Dataset is not provided.
+            validation_fraction_split (float):
+                The fraction of the input data that is to be
+                used to validate the Model. This is ignored if Dataset is not provided.
+            test_fraction_split (float):
+                The fraction of the input data that is to be
+                used to evaluate the Model. This is ignored if Dataset is not provided.
+            model (~.model.Model):
+                Optional. Describes the Model that may be uploaded (via
+                [ModelService.UploadMode][]) by this TrainingPipeline. The
+                TrainingPipeline's
+                ``training_task_definition``
+                should make clear whether this Model description should be
+                populated, and if there are any special requirements
+                regarding how it should be filled. If nothing is mentioned
+                in the
+                ``training_task_definition``,
+                then it should be assumed that this field should not be
+                filled and the training task either uploads the Model
+                without a need of this information, or that training task
+                does not support uploading a Model as part of the pipeline.
+                When the Pipeline's state becomes
+                ``PIPELINE_STATE_SUCCEEDED`` and the trained Model had been
+                uploaded into AI Platform, then the model_to_upload's
+                resource ``name``
+                is populated. The Model is always uploaded into the Project
+                and Location in which this pipeline is.
+    """
+
+        if self._has_run:
+            raise RuntimeError("Training has already run.")
+
+        input_data_config = self._create_input_data_config(
+            dataset=dataset,
+            training_fraction_split=training_fraction_split, 
+            validation_fraction_split=validation_fraction_split, 
+            test_fraction_split=test_fraction_split)
+
+        # create training pipeline
+        training_pipeline = gca_training_pipeline.TrainingPipeline(
+            display_name=self._display_name,
+            training_task_definition=training_task_definition,
+            training_task_inputs=training_task_inputs,
+            model_to_upload=model,
+            input_data_config=input_data_config,
+        )
+
+        training_pipeline = self.api_client.create_training_pipeline(
+            parent=initializer.global_config.common_location_path(
+                self.project, self.location
+            ),
+            training_pipeline=training_pipeline,
+        )
+
+        self._gca_resource = training_pipeline
+
+        _LOGGER.info("View Training:\n%s" % self._dashboard_uri())
+
+        model = self._get_model()
+
+        if model is None:
+            _LOGGER.warning(
+                "Training did not produce a Managed Model returning None. "
+                + self._model_upload_fail_string
+            )
+        return model
+
+    @property
+    def state(self) -> gca_pipeline_state.PipelineState:
+        """Current training state."""
+        self._assert_has_run()
+        return self._gca_resource.state
+
+    def get_model(self) -> Optional[models.Model]:
+        """AI Platform Model produced by this training, if one was produced.
+
+        Returns:
+            model: AI Platform Model produced by this training or None if a model was
+                not produced by this training.
+        """
+        self._assert_has_run()
+        if not self._gca_resource.model_to_upload:
+            raise RuntimeError(self._model_upload_fail_string)
+
+        return self._get_model()
+
+    def _get_model(self) -> Optional[models.Model]:
+        """Helper method to get and instantiate the Model to Upload.
+
+        Returns:
+            model: AI Platform Model if training succeeded and produced an AI Platform
+                Model. None otherwise.
+
+        Raises:
+            RuntimeError if Training failed.
+        """
+        self._block_until_complete()
+
+        if self.has_failed:
+            raise RuntimeError(
+                f"Training Pipeline {self.resource_name} failed. No model available."
+            )
+
+        if not self._gca_resource.model_to_upload:
+            return None
+
+        if self._gca_resource.model_to_upload.name:
+            fields = utils.extract_fields_from_resource_name(
+                self._gca_resource.model_to_upload.name
+            )
+            return models.Model(
+                fields.id, project=fields.project, location=fields.location
+            )
+
+    def _block_until_complete(self):
+        """Helper method to block and check on job until complete."""
+
+        # Used these numbers so failures surface fast
+        wait = 5  # start at five seconds
+        max_wait = 60 * 5  # 5 minute wait
+        multiplier = 2  # scale wait by 2 every iteration
+
+        while self.state not in _PIPELINE_COMPLETE_STATES:
+            self._sync_gca_resource()
+            time.sleep(wait)
+            _LOGGER.info(
+                "Training %s current state:\n%s"
+                % (self._gca_resource.name, self._gca_resource.state)
+            )
+            wait = min(wait * multiplier, max_wait)
+
+        self._raise_failure()
+
+        if self._gca_resource.model_to_upload and not self.has_failed:
+            _LOGGER.info(
+                "Model available at %s" % self._gca_resource.model_to_upload.name
+            )
+
+    def _raise_failure(self):
+        """Helper method to raise failure if TrainingPipeline fails.
+
+        Raises:
+            RuntimeError: If training failed."""
+        if self._gca_resource.error.code != code_pb2.OK:
+            raise RuntimeError("Training failed with:\n%s" % self._gca_resource.error)
+
+    @property
+    def has_failed(self) -> bool:
+        """Returns True if training has failed. False otherwise."""
+        self._assert_has_run()
+        return self.state == gca_pipeline_state.PipelineState.PIPELINE_STATE_FAILED
+
+    def _dashboard_uri(self) -> str:
+        """Helper method to compose the dashboard uri where training can be viewed."""
+        fields = utils.extract_fields_from_resource_name(self.resource_name)
+        url = f"https://console.cloud.google.com/ai/platform/locations/{fields.location}/training/{fields.id}?project={fields.project}"
+        return url
+
+    def _sync_gca_resource(self):
+        """Helper method to sync the local gca_source against the service."""
+        self._gca_resource = self.api_client.get_training_pipeline(
+            name=self.resource_name
+        )
+
+    @property
+    def _has_run(self) -> bool:
+        """Helper property to check if this training job has been run."""
+        return self._gca_resource is not None
+
+    def _assert_has_run(self):
+        """Helper method to assert that this training has run."""
+        if not self._has_run:
+            raise RuntimeError(
+                "TrainingPipeline has not been launched. You must run this"
+                " TrainingPipeline using TrainingPipeline.run. "
+            )
 
 
 def _timestamped_gcs_dir(root_gcs_path: str, dir_name_prefix: str) -> str:
@@ -524,15 +831,12 @@ class _DistributedTrainingSpec(NamedTuple):
 
 
 # TODO(b/172368325) add scheduling, custom_job.Scheduling
-class CustomTrainingJob(base.AiPlatformResourceNoun):
+class CustomTrainingJob(_TrainingJob):
     """Class to launch a Custom Training Job in AI Platform using a script.
 
     Takes a training implementation as a python script and executes that script
     in Cloud AI Platform Training.
     """
-
-    client_class = pipeline_service_client.PipelineServiceClient
-    _is_client_prediction_client = False
 
     # TODO(b/172365796) add remainder of model optional arguments
     def __init__(
@@ -617,15 +921,15 @@ class CustomTrainingJob(base.AiPlatformResourceNoun):
                 Bucket used to stage source and training artifacts. Overrides
                 staging_bucket set in aiplatform.init.
         """
-        utils.validate_display_name(display_name)
-        super().__init__(project=project, location=location, credentials=credentials)
-        self._display_name = display_name
-        self._script_path = script_path
+        super().__init__(
+            display_name=display_name,
+            project=project,
+            location=location,
+            credentials=credentials,
+        )
+
         self._container_uri = container_uri
         self._requirements = requirements
-        self._staging_bucket = staging_bucket
-        self._project = project
-        self._credentials = credentials
         self._model_serving_container_image_uri = model_serving_container_image_uri
         self._model_serving_container_predict_route = (
             model_serving_container_predict_route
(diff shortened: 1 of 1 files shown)


google/cloud/aiplatform/training_job_base.py
```diff
@@ -0,0 +1,298 @@
...
+        self._model_serving_container_image_uri = model_serving_container_image_uri
+        self._model_serving_container_predict_route = (
+            model_serving_container_predict_route
+        )
+        self._model_serving_container_health_route = (
+            model_serving_container_health_route
+        )
+        self._gca_resource = None
+
+    def _create_managed_model(self, model_display_name: str) -> gca_model.Model:
```
ivanmkc, line 84: Encapsulated this logic into a function to emphasize the parameter dependencies to generate a managed model.

google/cloud/aiplatform/training_job_base.py
```diff
@@ -0,0 +1,298 @@
...
+                health_route=self._model_serving_container_health_route,
+            )
+
+            managed_model = gca_model.Model(
+                display_name=model_display_name, container_spec=container_spec
+            )
+
+        return managed_model
+
+    def _create_input_data_config(self,
```
ivanmkc, line 100: Encapsulated this logic into a function to emphasize the parameter dependencies to generate a input data config.

google/cloud/aiplatform/training_job_base.py
```diff
@@ -0,0 +1,298 @@
...
+        
+        return input_data_config
+
+    def run_job(self,            
+            training_task_definition: str,
+            training_task_inputs: dict,
+            datasetWithSplits: Optional[DatasetWithSplits] = None,
+            model_display_name: Optional[str] = None,
+            base_output_dir: Optional[str] = None):
+        """Runs the training job.
```
ivanmkc, line 130: Does this need docstrings if it's not supposed to be used by users since they be using the specific TrainingJob subclasses?

google/cloud/aiplatform/training_job_base.py
```diff
@@ -0,0 +1,298 @@
...
+        """
+        if self._has_run:
```
ivanmkc, line 132: This is the only functionality that differs from the existing `CustomTrainingJob` implementation. 

This validation was previously performed earlier but I've moved it into the base class (i.e. it occurs later than before) to hide it from the subclasses implementation details.

google/cloud/aiplatform/training_job_base.py
```diff
@@ -0,0 +1,298 @@
...
+
+        if model is None:
+            _LOGGER.warning(
+                "Training did not produce a Managed Model returning None. "
+                + self._model_upload_fail_string
+            )
+        return model
+
+    @property
+    def state(self) -> gca_pipeline_state.PipelineState:
```
ivanmkc, line 180: Copied verbatim from `CustomModelTrainingJob`

google/cloud/aiplatform/training_job_base.py
```diff
@@ -0,0 +1,298 @@
...
+        """Current training state."""
+        self._assert_has_run()
+        return self._gca_resource.state
+
+    def get_model(self) -> Optional[models.Model]:
```
ivanmkc, line 185: Copied verbatim from `CustomModelTrainingJob`

google/cloud/aiplatform/training_job_base.py
```diff
@@ -0,0 +1,298 @@
...
+            model: AI Platform Model produced by this training or None if a model was
+                not produced by this training.
+        """
+        self._assert_has_run()
+        if not self._gca_resource.model_to_upload:
+            raise RuntimeError(self._model_upload_fail_string)
+
+        return self._get_model()
+
+    def _get_model(self) -> Optional[models.Model]:
```
ivanmkc, line 198: Copied verbatim from `CustomModelTrainingJob`

google/cloud/aiplatform/training_job_base.py
```diff
@@ -0,0 +1,298 @@
...
+
+        if self._gca_resource.model_to_upload.name:
+            fields = utils.extract_fields_from_resource_name(
+                self._gca_resource.model_to_upload.name
+            )
+            return models.Model(
+                fields.id, project=fields.project, location=fields.location
+            )
+
+    def _block_until_complete(self):
```
ivanmkc, line 226: Copied verbatim from `CustomModelTrainingJob`

google/cloud/aiplatform/training_job_base.py
```diff
@@ -0,0 +1,298 @@
...
+            wait = min(wait * multiplier, max_wait)
+
+        self._raise_failure()
+
+        if self._gca_resource.model_to_upload and not self.has_failed:
+            _LOGGER.info(
+                "Model available at %s" % self._gca_resource.model_to_upload.name
+            )
+
+    def _raise_failure(self):
```
ivanmkc, line 250: Copied verbatim from `CustomModelTrainingJob`

google/cloud/aiplatform/training_job_base.py
```diff
@@ -0,0 +1,298 @@
...
+        """Helper method to raise failure if TrainingPipeline fails.
+
+        Raises:
+            RuntimeError: If training failed."""
+        if self._gca_resource.error.code != code_pb2.OK:
+            raise RuntimeError("Training failed with:\n%s" % self._gca_resource.error)
+
+    @property
+    def has_failed(self) -> bool:
```
ivanmkc, line 259: Copied verbatim from `CustomModelTrainingJob`

google/cloud/aiplatform/training_job_base.py
```diff
@@ -0,0 +1,298 @@
...
+        """Returns True if training has failed. False otherwise."""
+        self._assert_has_run()
+        return self.state == gca_pipeline_state.PipelineState.PIPELINE_STATE_FAILED
+
+    def _dashboard_uri(self) -> str:
```
ivanmkc, line 264: Copied verbatim from `CustomModelTrainingJob`

google/cloud/aiplatform/training_job_base.py
```diff
@@ -0,0 +1,298 @@
...
+        """Helper method to compose the dashboard uri where training can be viewed."""
+        fields = utils.extract_fields_from_resource_name(self.resource_name)
+        url = f"https://console.cloud.google.com/ai/platform/locations/{fields.location}/training/{fields.id}?project={fields.project}"
+        return url
+
+    def _sync_gca_resource(self):
```
ivanmkc, line 270: Copied verbatim from `CustomModelTrainingJob`

google/cloud/aiplatform/training_job_base.py
```diff
@@ -0,0 +1,298 @@
...
+        """Helper method to sync the local gca_source against the service."""
+        self._gca_resource = self.api_client.get_training_pipeline(
+            name=self.resource_name
+        )
+
+    @property
+    def _model_upload_fail_string(self) -> str:
```
ivanmkc, line 277: Copied verbatim from `CustomModelTrainingJob`

google/cloud/aiplatform/training_job_base.py
```diff
@@ -0,0 +1,298 @@
...
+        return (
+            f"Training Pipeline {self.resource_name} is not configured to upload a "
+            "Model. Create the Training Pipeline with "
+            "model_serving_container_image_uri and model_display_name passed in. "
+            "Ensure that your training script saves to model to "
+            "os.environ['AIP_MODEL_DIR']."
+        )
+
+    @property
+    def _has_run(self) -> bool:
```
ivanmkc, line 288: Copied verbatim from `CustomModelTrainingJob`

google/cloud/aiplatform/training_job_base.py
```diff
@@ -0,0 +1,298 @@
...
+        """Helper property to check if this training job has been run."""
+        return self._gca_resource is not None
+
+    def _assert_has_run(self):
```
ivanmkc, line 292: Copied verbatim from `CustomModelTrainingJob`

google/cloud/aiplatform/training_job_base.py
```diff
@@ -0,0 +1,298 @@
...
+_PIPELINE_COMPLETE_STATES = set(
+    [
+        gca_pipeline_state.PipelineState.PIPELINE_STATE_SUCCEEDED,
+        gca_pipeline_state.PipelineState.PIPELINE_STATE_FAILED,
+        gca_pipeline_state.PipelineState.PIPELINE_STATE_CANCELLED,
+        gca_pipeline_state.PipelineState.PIPELINE_STATE_PAUSED,
+    ]
+)
+
+class DatasetWithSplits():
```
ivanmkc, line 39: New class to encapsulate all the parameters that need to be passed together.

google/cloud/aiplatform/training_job_base.py
```diff
@@ -0,0 +1,298 @@
...
+        validation_fraction_split: float,
+        test_fraction_split: float
+    ):
+        self.dataset=dataset
+        self.training_fraction_split=training_fraction_split
+        self.validation_fraction_split=validation_fraction_split
+        self.test_fraction_split=test_fraction_split
+
+
+class TrainingJob(base.AiPlatformResourceNoun):
```
ivanmkc, line 53: Can potentially make an abc to prevent instantiation. Thoughts?

google/cloud/aiplatform/training_job_base.py
ivanmkc: Hm, seems like this isn't used by AutoML. I guess it's specific to Custom Training then? Will move there then

sasha-gitg:
> ## Remaining questions
> 1. In the `TrainingJob.run` method, `training_fraction_split`, `validation_fraction_split`, `test_fraction_split: float` are all only needed when the `dataset` parameter is also provided. I've grouped this together into `DatasetWithSplits`. I'd want to replace the `CustomTrainingJob.run` parameters with this as well. What do you all think?

The preference from the design is to flatten coupled arguments as long the objects they represent are not deeply nested. There is precedence in ML libraries for this pattern:

[keras.Model.fit](https://www.tensorflow.org/api_docs/python/tf/keras/Model#fit) with validation_split, similar the API surface discussed here.
[sklearn.linear_model.LogisticRegression](https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.LogisticRegression.html) Particular parameter value combinations are invalid.

The aim for this SDK is to feel familiar to practitioners that use these libraries. We'll gate on feedback on whether we should change this.

> 2. Should we put `TrainingJob` into its own file and separate out `CustomTrainingJob` and `AutoMLTablesTrainingJob` into their own? If not, how should we organize things as to not have massive files?

We should gate this on whether the organization makes sense to group these classes together in the same module not necessarily that there is some threshold of number of lines we want to stay under. Since these all relate to training using the PipelineService I think we're safe to keep them in a single module.   
 
> 3. Should I make `TrainingJob` an abstract base class to prevent initialization? I can also do this by overriding `new`:

Looking at the code, I'm not sure it would prevent initialization because it looks like the interface is completely defined.

We should use an ABC to define an interface that concrete classes should implement. If our parent class already fully implements that interface it should be usable. To me, based on the code, it looks like you may want to separate out some of the shared logic as private methods in the ABC and require the concrete classes implement some of the public interface. Either that or derive AutoMLTablesTrainingJob from CustomTrainingJob.

google/cloud/aiplatform/training_job_base.py
ivanmkc: I ended up having each subclass create their own `Model` and pass it in to the base class. This is due to the fact that each training job type has different requirements as defined in the yaml files.

ivanmkc:
Thanks for the clarifications @sasha-gitg, they all make sense to me. I will make the changes.

ivanmkc:
> We should gate this on whether the organization makes sense to group these classes together in the same module not necessarily that there is some threshold of number of lines we want to stay under. Since these all relate to training using the PipelineService I think we're safe to keep them in a single module.

Sure thing. One consideration for splitting off the base class into its own module is whether other teams will be building off the base class or will it just be us.  Since it looks like its just us, a relatively small group of developers, keeping it in the same module sgtm.

google/cloud/aiplatform/training_job_base.py
```diff
@@ -0,0 +1,262 @@
...
+            fraction_split = gca_training_pipeline.FractionSplit(
+                training_fraction=training_fraction_split,
+                validation_fraction=validation_fraction_split,
+                test_fraction=test_fraction_split,
+            )
+
+            # create input data config
+            input_data_config = gca_training_pipeline.InputDataConfig(
+                fraction_split=fraction_split,
+                dataset_id=dataset.name,
```
ivanmkc, line 81: @sasha-gitg The only thing `dataset` is used for is to get the name, should the name just be passed in directly? I'll leave it as a dataset object for now.

ivanmkc:
@sasha-gitg What do you want me to do with the tests? Your CustomTraining tests already ensure coverage of the super class. Do you want me to refactor the tests in some way as well?

google/cloud/aiplatform/training_job_base.py
sasha-gitg: Eventually we should accept either a Dataset object, Dataset resource name, or a Dataset resource id though that doesn't have to be added in this PR.

sasha-gitg:
> @sasha-gitg What do you want me to do with the tests? Your CustomTraining tests already ensure coverage of the super class. Do you want me to refactor the tests in some way as well?

If test are passing with full coverage after refactoring then we can move forward with the tests we have.

sasha-gitg:
> However, putting everything in one module has the issue that we're combining all the import statements.
> 
> Examining the import dependencies typically are a great way to check if the coupling of your class. For example, if `AutoMLTablesTrainingJob` was in its own module, then I can quickly see that it only depends on Tables related classes and not CustomTraining ones. Putting everything into one module makes that hard to see.

This is a more compelling argument to split the files but I would move this decision to the PR that implements `AutoMLTablesTrainingJob` so we can concretely see the coupling we are trying to avoid. 

> However, we also want to be able import training jobs in one line:
> 
> `from training_jobs import CustomTrainingJob, AutoMLTablesTrainingJob`
> 

I don't see value in this considering we expose all classes on the SDKs surface and the expectation is that our internal code imports implementations directly. Our style guide also enforces module level imports. 

> It seems like we can keep everything in one TrainingJob module but still have multiple files:
> 
> https://stackoverflow.com/questions/24100558/how-can-i-split-a-module-into-multiple-files-without-breaking-a-backwards-compa/24100645
> 
> Would you be open to this approach or do you still prefer all training_job's in one file?

Yes, this is the approach we would take to namespace all these classes to the same module when splitting them into different files. But if we're going to add a level of indirection we should justify that tradeoff with concrete benefits.

As the SDK grows these benefits we'll become more apparent but I would prefer we avoid adding premature complexity. I do think that implementing AutoMLTablesTrainingJob will surface those concrete benefits.

ivanmkc:
Sounds good, thanks for the review! I'll move everything into the `training_jobs.py` file then.

ivanmkc:
> > However, we also want to be able import training jobs in one line:
> > `from training_jobs import CustomTrainingJob, AutoMLTablesTrainingJob`
> 
> I don't see value in this considering we expose all classes on the SDKs surface and the expectation is that our internal code imports implementations directly. Our style guide also enforces module level imports.

Sorry, I meant, we want to import CustomTrainingJob and AutoMLTablesTrainingJob as part of the same module.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -66,6 +66,290 @@
...
+            dataset (aiplatform.Dataset):
+                Optional. The dataset within the same Project from which data will be used to train the Model. The
+                Dataset must use schema compatible with Model being trained,
+                and what is compatible should be described in the used
+                TrainingPipeline's [training_task_definition]
+                [google.cloud.aiplatform.v1beta1.TrainingPipeline.training_task_definition].
+                For tabular Datasets, all their data is exported to
+                training, to pick and choose from.
+            training_fraction_split (float):
+                The fraction of the input data that is to be
```
ivanmkc, line 146: @sasha-gitg sometimes I see `Required` and sometimes `Optional` and sometimes nothing is written here. Any guidance on why nothing is written for this parameter (I copied it from CustomTrainingJobClass).

google/cloud/aiplatform/training_job_base.py
ivanmkc: Got it.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -66,6 +66,290 @@
...
+        project: Optional[str] = None,
+        location: Optional[str] = None,
+        credentials: Optional[auth_credentials.Credentials] = None,
+    ):
+        """Constructs a Training Job.
+
+        Args:
+            display_name (str):
+                Required. The user-defined name of this TrainingPipeline.
+            container_uri (str):
```
ivanmkc, line 86: Is this only relevant to Custom Training?

@sasha-gitg

sasha-gitg: Yes.

google/cloud/aiplatform/training_jobs.py
sasha-gitg: These are generally copied from the protos if they represent the same field: https://github.com/googleapis/python-aiplatform/blob/dev/google/cloud/aiplatform_v1beta1/types/training_pipeline.py#L260

So we inherit the arg commenting from the service.

For comments we add, it's only necessary we mark them as Required if they are indeed so.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -602,195 +871,51 @@ def run(
...
-            training_task_inputs=json_format.ParseDict(
-                {
-                    "workerPoolSpecs": [worker_pool_spec],
-                    "baseOutputDirectory": {"output_uri_prefix": base_output_dir},
-                },
-                struct_pb2.Value(),
-            ),
-            model_to_upload=managed_model,
-            input_data_config=input_data_config,
+        managed_model = gca_model.Model(
```
sasha-gitg, line 904: Setting this value will throw an exception if the Training script does not produce model. I think we want to previous logic where it's possible managed_model is None.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -66,6 +66,288 @@
...
+                uploaded into AI Platform, then the model_to_upload's
+                resource ``name``
+                is populated. The Model is always uploaded into the Project
+                and Location in which this pipeline is.
+            base_output_dir (str):
+                Optional. GCS output directory of job. If not provided a
+                timestamped directory in the staging directory will be used.
+    """
+
+        if self._has_run:
```
sasha-gitg, line 177: The issue with calling this here is that the child class, Custom Training, would have already executed logic with side effects that should have been avoided. In this particular instance, the python package would have been created and uploaded to GCS and then we would raise here when we should have raised earlier. This should be moved to the top of the child classes run call.

sasha-gitg (changes requested):
The tests are broken in the build. Those should pass to ensure this refactor did not break the current class.

Nit: The PR title should technically be called "refactor:..." as we shouldn't be altering behavior with this change.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -66,6 +66,288 @@
 )
 
 
+class TrainingJob(base.AiPlatformResourceNoun):
```
sasha-gitg, line 69: This class seems to implement the entire interface so it's a concrete class which gives the impression that it can and should be used. That is not our intention. I think the fastest way to resolve this is:

1. `run_job` should be private `_run_job`
2. `run` should be [abstractmethod](https://docs.python.org/3/library/abc.html#abc.abstractmethod) so all child classes are required to implement it. The expectation is that `run` calls `_run_job`. That will make this class an ABC.
3. Exception strings specific to the child class should be abstract properties which child classes implement. ie:
```python
@property
@abc.abstractmethod
def _model_upload_fail_string(self) -> str:
    pass
```
4. This class should be a private class since it's only consumed in this module.

sasha-gitg (changes requested):

ivanmkc:
sgtm @sasha-gitg! will make the changes today

google/cloud/aiplatform/training_jobs.py
```diff
@@ -66,6 +66,288 @@
...
+                Optional credentials to use to retrieve the model.
+        """
+
+        super().__init__(project=project, location=location, credentials=credentials)
+        self._display_name = display_name
+        self._project = project
+        self._credentials = credentials
+        self._gca_resource = None
+
+    def run_job(
```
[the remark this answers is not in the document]
ivanmkc: done

google/cloud/aiplatform/training_jobs.py
ivanmkc: Ok removed from base class

google/cloud/aiplatform/training_jobs.py
```diff
@@ -602,195 +871,51 @@ def run(
...
-    @property
-    def _model_upload_fail_string(self) -> str:
-        """Helper property for model upload failure."""
-        return (
-            f"Training Pipeline {self.resource_name} is not configured to upload a "
-            "Model. Create the Training Pipeline with "
-            "model_serving_container_image_uri and model_display_name passed in. "
-            "Ensure that your training script saves to model to "
-            "os.environ['AIP_MODEL_DIR']."
+        return super().run_job(
```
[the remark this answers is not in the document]
ivanmkc: Fixed

google/cloud/aiplatform/training_jobs.py
```diff
@@ -66,6 +66,288 @@
...
+                resource ``name``
+                is populated. The Model is always uploaded into the Project
+                and Location in which this pipeline is.
+            base_output_dir (str):
+                Optional. GCS output directory of job. If not provided a
+                timestamped directory in the staging directory will be used.
+    """
+
+        if self._has_run:
+            raise RuntimeError("Custom Training has already run.")
```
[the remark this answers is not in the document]
ivanmkc: Fixed

google/cloud/aiplatform/training_jobs.py
ivanmkc: Ok, got it

google/cloud/aiplatform/training_jobs.py
ivanmkc: all done

google/cloud/aiplatform/training_jobs.py
ivanmkc: Fixed

google/cloud/aiplatform/training_jobs.py
ivanmkc: Good catch, this was unintended.

ivanmkc:
@sasha-gitg put in the fixes!

google/cloud/aiplatform/training_jobs.py
```diff
@@ -66,6 +66,288 @@
...
+        """Helper method to sync the local gca_source against the service."""
+        self._gca_resource = self.api_client.get_training_pipeline(
+            name=self.resource_name
+        )
+
+    @property
+    def _model_upload_fail_string(self) -> str:
+        """Helper property for model upload failure."""
+        return (
+            f"Training Pipeline {self.resource_name} is not configured to upload a "
```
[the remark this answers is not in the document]

google/cloud/aiplatform/training_jobs.py
```diff
@@ -26,6 +26,7 @@
 import time
 from typing import Callable, List, Optional, Sequence, Union
 
+from abc import ABC, abstractmethod
```
sasha-gitg, line 29: Module level import here.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -66,6 +67,289 @@
 )
 
 
+class _TrainingJob(ABC, base.AiPlatformResourceNoun):
```
sasha-gitg, line 70: You shouldn't need to inherit `ABC` here because there are unimplemented `abstractmethods` in this class.

google/cloud/aiplatform/training_jobs.py
sasha-gitg: Please remove the docstring.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -616,126 +910,17 @@ def run(
...
-            training_task_inputs=json_format.ParseDict(
-                {
-                    "workerPoolSpecs": [worker_pool_spec],
-                    "baseOutputDirectory": {"output_uri_prefix": base_output_dir},
-                },
-                struct_pb2.Value(),
-            ),
-            model_to_upload=managed_model,
-            input_data_config=input_data_config,
+        return super()._run_job(
```
sasha-gitg, line 913: This should be `self._run_job` since we didn't override `_run_job`.

sasha-gitg (approved):
Looks good! Requested a few minor changes. Also need to resolve merge conflicts with distributed training PR (those changes should only affect the `CustomTrainingJob` class).

google/cloud/aiplatform/training_jobs.py
```diff
@@ -66,6 +67,289 @@
...
+                training_fraction=training_fraction_split,
+                validation_fraction=validation_fraction_split,
+                test_fraction=test_fraction_split,
+            )
+
+            # create input data config
+            input_data_config = gca_training_pipeline.InputDataConfig(
+                fraction_split=fraction_split,
+                dataset_id=dataset.name,
+                gcs_destination=gca_io.GcsDestination(
```
ivanmkc, line 207: Is this only used for Custom Training? The site https://cloud.google.com/ai-platform-unified/docs/reference/rest/v1beta1/projects.locations.trainingPipelines#TrainingPipeline.FIELDS.training_task_definition only says:

```
object (GcsDestination)

The Cloud Storage location where the training data is to be written to. In the given directory a new directory is created with name: dataset-<dataset-id>-<annotation-type>-<timestamp-of-training-call> where timestamp is in YYYY-MM-DDThh:mm:ss.sssZ ISO-8601 format. All training input data is written into that directory.

The AI Platform environment variables representing Cloud Storage data URIs are represented in the Cloud Storage wildcard format to support sharded data. e.g.: "gs://.../training-*.jsonl"

AIP_DATA_FORMAT = "jsonl" for non-tabular data, "csv" for tabular data
AIP_TRAINING_DATA_URI =
"gcsDestination/dataset---/training-*.${AIP_DATA_FORMAT}"

AIP_VALIDATION_DATA_URI =
"gcsDestination/dataset---/validation-*.${AIP_DATA_FORMAT}"

AIP_TEST_DATA_URI =
"gcsDestination/dataset---/test-*.${AIP_DATA_FORMAT}"
```

What training data being written?

ivanmkc: @sasha-gitg

google/cloud/aiplatform/training_jobs.py
ivanmkc: will fix

google/cloud/aiplatform/training_jobs.py
ivanmkc: I was wondering about this. So according to the style guide I should be import abc, and then refer use `abc.abstractmethod` every time in the code?

What about the line above it? 
`from typing import Callable, List, Optional, Sequence, Union`

google/cloud/aiplatform/training_jobs.py
sasha-gitg: Good point. Yes, I think this is custom training specific. Though, it would be worth confirming but using the API without passing the field.

google/cloud/aiplatform/training_jobs.py
sasha-gitg: Yes on `abc.abstractmethod`.

`typing` is an exception to the rule [typing-imports](https://engdoc.corp.google.com/eng/doc/devguide/py/style/index.md?cl=head#typing-imports).

PR #62, opened 2020-11-13
feat: Added AutoMLTablesTrainingJob and tests

ivanmkc:
Added support for AutoMLTablesTraining.

Fixes https://b.corp.google.com/issues/172282518
Depends on https://github.com/googleapis/python-aiplatform/pull/49

Could possibly add more client-side validation, but currently deferring validation to the backend services pending more discussion.

### DIFF
--- a/google/cloud/aiplatform/__init__.py
+++ b/google/cloud/aiplatform/__init__.py
@@ -21,8 +21,11 @@ from google.cloud.aiplatform import initializer
 from google.cloud.aiplatform.datasets import Dataset
 from google.cloud.aiplatform.models import Endpoint
 from google.cloud.aiplatform.models import Model
-from google.cloud.aiplatform.training_jobs import CustomTrainingJob
 from google.cloud.aiplatform.jobs import BatchPredictionJob
+from google.cloud.aiplatform.training_jobs import (
+    CustomTrainingJob,
+    AutoMLTablesTrainingJob,
+)
 
 """
 Usage:
@@ -36,6 +39,7 @@ __all__ = (
     "gapic",
     "BatchPredictionJob",
     "CustomTrainingJob",
+    "AutoMLTablesTrainingJob",
     "Model",
     "Dataset",
     "Endpoint",
--- a/google/cloud/aiplatform/training_jobs.py
+++ b/google/cloud/aiplatform/training_jobs.py
@@ -48,7 +48,6 @@ from google.cloud.aiplatform_v1beta1.types import (
     training_pipeline as gca_training_pipeline,
 )
 
-
 from google.cloud import storage
 from google.protobuf import json_format
 from google.protobuf import struct_pb2
@@ -180,7 +179,7 @@ class _TrainingJob(base.AiPlatformResourceNoun):
                 than the one given on input. The output URI will
                 point to a location where the user only has a
                 read access.
-            training_task_inputs (~.struct.Value):
+            training_task_inputs (dict):
                 Required. The training task's parameter(s), as specified in
                 the
                 ``training_task_definition``'s
@@ -238,7 +237,9 @@ class _TrainingJob(base.AiPlatformResourceNoun):
         training_pipeline = gca_training_pipeline.TrainingPipeline(
             display_name=self._display_name,
             training_task_definition=training_task_definition,
-            training_task_inputs=training_task_inputs,
+            training_task_inputs=json_format.ParseDict(
+                training_task_inputs, struct_pb2.Value()
+            ),
             model_to_upload=model,
             input_data_config=input_data_config,
         )
@@ -890,14 +891,6 @@ class CustomTrainingJob(_TrainingJob):
                 Required: Uri of the training container image in the GCR.
             requirements (Sequence[str]):
                 List of python packages dependencies of script.
-            project (str):
-                Optional project to retrieve model from. If not set, project set in
-                aiplatform.init will be used.
-            location (str):
-                Optional location to retrieve model from. If not set, location set in
-                aiplatform.init will be used.
-            credentials (auth_credentials.Credentials):
-                Optional credentials to use to retrieve the model.
             model_serving_container_image_uri (str):
                 If the training produces a managed AI Platform Model, the URI of the
                 Model serving container suitable for serving the model produced by the
@@ -1055,13 +1048,13 @@ class CustomTrainingJob(_TrainingJob):
                 The number of accelerators to attach to a worker replica.
             training_fraction_split (float):
                 The fraction of the input data that is to be
-                used to train the Model.
+                used to train the Model. This is ignored if Dataset is not provided.
             validation_fraction_split (float):
                 The fraction of the input data that is to be
-                used to validate the Model.
+                used to validate the Model. This is ignored if Dataset is not provided.
             test_fraction_split (float):
                 The fraction of the input data that is to be
-                used to evaluate the Model.
+                used to evaluate the Model. This is ignored if Dataset is not provided.
 
         Returns:
             model: The trained AI Platform Model resource or None if training did not
@@ -1129,13 +1122,10 @@ class CustomTrainingJob(_TrainingJob):
             if args:
                 spec["pythonPackageSpec"]["args"] = args
 
-        training_task_inputs = json_format.ParseDict(
-            {
-                "workerPoolSpecs": worker_pool_specs,
-                "baseOutputDirectory": {"output_uri_prefix": base_output_dir},
-            },
-            struct_pb2.Value(),
-        )
+        training_task_inputs = {
+            "workerPoolSpecs": worker_pool_specs,
+            "baseOutputDirectory": {"output_uri_prefix": base_output_dir},
+        }
 
         training_task_definition = schema.training_job.definition.custom_task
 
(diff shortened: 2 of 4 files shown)


google/cloud/aiplatform/training_jobs.py
```diff
@@ -616,181 +908,233 @@ def run(
...
-            )
+        training_task_definition = schema.training_job.definition.tabular_task
+
+        training_task_inputs_dict = {
+            # required inputs
+            "targetColumn": target_column,
+            "transformations": self._column_transformations,
+            "trainBudgetMilliNodeHours": budget_milli_node_hours,
+            # optional inputs
+            "weightColumnName": weight_column,
```
ivanmkc, line 1111: In [google3/google/cloud/aiplatform/publicfiles/trainingjob/definition/automl_tables.proto](https://source.corp.google.com/piper///depot/google3/google/cloud/aiplatform/publicfiles/trainingjob/definition/automl_tables.proto;rcl=342552818;l=21), this column is referred to as "weight_column_name".

However, in `gs://google-cloud-aiplatform/schema/trainingjob/definition/automl_tabular_1.0.0.yaml`, it gives the name as "weightColumn". This is incorrect and results in an error.

@sasha-gitg why is there a discrepancy here?

google/cloud/aiplatform/training_jobs.py
```diff
@@ -616,181 +908,233 @@ def run(
...
-            _LOGGER.warn(
-                "Training did not produce a Managed Model returning None. "
-                + self._model_upload_fail_string
-            )
-        return model
+class AutoMLTablesTrainingJob(_TrainingJob):
+    def __init__(
+        self,
+        display_name: str,
+        optimization_objective: str,
```
ivanmkc, line 938: According to the protos and yaml, the `optimization_objective` is optional since there is a default if it's not supplied.

Also, the design doc has this signature, which doesn't have the `optimization_prediction_type` parameter either. We can support this by inferring that info based on the `optimization_objective`, but that can be done a multitude of ways. 

My personal preferred way is to create an OptimizationObjective abstract class and create subclasses of each for each `optimization_objective` type. That would encapsulate the `optimization_objective`, `optimization_prediction_type `, `optimization_objective_recall_value` and `optimization_objective_precision_value` parameters together into one. 

However, from our other convos it seems like you might prefer just passing in a string for `optimization_objective` and having the other parameters be optional.

Let me know what you prefer.

ivanmkc: @sasha-gitg

google/cloud/aiplatform/__init__.py
```diff
@@ -31,4 +34,11 @@
 """
 init = initializer.global_config.init
 
-__all__ = ("gapic", "CustomTrainingJob", "Model", "Dataset", "Endpoint")
+__all__ = (
```
ivanmkc, line 37: The linter did this

google/cloud/aiplatform/training_jobs.py
sasha-gitg: With respect to `optimization_prediction_type` we received feedback to include that as the top level argument because it requires less knowledge overhead than `optimization_objective`. Additionally, we should include all the values flattened in the API surface and validate or ignore arguments based on valid combinations. There's precedence for this pattern:

https://github.com/scikit-learn/scikit-learn/blob/0fb307bf3/sklearn/linear_model/_logistic.py#L1011

So the current state of the input arguments LGTM just elevate `optimization_prediction_type` over `optimization_objective` and make `optimization_objective` optional but add comments to explain the defaults when selecting `optimization_prediction_type`.

google/cloud/aiplatform/training_jobs.py
sasha-gitg: Following up on this.

sasha-gitg: Synced with the team. The yaml is incorrect and will be updated. The field should be ```weightColumnName```. If users are using our Model Builder SDK then hopefully we have caught these issues early during our development. This should become less of an issue for the service as we move forward because it will support all previously versioned yaml schemas.

google/cloud/aiplatform/training_jobs.py
ivanmkc: Thanks for the context, sgtm!

google/cloud/aiplatform/__init__.py
sirtorry: looks good.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -1185,4 +1177,207 @@ def _model_upload_fail_string(self) -> str:
...
+        optimization_prediction_type: str,
+        optimization_objective: Optional[str],
+        column_transformations: Optional[Union[Dict, List[Dict]]],
+        optimization_objective_recall_value: Optional[float],
+        optimization_objective_precision_value: Optional[float],
+        project: Optional[str] = None,
+        location: Optional[str] = None,
+        credentials: Optional[auth_credentials.Credentials] = None,
+    ):
+        # TODO: Fix doc string
```
sasha-gitg, line 1192: Remove TODO

google/cloud/aiplatform/training_jobs.py
```diff
@@ -1185,4 +1177,207 @@ def _model_upload_fail_string(self) -> str:
...
+        """Constructs a AutoML Tables Training Job.
+
+        Args:
+            display_name (str):
+                Required. The user-defined name of this TrainingPipeline.
+            optimization_prediction_type (str):
```
sasha-gitg, line 1198: Include allowed values.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -1185,4 +1177,207 @@ def _model_upload_fail_string(self) -> str:
...
+        self._optimization_objective = optimization_objective
+        self._optimization_prediction_type = optimization_prediction_type
+        self._optimization_objective_recall_value = optimization_objective_recall_value
+        self._optimization_objective_precision_value = (
+            optimization_objective_precision_value
+        )
+
+    def run(
+        self,
+        dataset: Optional[datasets.Dataset],
```
sasha-gitg, line 1272: Shouldn't this be required?

google/cloud/aiplatform/training_jobs.py
```diff
@@ -1185,4 +1177,207 @@ def _model_upload_fail_string(self) -> str:
...
+        target_column: str,
+        training_fraction_split: float,
```
sasha-gitg, line 1274: Add defaults to 0.8, 0.1, 0.1 for the splits.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -1185,4 +1177,207 @@ def _model_upload_fail_string(self) -> str:
 
 
 class AutoMLTablesTrainingJob(_TrainingJob):
-    pass
+    def __init__(
+        self,
+        display_name: str,
+        optimization_prediction_type: str,
+        optimization_objective: Optional[str],
```
sasha-gitg, line 1184: All of these optional arguments should default to None.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -1185,4 +1177,207 @@ def _model_upload_fail_string(self) -> str:
...
+            # optional inputs
+            "weightColumnName": weight_column,
+            "disableEarlyStopping": disable_early_stopping,
+            "optimizationObjective": self._optimization_objective,
+            "predictionType": self._optimization_prediction_type,
+            "optimizationObjectiveRecallValue": self._optimization_objective_recall_value,
+            "optimizationObjectivePrecisionValue": self._optimization_objective_precision_value,
+        }
+
+        model = gca_model.Model(display_name=model_display_name,)
```
sasha-gitg, line 1363: If the display name is None does the service generate a name for the Model?

google/cloud/aiplatform/training_jobs.py
```diff
@@ -1185,4 +1177,207 @@ def _model_upload_fail_string(self) -> str:
...
+
+        return self._run_job(
+            training_task_definition=training_task_definition,
+            training_task_inputs=json_format.ParseDict(
```
sasha-gitg, line 1367: Both `AutoMLTrainingJob` and `CustomTrainingJob` call json_format.ParseDict on their training_task_inputs before passing to `_run_job`. I wonder if it makes sense to move that logic into `_run_job` to dedup the code.

sasha-gitg (approved):
LGTM! Minor comments.

google/cloud/aiplatform/training_jobs.py
ivanmkc: Done

google/cloud/aiplatform/training_jobs.py
ivanmkc: Done

google/cloud/aiplatform/training_jobs.py
ivanmkc: Ah, I see that the behaviour is different for every datatype. Tables requires a model to be provided according to the yaml and is not actually optional as the protos suggest.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -1185,4 +1177,207 @@ def _model_upload_fail_string(self) -> str:
...
+
+    def run(
+        self,
+        dataset: Optional[datasets.Dataset],
+        target_column: str,
+        training_fraction_split: float,
+        validation_fraction_split: float,
+        test_fraction_split: float,
+        weight_column: Optional[str] = None,
+        budget_milli_node_hours: int = 1000,
```
ivanmkc, line 1278: I'm guessing setting this to less than 1000 hours will cause the backend to use 1000 hours instead. Probably same with the maximum value.

I have no idea if this is true though and this seems opaque to users.

Ideally, the backend should respond with a warning if the parameters are out of bounds.

google/cloud/aiplatform/training_jobs.py
ivanmkc: Fixed

google/cloud/aiplatform/training_jobs.py
ivanmkc: Yup, done

google/cloud/aiplatform/training_jobs.py
ivanmkc: Done

PR #66, opened 2020-11-13
feat: add support for distributed custom training

sasha-gitg:
Adds support for chief-worker distributed training to custom training. If replica_count > 1 the remainder as provisioned as workers

The _DistributedTrainingSpec can also support more custom provisioning. We will have a follow up PR after we consider how we want to expose more custom provisioning on the API surface.

Note: If library upgrades to 3.7 we should switch the Spec classes to dataclass.

Fixes [b/172369809](b/172369809) 🦕

### DIFF
--- a/google/cloud/aiplatform/training_jobs.py
+++ b/google/cloud/aiplatform/training_jobs.py
@@ -653,6 +653,12 @@ class CustomTrainingJob(base.AiPlatformResourceNoun):
     ) -> Optional[models.Model]:
         """Runs the custom training job.
 
+        Distributed Training Support:
+        If replica count = 1 then one chief replica will be provisioned. If
+        replica_count > 1 the remainder will be provisioned as a worker replica pool.
+        ie: replica_count = 10 will result in 1 chief and 9 workers
+        All replicas have same machine_type, accelerator_type, and accelerator_count
+
         Data fraction splits:
         Any of ``training_fraction_split``, ``validation_fraction_split`` and
         ``test_fraction_split`` may optionally be provided, they must sum to up to 1. If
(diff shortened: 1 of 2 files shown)


google/cloud/aiplatform/training_jobs.py
```diff
@@ -339,6 +339,190 @@ def package_and_copy_to_gcs(
...
+                replica_count=10,
+                machine_type='n1-standard-2',
+                accelerator_count=2,
+                accelerator_type='NVIDIA_TESLA_K80')
+
+    Note that container and python package specs are not stored with this spec.
+    """
+
+    replica_count: int = 0
+    machine_type: str = "n1-standard-2"
```
vinnysenthil, line 357: `n1-standard-2` is supported in Deploy but not for Training. 
`n1-standard-4` is currently the lowest common denominator.

vinnysenthil: - [Machine types for prediction](https://cloud.devsite.corp.google.com/ai-platform-unified/docs/predictions/machine-types#machine_type_comparison)
- [Machine types for training](https://cloud.devsite.corp.google.com/ai-platform-unified/docs/training/configure-compute?db=piac#specifying_machine_types)

vinnysenthil (approved):
LGTM + tests pass, added a note about the default `machine_type`

google/cloud/aiplatform/training_jobs.py
sasha-gitg: Thanks! That's good to know. I was working off the machine_resource.MachineSpec proto comments.

PR #72, opened 2020-11-16
fix: display name length

sirtorry:
b/172932277

### DIFF
--- a/google/cloud/aiplatform/datasets.py
+++ b/google/cloud/aiplatform/datasets.py
@@ -22,7 +22,7 @@ from google.auth import credentials as auth_credentials
 
 from google.cloud.aiplatform import base
 from google.cloud.aiplatform import initializer
-from google.cloud.aiplatform.utils import full_resource_name
+from google.cloud.aiplatform import utils
 from google.cloud.aiplatform import schema
 
 from google.cloud.aiplatform_v1beta1 import GcsSource
@@ -65,7 +65,7 @@ class Dataset(base.AiPlatformResourceNoun):
                 credentials set in aiplatform.init.
         """
 
-        dataset_name = full_resource_name(
+        dataset_name = utils.full_resource_name(
             resource_name=dataset_name,
             resource_noun="datasets",
             project=project,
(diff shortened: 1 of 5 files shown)


google/cloud/aiplatform/utils.py
```diff
@@ -194,7 +194,21 @@ def validate_project(project: str) -> bool:
...
-def validate_region(region: str):
+def validate_display_name(display_name: str):
+    """Verify display name is at most 128 chars and UTF-8 only.
+
+    Args:
+        display_name: display name to verify
+    Raises:
+        ValueError: display name is longer than 128 characters
+    """
+    if display_name is None:
```
sasha-gitg, line 205: What would be the scenario when this is `None`? Also, if this is the case then the type hint should be `Optional[str]`.

google/cloud/aiplatform/utils.py
```diff
@@ -194,7 +194,21 @@ def validate_project(project: str) -> bool:
     raise ValueError("Please provide a valid project ID")
 
 
-def validate_region(region: str):
+def validate_display_name(display_name: str):
+    """Verify display name is at most 128 chars and UTF-8 only.
```
sasha-gitg, line 198: How does this method validate `UTF-8`? If not, I think it's OK to capture in buganizer and TODO here.

tests/unit/aiplatform/test_utils.py
```diff
@@ -193,3 +193,10 @@ def test_full_resource_name_raises_value_error(
             project=project,
             location=location,
         )
+
+
+def test_validate_display_name_raises_value_error():
```
sasha-gitg, line 198: Should also test the non failure case.

google/cloud/aiplatform/datasets.py
```diff
@@ -22,7 +22,7 @@
 
 from google.cloud.aiplatform import base
 from google.cloud.aiplatform import initializer
-from google.cloud.aiplatform.utils import full_resource_name
+from google.cloud.aiplatform import utils
```
sasha-gitg, line 25: Thanks for changing to a module level import!

sasha-gitg (approved):
A few simple comments. Otherwise LGTM!

google/cloud/aiplatform/utils.py
sirtorry: there are some methods that have optional display names. i've moved the None logic into those methods.

PR #78, opened 2020-11-17
feat: add data_labeling samples

morgandu:


kurtisvg (approved):

samples/snippets/create_data_labeling_job_image_segmentation_sample.py
```diff
@@ -0,0 +1,58 @@
...
+from google.protobuf.struct_pb2 import Value
+
+
+def create_data_labeling_job_image_segmentation_sample(
+    project: str,
+    display_name: str,
+    dataset: str,
+    instruction_uri: str,
+    inputs_schema_uri: str,
+    annotation_spec: dict,
```
dizcology, line 27: should this be a `dict` or a `str`?

morgandu: An example usage of annotation_spec for image segmentation labeling is {"color": {"red": 1.0}, "displayName": "rose"}, thus dict is specified instead of str as in i.e. classification.

PR #79, opened 2020-11-17
feat: Job and BatchPredictionJob classes

vinnysenthil:
PR for batch prediction method in Model class to follow once this PR and #66 get merged.

Includes the following:
- Job base class for remaining Job subclasses
  - Adds a shared `status()` method for Job
- BatchPredictionJob class
  - Adds a `iter_outputs()` method which returns either a BQ `QueryJob` or list of GCS `Blob`s
- 7 unit tests covering 94% of `jobs.py`
- Addition of dependency `google-cloud-bigquery`
- Addition of a `constants.py` to have single source of truth on re-used constants (i.e. supported parameters, [API constants](http://google3/java/com/google/cloud/ai/platform/boq/common/constants/ApiConstants.java?rcl=341905673))

Fixes [b/169783178](http://b/169783178), [b/171074104](http://b/171074104) 🦕

### DIFF
--- a/google/cloud/aiplatform/constants.py
+++ b/google/cloud/aiplatform/constants.py
@@ -17,7 +17,7 @@
 
 DEFAULT_REGION = "us-central1"
 SUPPORTED_REGIONS = ("us-central1", "europe-west4", "asia-east1")
-PROD_API_ENDPOINT = "aiplatform.googleapis.com"
+API_BASE_PATH = "aiplatform.googleapis.com"
 
 # Batch Prediction
 BATCH_PREDICTION_INPUT_STORAGE_FORMATS = (
--- a/google/cloud/aiplatform/constants.py
+++ b/google/cloud/aiplatform/constants.py
@@ -1,0 +1,31 @@
+# -*- coding: utf-8 -*-
+
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+DEFAULT_REGION = "us-central1"
+SUPPORTED_REGIONS = ("us-central1", "europe-west4", "asia-east1")
+PROD_API_ENDPOINT = "aiplatform.googleapis.com"
+
+# Batch Prediction
+BATCH_PREDICTION_INPUT_STORAGE_FORMATS = (
+    "jsonl",
+    "csv",
+    "tf-record",
+    "tf-record-gzip",
+    "bigquery",
+    "file-list",
+)
+BATCH_PREDICTION_OUTPUT_STORAGE_FORMATS = ("jsonl", "csv", "bigquery")
--- a/google/cloud/aiplatform/initializer.py
+++ b/google/cloud/aiplatform/initializer.py
@@ -148,7 +148,7 @@ class _Config:
         utils.validate_region(region)
 
         return client_options.ClientOptions(
-            api_endpoint=f"{region}-{prediction}{constants.PROD_API_ENDPOINT}"
+            api_endpoint=f"{region}-{prediction}{constants.API_BASE_PATH}"
         )
 
     def common_location_path(
--- a/google/cloud/aiplatform/initializer.py
+++ b/google/cloud/aiplatform/initializer.py
@@ -24,6 +24,7 @@ import google.auth
 from google.auth import credentials as auth_credentials
 from google.auth.exceptions import GoogleAuthError
 from google.cloud.aiplatform import utils
+from google.cloud.aiplatform import constants
 
 
 class _Config:
@@ -97,7 +98,7 @@ class _Config:
     @property
     def location(self) -> str:
         """Default location."""
-        return self._location or utils.DEFAULT_REGION
+        return self._location or constants.DEFAULT_REGION
 
     @property
     def experiment(self) -> Optional[str]:
@@ -147,7 +148,7 @@ class _Config:
         utils.validate_region(region)
 
         return client_options.ClientOptions(
-            api_endpoint=f"{region}-{prediction}{utils.PROD_API_ENDPOINT}"
+            api_endpoint=f"{region}-{prediction}{constants.PROD_API_ENDPOINT}"
         )
 
     def common_location_path(
--- a/google/cloud/aiplatform/jobs.py
+++ b/google/cloud/aiplatform/jobs.py
@@ -145,7 +145,7 @@ class BatchPredictionJob(_Job):
         self, bq_max_results: Optional[int] = 100
     ) -> Union[Iterable[storage.Blob], Iterable[bigquery.table.RowIterator]]:
         """Returns an Iterable object to traverse the output files, either a list
-        of GCS Blobs or a BigQuery QueryJob depending on the output config set
+        of GCS Blobs or a BigQuery RowIterator depending on the output config set
         when the BatchPredictionJob was created.
 
         Args:
@@ -184,7 +184,7 @@ class BatchPredictionJob(_Job):
             blobs = storage_client.list_blobs(output_info.gcs_output_directory)
             return blobs
 
-        # BigQuery Destination, return QueryJob
+        # BigQuery Destination, return RowIterator
         elif output_info.bigquery_output_dataset:
             bq_client = bigquery.Client()
 
--- a/google/cloud/aiplatform/jobs.py
+++ b/google/cloud/aiplatform/jobs.py
@@ -15,7 +15,7 @@
 # limitations under the License.
 #
 
-from abc import abstractclassmethod
+import abc
 from typing import Iterable, Optional, Union
 
 from google.cloud import storage
@@ -24,9 +24,9 @@ from google.cloud import bigquery
 from google.auth import credentials as auth_credentials
 
 from google.cloud.aiplatform import base
-from google.cloud.aiplatform.gapic import JobState
-from google.cloud.aiplatform.gapic import JobServiceClient
-from google.cloud.aiplatform.utils import full_resource_name
+from google.cloud.aiplatform import utils
+from google.cloud.aiplatform_v1beta1.types import job_state
+from google.cloud.aiplatform_v1beta1.services import job_service
 
 
 class _Job(base.AiPlatformResourceNoun):
@@ -39,18 +39,22 @@ class _Job(base.AiPlatformResourceNoun):
 
     Subclasses requires one class attribute:
 
-    getter_method (str): The name of JobServiceClient getter method for specific
+    _getter_method (str): The name of JobServiceClient getter method for specific
     Job type, i.e. 'get_custom_job' for CustomJob
     """
 
-    client_class = JobServiceClient
+    client_class = job_service.JobServiceClient
     _is_client_prediction_client = False
 
     @property
-    @abstractclassmethod
-    def getter_method(cls) -> str:
+    @abc.abstractclassmethod
+    def _getter_method(cls) -> str:
         """Name of getter method of Job subclass, i.e. 'get_custom_job' for CustomJob"""
         pass
+
+    def _get_job(self, job_name: str):
+        """Returns GAPIC service representation of Job subclass resource"""
+        return getattr(self.api_client, self._getter_method)(name=job_name)
 
     def __init__(
         self,
@@ -78,28 +82,25 @@ class _Job(base.AiPlatformResourceNoun):
                 aiplatform.init will be used.
         """
         super().__init__(project=project, location=location, credentials=credentials)
-        self.job_subclass_getter_method = getattr(self.api_client, self.getter_method)
-        self._gca_resource = self.job_subclass_getter_method(name=valid_job_name)
+        self._gca_resource = self._get_job(job_name=valid_job_name)
 
-    def status(self) -> JobState:
+    def status(self) -> job_state.JobState:
         """Fetch Job again and return the current JobState.
 
         Returns:
-            state (JobState):
+            state (job_state.JobState):
                 Enum that describes the state of a AI Platform job.
         """
 
         # Fetch the Job again for most up-to-date job state
-        self._gca_resource = self.job_subclass_getter_method(
-            name=self._gca_resource.name
-        )
+        self._gca_resource = self._get_job(job_name=self._gca_resource.name)
 
         return self._gca_resource.state
 
 
 class BatchPredictionJob(_Job):
 
-    getter_method = "get_batch_prediction_job"
+    _getter_method = "get_batch_prediction_job"
 
     def __init__(
         self,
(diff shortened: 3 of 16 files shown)


google/cloud/aiplatform/jobs.py
```diff
@@ -15,10 +15,214 @@
...
+    getter_method (str): The name of JobServiceClient getter method for specific
+    Job type, i.e. 'get_custom_job' for CustomJob
+    """
+
+    client_class = JobServiceClient
+    _is_client_prediction_client = False
+
+    @property
+    @abstractclassmethod
+    def getter_method(cls) -> str:
```
sasha-gitg, line 51: Maybe this should be private.

google/cloud/aiplatform/jobs.py
```diff
@@ -15,10 +15,214 @@
...
+from typing import Iterable, Optional, Union
 
-class Job:
+from google.cloud import storage
+from google.cloud import bigquery
+
+from google.auth import credentials as auth_credentials
+
+from google.cloud.aiplatform import base
+from google.cloud.aiplatform.gapic import JobState
```
sasha-gitg, line 27: These should be module level imports and imported directly from the module they are implemented in.

google/cloud/aiplatform/jobs.py
```diff
@@ -15,10 +15,214 @@
...
+        super().__init__(
+            valid_job_name=valid_batch_prediction_job_name,
+            project=project,
+            location=location,
+            credentials=credentials,
+        )
+
+    def iter_outputs(
+        self, bq_query_limit: Optional[int] = 100
+    ) -> Iterable[Union[storage.Blob, bigquery.job.QueryJob]]:
```
sasha-gitg, line 145: The output type hint should be `Union[Iterable[storage.Blob], Iterable[bigquery.job.QueryJob]]` based on the docstring description.

google/cloud/aiplatform/jobs.py
```diff
@@ -15,10 +15,214 @@
 # limitations under the License.
 #
 
+from abc import abstractclassmethod
```
sasha-gitg, line 18: Module level import for `abc`.

google/cloud/aiplatform/jobs.py
```diff
@@ -15,10 +15,214 @@
...
-class Job:
+from google.cloud import storage
+from google.cloud import bigquery
+
+from google.auth import credentials as auth_credentials
+
+from google.cloud.aiplatform import base
+from google.cloud.aiplatform.gapic import JobState
+from google.cloud.aiplatform.gapic import JobServiceClient
+from google.cloud.aiplatform.utils import full_resource_name
```
sasha-gitg, line 29: Module level import here.

google/cloud/aiplatform/utils.py
```diff
@@ -35,9 +35,8 @@
     client as prediction_client,
 )
 
-DEFAULT_REGION = "us-central1"
-SUPPORTED_REGIONS = ("us-central1", "europe-west4", "asia-east1")
-PROD_API_ENDPOINT = "aiplatform.googleapis.com"
+from google.cloud.aiplatform.constants import SUPPORTED_REGIONS
```
sasha-gitg, line 38: Module level import here.

tests/unit/aiplatform/test_jobs.py
```diff
@@ -0,0 +1,268 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import pytest
+
+from importlib import reload
+from unittest.mock import patch
+
+from google.cloud.storage import Blob
```
sasha-gitg, line 23: Module level and direct imports in import statements below.

sasha-gitg (approved):
LGTM! Minor requests.

google/cloud/aiplatform/constants.py
```diff
@@ -0,0 +1,31 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+DEFAULT_REGION = "us-central1"
+SUPPORTED_REGIONS = ("us-central1", "europe-west4", "asia-east1")
+PROD_API_ENDPOINT = "aiplatform.googleapis.com"
```
dizcology, line 20: but this is not really an api endpoint.

vinnysenthil: Does `PROD_API_BASE_PATH` or `PROD_API_BASE_URL` work?

google/cloud/aiplatform/jobs.py
vinnysenthil: Done, also changed `self._job_subclass_getter_method`

google/cloud/aiplatform/jobs.py
```diff
@@ -15,10 +15,214 @@
...
+        if job_status != JobState.JOB_STATE_SUCCEEDED:
+            raise RuntimeError(
+                f"Cannot read outputs until BatchPredictionJob has succeeded, "
+                f"current status: {job_status}"
+            )
+
+        output_info = self._gca_resource.output_info
+
+        # GCS Destination, return Blobs
+        if output_info.gcs_output_directory not in ("", None):
```
dizcology, line 181: maybe just `if not output_info.gcs_output_directory`?

google/cloud/aiplatform/jobs.py
```diff
@@ -15,10 +15,214 @@
...
+        elif output_info.bigquery_output_dataset not in ("", None):
+            bq_client = bigquery.Client()
+            bigquery.Client
+
+            # Format from service is `bq://projectId.bqDatasetId`
+            bq_dataset = output_info.bigquery_output_dataset
+
+            if bq_dataset.startswith("bq://"):
+                bq_dataset = bq_dataset[5:]
+            if bq_dataset.endswith(("/", ".")):
```
dizcology, line 196: in what cases would we have `"."` at the end?

google/cloud/aiplatform/jobs.py
```diff
@@ -15,10 +15,214 @@
...
+                bq_dataset = bq_dataset[:-1]
+
+            # # Split project ID and BQ dataset ID
+            _, bq_dataset_id = bq_dataset.split(".", 1)
+
+            query_limit = f"LIMIT {bq_query_limit}" if bq_query_limit else ""
+            query = f"SELECT * FROM {bq_dataset_id}.predictions {query_limit}"
```
dizcology, line 203: Can we get some folks from BigQuery to take a look?  feels like this could be vulnerable to injection attacks?

google/cloud/aiplatform/initializer.py
```diff
@@ -97,7 +98,7 @@ def project(self) -> str:
     @property
     def location(self) -> str:
         """Default location."""
-        return self._location or utils.DEFAULT_REGION
+        return self._location or constants.DEFAULT_REGION
```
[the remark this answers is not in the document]
vinnysenthil: Perhaps raising a ValueError if they attempt to pass in a Dataset resource name with a different location than the one the SDK is about to use? Also adding a note in the docstring of the Dataset constructor.

google/cloud/aiplatform/jobs.py
vinnysenthil: Unfortunately the default for `output_info` has its fields set to empty string, both GCS and BQ. Another option is to look at `output_config.predictions_format` to see if it's `"bigquery"`, `"jsonl"`, `"csv"`, etc.

google/cloud/aiplatform/jobs.py
vinnysenthil: +1 on BigQuery review. 

I thought about the injection attack issue but this method would only succeed if the user had valid credentials to run this query on BQ instance within the same GCP project, but I could be overlooking something. To prevent a mistake, I could validate `bq_query_limit` is actually an `int`

google/cloud/aiplatform/jobs.py
```diff
@@ -15,10 +15,214 @@
...
+    Job type, i.e. 'get_custom_job' for CustomJob
+    """
+
+    client_class = JobServiceClient
+    _is_client_prediction_client = False
+
+    @property
+    @abstractclassmethod
+    def getter_method(cls) -> str:
+        """Name of getter method of Job subclass, i.e. 'get_custom_job' for CustomJob"""
```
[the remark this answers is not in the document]
vinnysenthil: Perhaps by looking at the subclass name and following the pattern `get_{camel_to_snake_case(subclass_name)}`. Should I make that change?

google/cloud/aiplatform/jobs.py
vinnysenthil: Good point, guess I can remove this conditional altogether since `/` wouldn't show up. IIUC, this `bq://` notation is just for the service to distinguish a BigQuery path and is not used or documented by BigQuery itself.

vinnysenthil:
Hey @tswast , looping you in here for your BigQuery expertise.

Context:
This is for the Model Builder SDK and this PR implements a `BatchPredictionJob` class where users can send many instances to a trained model and get many predictions back. There is an option provide a BQ project to write prediction output directly into. The service creates a new dataset in the provided BQ project and writes the predictions into a table.

We're implementing a helper method `iter_outputs` for users to get an iterable object to traverse the predictions. In the scenario they chose a BQ output, I return a QueryJob that runs `SELECT * {generated_dataset_name}.predictions` and allows a user to pass in an int value for `LIMIT`.

Here are the lines in question:
- https://github.com/googleapis/python-aiplatform/pull/79/files#diff-cc0613c901a9b3ee26167824eeaf859c223b19f1c4c0c48dc01423fb9815e7c5R187-R206
- https://github.com/googleapis/python-aiplatform/pull/79/files#diff-5469ef5507566e6890ac7691b9379a5ce66c72e21b7f657484905fedd076e3b4R234-R243
- https://github.com/googleapis/python-aiplatform/pull/79/files#diff-5469ef5507566e6890ac7691b9379a5ce66c72e21b7f657484905fedd076e3b4R176-R185

PTAL and share your thoughts when you get a chance 😄

google/cloud/aiplatform/jobs.py
tswast: `SELECT *` is actually incredibly wasteful. `bq_client.list_rows` is anywhere from 100% (free) to 80% cheaper (BQ Storage API). ([API reference](https://googleapis.dev/python/bigquery/latest/generated/google.cloud.bigquery.client.Client.html#google.cloud.bigquery.client.Client.list_rows)) It accepts a `max_results` parameter, which can be used instead of a query with a `LIMIT` clause.

Aside: If this was a more complex query. It would be helpful to have some validation here on the dataset ID. Also, I believe you can set a limit using query parameters ([API reference](https://googleapis.dev/python/bigquery/latest/generated/google.cloud.bigquery.job.QueryJobConfig.html#google.cloud.bigquery.job.QueryJobConfig.query_parameters), [docs, with samples](https://cloud.google.com/bigquery/docs/parameterized-queries))

google/cloud/aiplatform/jobs.py
```diff
@@ -15,10 +15,214 @@
...
+        when the BatchPredictionJob was created.
+
+        Args:
+            bq_query_limit: Optional[int] = 100
+                Limit on rows to select from prediction table in BigQuery dataset.
+                Only used when retrieving predictions from a bigquery_destination_prefix.
+                Default is 100.
+
+        Returns:
+            Iterable[Union[storage.Blob, bigquery.job.QueryJob]]:
```
tswast, line 157: If we use `list_rows`, this will be [bigquery.table.RowIterator](https://googleapis.dev/python/bigquery/latest/generated/google.cloud.bigquery.table.RowIterator.html#google.cloud.bigquery.table.RowIterator).

setup.py
```diff
@@ -43,6 +43,7 @@
     install_requires=(
         "google-api-core[grpc] >= 1.22.2, < 2.0.0dev",
         "google-cloud-storage >= 1.32.0, < 2.0.0dev",
+        "google-cloud-bigquery >= 2.3.1, < 3.0.0dev",
```
tswast, line 46: Any reason that an older version won't work? There actually aren't all that many differences between 1.x and 2.x in `google-cloud-bigquery`, just some slightly different types for the BQML query stats.

tswast (changes requested):

google/cloud/aiplatform/jobs.py
vinnysenthil: Thanks for pointing this out - I've made the change to `list_rows()`.

setup.py
vinnysenthil: Dropped lower bound to `1.15.0` - [the earliest ref docs I see with `list_rows()`](https://googleapis.dev/python/bigquery/1.15.0/generated/google.cloud.bigquery.client.Client.html#google.cloud.bigquery.client.Client.list_rows)

google/cloud/aiplatform/jobs.py
```diff
@@ -15,10 +15,212 @@
...
+            project=project,
+            location=location,
+            credentials=credentials,
+        )
+
+    def iter_outputs(
+        self, bq_max_results: Optional[int] = 100
+    ) -> Union[Iterable[storage.Blob], Iterable[bigquery.table.RowIterator]]:
+        """Returns an Iterable object to traverse the output files, either a list
+        of GCS Blobs or a BigQuery QueryJob depending on the output config set
```
tswast, line 148: need to update docstring

google/cloud/aiplatform/jobs.py
```diff
@@ -15,10 +15,212 @@
...
+
+        # GCS Destination, return Blobs
+        if output_info.gcs_output_directory:
+            storage_client = storage.Client()
+            blobs = storage_client.list_blobs(output_info.gcs_output_directory)
+            return blobs
+
+        # BigQuery Destination, return QueryJob
+        elif output_info.bigquery_output_dataset:
+            bq_client = bigquery.Client()
```
tswast, line 189: I'm a bit concerned about this. Ideally you'd re-use the credentials from uCAIP client.

Also, there's a risk of leaking sockets when you create clients on-the-fly. Not as big a deal for REST clients, but definitely a concern for gRPC clients. https://github.com/googleapis/google-cloud-python/issues/9790 https://github.com/googleapis/google-cloud-python/issues/9457

tswast (approved):
LGTM for BigQuery usage. I have some concerns about constructing both Storage and BigQuery clients within a client method, but depending on the performance expectations of this could be considered a minor issue.

google/cloud/aiplatform/jobs.py
vinnysenthil:
```suggestion
            bq_client = bigquery.Client(
                credentials=self.api_client._transport._credentials
            )
```

^ This change would build a BigQuery Client using the same credentials as uCAIP's `JobServiceClient`.

In regards to the leaking sockets, would [the solution referenced in that issue](https://github.com/googleapis/google-cloud-python/issues/9790#issuecomment-558263473) work? See below
```
# Close sockets opened by BQ Client
bq_client._http._auth_request.session.close()
bq_client._http.close()
```

tswast: This change for credentials LGTM. (Storage should get similar treatment).

It's a little trickier in our case, because we want the client to live for the lifetime of the RowIterator.

Unless you want to convert to full list of rows / pandas dataframe before returning? In which case all the API requests would be made here and we could close the client when done (FWIW, the client does have a `close` function in BQ. https://googleapis.dev/python/bigquery/latest/generated/google.cloud.bigquery.client.Client.html#google.cloud.bigquery.client.Client.close)

vinnysenthil: Made the credentials change on both BQ and Storage.

Re: closing connections - this is indeed tricky since the method is meant to return an iterator. However your comment made realize a larger issue of us instantiating a GAPIC client for every instance of a high-level SDK object. I'm capturing this in [b/174111905](http://b/174111905).

Will merge this blocking PR for now, thanks for calling this issue out!

PR #86, opened 2020-11-24
feat: Added predefined splits to custom training jobs

ivanmkc:
This adds predefined splits to custom training jobs.

Fixes https://b.corp.google.com/issues/172366411

## Tasks
- [x] P0 is to get this field supported in custom training. We can allow the service to raise an exception if a non-tabular dataset is passed.
- [ ] P1 to add additional client side validation to ensure the dataset is tabular. <- May do this in a subsequent PR
- [x] Add unit test coverage

## How was this tested
Unit-tested by checking the expected Gapic input data config is created.

### DIFF
--- a/google/cloud/aiplatform/training_jobs.py
+++ b/google/cloud/aiplatform/training_jobs.py
@@ -941,6 +941,7 @@ class CustomTrainingJob(_TrainingJob):
         training_fraction_split: float,
         validation_fraction_split: float,
         test_fraction_split: float,
+        predefined_split_column_name: Optional[str] = None,
     ) -> gca_training_pipeline.InputDataConfig:
         """Constructs a input data config to pass to the training pipeline.
             Override this to create a custom config
@@ -958,6 +959,16 @@ class CustomTrainingJob(_TrainingJob):
                 test_fraction_split (float):
                     The fraction of the input data that is to be
                     used to evaluate the Model. This is ignored if Dataset is not provided.
+                predefined_split_column_name (str):
+                    Optional. The key is a name of one of the Dataset's data
+                    columns. The value of the key (either the label's value or
+                    value in the column) must be one of {``training``,
+                    ``validation``, ``test``}, and it defines to which set the
+                    given piece of data is assigned. If for a piece of data the
+                    key is not present or has an invalid value, that piece is
+                    ignored by the pipeline.
+
+                    Supported only for tabular Datasets.
             """
 
         input_data_config = None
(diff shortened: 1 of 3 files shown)


ivanmkc:
@sasha-gitg how about adding this param to AutoMLTablesTrainingJob? Should we do this in this PR too?

sasha-gitg:
> @sasha-gitg how about adding this param to AutoMLTablesTrainingJob? Should we do this in this PR too?

Yes, that would be great!

sirtorry:
we discussed this for filter split as well. but is no test ok?

google/cloud/aiplatform/training_jobs.py
```diff
@@ -221,6 +266,18 @@ def _run_job(
                 resource ``name``
                 is populated. The Model is always uploaded into the Project
                 and Location in which this pipeline is.
+            gcs_destination (~.io.GcsDestination):
```
ivanmkc, line 269: Consolidated the CustomTrainingJob's version of this function by moving the parameter here.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -935,54 +994,7 @@ def __init__(
         self._script_path = script_path
         self._staging_bucket = staging_bucket
 
-    def _create_input_data_config(
```
ivanmkc, line 938: Consolidated this function by removing it and augmenting the base class's version

google/cloud/aiplatform/training_jobs.py
```diff
@@ -1153,7 +1176,11 @@ def run(
             training_fraction_split=training_fraction_split,
             validation_fraction_split=validation_fraction_split,
             test_fraction_split=test_fraction_split,
+            predefined_split_column_name=predefined_split_column_name,
             model=managed_model,
+            gcs_destination=gca_io.GcsDestination(
```
sasha-gitg, line 1181: Prefer to pass the gcs string and create the GcsDestination proto in the base class.

sasha-gitg (approved):
LGTM! One minor comment.

google/cloud/aiplatform/training_jobs.py
ivanmkc: Will do

sasha-gitg:
> [ ] P1 to add additional client side validation to ensure the dataset is tabular. <- May do this in a subsequent PR

Is there a ticket open for this?

google/cloud/aiplatform/training_jobs.py
ivanmkc: I called it `gcs_destination_uri_prefix`

ivanmkc:
@sasha-gitg https://b.corp.google.com/issues/174495379

PR #87, opened 2020-11-24
AutoMLTablesTrainingJob: Make model_display_name optional and use the job's display_name if not passed in.

ivanmkc:
- Uses the job display name if is not passed in.
- Added unit test

Fixes https://b.corp.google.com/issues/174146765

### DIFF
--- a/google/cloud/aiplatform/training_jobs.py
+++ b/google/cloud/aiplatform/training_jobs.py
@@ -1273,7 +1273,6 @@ class AutoMLTablesTrainingJob(_TrainingJob):
 
     def run(
         self,
-        model_display_name: str,
         dataset: datasets.Dataset,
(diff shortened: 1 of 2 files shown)


google/cloud/aiplatform/training_jobs.py
```diff
@@ -1294,9 +1294,11 @@ def run(
 
         Args:
             model_display_name (str):
```
sasha-gitg, line 1296: Move this arg docstring down so it's in the same order as the method signature.

sasha-gitg (approved):

google/cloud/aiplatform/training_jobs.py
ivanmkc: Will do

PR #97, opened 2020-11-30
feat: set user_agent

dizcology:
Similar to https://github.com/googleapis/python-bigquery/blob/809e4a27b94ba30c10e0c9a7e89576a9de9fda2b/google/cloud/bigquery/magics/magics.py#L596


sasha-gitg (approved):
LGTM!

PR #98, opened 2020-12-01
feat: Add Model.batch_predict()

vinnysenthil:
List of changes:
- Added a `batch_predict()` method to the Model class
  - Created two TODO issues: [b/174502675](http://b/174502675) & [b/174502913](http://b/174502913)
- 8 unit tests for `Model.batch_predict()`
- Resolves a bug in `BatchPredictionJob.iter_outputs()` that did not call `list_blobs()` in the storage client correctly

Fixes [b/171074104](http://b/171074104) 🦕


google/cloud/aiplatform/models.py
```diff
@@ -314,6 +316,266 @@ def deploy(
 
         return endpoint
 
+    def batch_predict(
+        self,
+        job_display_name: str,
+        gcs_source: Optional[Sequence[str]] = None,
+        bigquery_source: Optional[str] = None,
```
vinnysenthil, line 323: Should these merged into a single parameter called `source` that distinguished between the two based on prefix? I.e. `bq://` or `gs://`

google/cloud/aiplatform/models.py
```diff
@@ -314,6 +316,266 @@ def deploy(
...
+        instances_format: Optional[str] = "jsonl",
+        gcs_destination_prefix: Optional[str] = None,
+        bigquery_destination_prefix: Optional[str] = None,
```
vinnysenthil, line 326: Same as above - should this be merged into a `destination_prefix` parameter?

google/cloud/aiplatform/models.py
```diff
@@ -314,6 +316,266 @@ def deploy(
...
+        predictions_format: Optional[str] = "jsonl",
+        model_parameters: Optional[Dict] = None,
+        machine_type: Optional[str] = None,
+        accelerator_type: Optional[str] = None,
+        accelerator_count: Optional[int] = None,
+        starting_replica_count: Optional[int] = None,
+        max_replica_count: Optional[int] = None,
+        labels: Optional[dict] = None,
+        location: Optional[str] = None,
```
vinnysenthil, line 335: Should we drop the `location` param since the Model is already associated to a region?

vinnysenthil (commented):
Some questions for the reviewers

google/cloud/aiplatform/models.py
sasha-gitg: Yes, if the service doesn't allow execution of the batch prediction job from a different location than the Model's location.

google/cloud/aiplatform/models.py
```diff
@@ -314,6 +316,266 @@ def deploy(
...
+                prediction succeeded, it has columns as per a concatenation
+                of the Model's instance and prediction schemata. The
+                ``errors`` table contains rows for which the prediction has
+                failed, it has instance columns, as per the instance schema,
+                followed by a single "errors" column, which as values has
+                ```google.rpc.Status`` <Status>`__ represented as a STRUCT,
+                and containing only ``code`` and ``message``.
+            predictions_format: Optional[str] = "jsonl"
+                Required. The format in which AI Platform gives the
+                predictions, must be one of "jsonl", "csv", or "bigquery".
```
sasha-gitg, line 411: If `None` is not valid then this type hint should just be `str`.

I think this should also say that if BQ destination is provided then this default to `bigquery`. Same for source and instances.

google/cloud/aiplatform/models.py
```diff
@@ -314,6 +316,266 @@ def deploy(
...
+                of any UTF-8 characters.
+            gcs_source: Optional[Sequence[str]] = None
+                Google Cloud Storage URI(-s) to your instances to run
+                batch prediction on. They must match `instances_format`.
+                May contain wildcards. For more information on wildcards, see
+                https://cloud.google.com/storage/docs/gsutil/addlhelp/WildcardNames.
+            bigquery_source: Optional[str] = None
+                BigQuery URI to a table, up to 2000 characters long. For example:
+                `projectId.bqDatasetId.bqTableId`
+            instances_format: Optional[str] = "jsonl"
```
sasha-gitg, line 364: If `None` is not valid then this type hint should just be `str`.

sasha-gitg (approved):
LGTM with comments.

google/cloud/aiplatform/jobs.py
```diff
@@ -186,7 +186,25 @@ def iter_outputs(
                 credentials=self.api_client._transport._credentials
             )
 
-            blobs = storage_client.list_blobs(output_info.gcs_output_directory)
+            gcs_path = output_info.gcs_output_directory
+
+            if gcs_path.startswith("gs://"):
```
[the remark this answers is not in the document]
vinnysenthil: Added [utility function](https://github.com/googleapis/python-aiplatform/blob/4a2e5b6416ec132362acf794605a7b8bda5f7e80/google/cloud/aiplatform/utils.py#L233-L264)

PR #99, opened 2020-12-01
fix: Added client check for a tabular dataset when using pre-defined splits

ivanmkc:
According to the yaml file, pre-defined splits can only be used with a tabular dataset. A ValueError is now raised if that's not the case.

fixes: https://b.corp.google.com/issues/174495379

### DIFF
--- a/tests/unit/aiplatform/test_automl_tabular_training_jobs.py
+++ b/tests/unit/aiplatform/test_automl_tabular_training_jobs.py
@@ -166,7 +166,10 @@ class TestAutoMLTabularTrainingJob:
         return ds
 
     def test_run_call_pipeline_service_create(
-        self, mock_pipeline_service_create, mock_dataset_tabular, mock_model_service_get,
+        self,
+        mock_pipeline_service_create,
+        mock_dataset_tabular,
+        mock_model_service_get,
     ):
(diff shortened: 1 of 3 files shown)


tests/unit/aiplatform/test_automl_tabular_training_jobs.py
```diff
@@ -259,8 +284,33 @@ def test_run_call_pipeline_if_no_model_display_name(
             training_pipeline=true_training_pipeline,
         )
 
+    def test_run_called_predefined_split_nontabular_raises(
```
ivanmkc, line 293: Added test to test that the exception raises properly.

tests/unit/aiplatform/test_training_jobs.py
```diff
@@ -387,6 +392,13 @@ def mock_python_package_to_gcs(self):
     def mock_dataset(self):
         ds = mock.MagicMock(datasets.Dataset)
         ds.name = _TEST_DATASET_NAME
+        ds._gca_resource = GapicDataset(
```
ivanmkc, line 395: Added this as the client SDK tests for `Dataset._gca_resource.metadata_schema_uri`

sasha-gitg (approved):
LGTM

PR #100, opened 2020-12-01
feat: Added predict schemata support to Model.upload

ivanmkc:
Added predict schemata support to Model.upload. 

Fixes https://b.corp.google.com/issues/170979552


ivanmkc:
@sasha-gitg this is my guess at what the feature requirements are. 

@dizcology mentioned that that the URI's aren't actually used? If so, then should we note that in the doc strings so users don't waste time debugging them if something goes wrong?

sasha-gitg (approved):
LGTM

PR #116, opened 2020-12-03
fix: Added to comments for predict_schemata fields.

ivanmkc:
According to backend engineers, these fields are unused.

Fixes https://b.corp.google.com/issues/170979552

### DIFF
--- a/google/cloud/aiplatform/models.py
+++ b/google/cloud/aiplatform/models.py
@@ -832,7 +832,7 @@ class Endpoint(base.AiPlatformResourceNounWithFutureManager):
 
         Args:
             instances (List):
-                Required. Required. The instances that are the input to the
+                Required. The instances that are the input to the
                 prediction call. A DeployedModel may have an upper limit
                 on the number of instances it supports per request, and
                 when it is exceeded the prediction call errors in case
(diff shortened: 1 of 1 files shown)


google/cloud/aiplatform/models.py
```diff
@@ -1129,7 +1135,7 @@ def predict(self, instances: List, parameters: Optional[Dict] = None) -> Predict
 
         Args:
             instances (List):
-                Required. Required. The instances that are the input to the
```
ivanmkc, line 1132: Just fixed this too since I saw it.

google/cloud/aiplatform/models.py
```diff
@@ -158,7 +158,9 @@ def upload(
                 must be supported by it. If not specified a standard HTTP path will be
                 used by AI Platform.
             instance_schema_uri (str):
-                Optional. Points to a YAML file stored on Google Cloud
+                Optional. Currently not used for any purpose.
```
sasha-gitg, line 161: Prefer "Currently not used." and add to the end of the arg string.

sasha-gitg (approved):
LGTM. One comment on language and formatting.

PR #118, opened 2020-12-03
feat: add optional async support

sasha-gitg:
Adds optional sync support to resource noun methods. Methods with optional sync support are decorated with `base.optional_sync`.

Main changes to support optional sync are in `base.py`.

Test have been adapted and additional end to end unit test was created.

Fixes b/171275584 🦕


google/cloud/aiplatform/base.py
```diff
@@ -16,13 +16,196 @@
...
 
 
+class FutureManager(metaclass=abc.ABCMeta):
+    """Tracks concurrent futures against this object."""
+
+    def __init__(self):
+        self.__latest_future_lock = threading.Lock()
+
+        # Always points to the latest future. All submitted futures will always
+        # form a dependency on the latest future.
```
dizcology, line 37: Does this mean that if future1 and future2 are submitted in that order, then self.__latest_future will point to future1, while future2 will be remembered as a dependency of future1.  The moment future1 is resolved, future2 becomes self.__latest_future?

google/cloud/aiplatform/base.py
```diff
@@ -16,13 +16,196 @@
...
+    def _complete_future(self, future: futures.Future):
+        """Checks for exception of future and removes the pointer if it's still latest.
+
+        Args:
+            future (futures.Future): Required. A future to complete.
+        """
+
+        with self.__latest_future_lock:
+            try:
+                future.result()  # raises
```
dizcology, line 59: Could this `result` call contain wait-until-complete in it like LRO?

google/cloud/aiplatform/base.py
```diff
@@ -16,13 +16,196 @@
...
+        if future:
+            futures.wait([future], return_when=futures.FIRST_EXCEPTION)
+
+        self._raise_future_exception()
+
+    @property
+    def _latest_future(self) -> Optional[futures.Future]:
+        """Get the latest future if it exists"""
+        with self.__latest_future_lock:
+            return self.__latest_future
```
dizcology, line 87: How does locking work here? Is it simply that an error would be raised if someone else holds the lock at the execution time of the return statement here?

google/cloud/aiplatform/base.py
```diff
@@ -16,13 +16,196 @@
...
+        if additional_dependencies:
+            deps.extend(additional_dependencies)
+
+        with self.__latest_future_lock:
+
+            # form a dependency on the latest future of this object
+            if self.__latest_future:
+                deps.append(self.__latest_future)
+
+            self.__latest_future = initializer.global_pool.submit(
```
dizcology, line 176: Do we maintain a single global pool that contains all the futures for different resources?  (as opposed to every resource noun instance maintains its own graph of futures)

Does this make debugging more difficult?

google/cloud/aiplatform/base.py
```diff
@@ -108,3 +292,231 @@ def resource_name(self) -> str:
     def display_name(self) -> str:
         """Display name of this resource."""
         return self._gca_resource.display_name
+
+
+class AiPlatformResourceNounWithFutureManager(AiPlatformResourceNoun, FutureManager):
```
dizcology, line 297: Will any resource still be under AiPlatformResourceNoun (without FutureManager)?  Such a resource would not have some public method such as `wait`.  Do the users have an easy way to tell if something is with/without future manager?

google/cloud/aiplatform/base.py
dizcology: Related: do we see the need for the user to be able to inspect the completed and yet-to-complete futures for debugging?

sasha-gitg: > The moment future1 is resolved, future2 becomes self.__latest_future?

`future2` will become the latest future immediately so any down stream call against this noun can form a dependency on future2. 

I'll change this comment to say:

```python
# Always points to the latest future. All submitted futures will always
# form a dependency on the latest future and then be set as the latest future.
```

google/cloud/aiplatform/base.py
sasha-gitg: The underlying future could be executing code that calls and waits on an LRO and that happens in this implementation. Additionally we could return an LRO as the result but for right now we are only returning either a AiPlatformResourceNounWithFutureManager class or None.

google/cloud/aiplatform/base.py
sasha-gitg: If someone else holds the lock it will block until that lock is released.

google/cloud/aiplatform/base.py
sasha-gitg: We maintain a global pool. I'm not sure if it would make debugging less difficult if we maintained a pool per resource because a resource noun's future can be dependent on another resource noun's future. I do think there's additional work that can be done to provide more information about the current state of all futures.

google/cloud/aiplatform/base.py
sasha-gitg: Every resource noun should be with FutureManager.

google/cloud/aiplatform/base.py
dizcology: Yes that makes sense - it would be great if we have a way to allow the user to inspect all futures of the current session, but that may be more of a feature request.

google/cloud/aiplatform/base.py
dizcology: Depending on the user's perspective, the term "latest" might be interpreted as future2 in the example above even before future1 has been completed, since future2 was submitted the latest.  But i don't think this is a serious problem.

dizcology (commented):
Approving this now so we can merge.

dizcology (approved):

google/cloud/aiplatform/models.py
```diff
@@ -345,589 +468,768 @@ def deploy(
...
+                ][google.cloud.aiplatform.v1beta1.DeployedModel.model]
+                [PredictSchemata's][google.cloud.aiplatform.v1beta1.Model.predict_schemata]
+                ``parameters_schema_uri``.
+        Returns:
+            prediction: Prediction with returned predictions and Model Id.
 
-        return jobs.BatchPredictionJob(
-            batch_prediction_job_name=new_batch_prediction_job_name
+        """
+        prediction_response = self._prediction_client.predict(
```
ivanmkc, line 857: I noticed that this raises the following when the training job is not finished when this is called:

AttributeError: 'Endpoint' object has no attribute '_prediction_client'

A check for pending futures would be nice.

sasha-gitg: I'll add that. I considered this earlier but there's a tradeoff because the predict API can be latency sensitive for some use cases.

google/cloud/aiplatform/models.py
```diff
@@ -43,274 +43,397 @@
 from google.protobuf import json_format
 
 
-class Model(base.AiPlatformResourceNoun):
+class Prediction(NamedTuple):
```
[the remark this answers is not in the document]
sasha-gitg: Capture as a b/175232224

PR #122, opened 2020-12-06
feat: adds function/method enhancements, demo samples

telpirion:
This PR does the following:

- adds the `to_value()`, `from_value()`, and `from_map()` utility functions
- assigns the utility functions to enhanced types (at the module level)
- adds a new conversion rule for marshaling `protobuf.Value` objects
- demonstrates two different usages of the enhanced classes (dict with to_value(); direct assignment with marshaling)


tests/unit/test_value_converter.py
```diff
@@ -0,0 +1,75 @@
...
+    def test_convert_value_to_message(self):
+        actual_from_value_output = value_converter.from_value(
+            SomeMessage, self.input_value)
+        expected_type = SomeMessage(self.input_dict)
+
+        # Following assert fails.
+        # `expected_type` is `test_value_converter.SomeMessage` while
+        # `actual_from_value_output` is just `SomeMessage`
+        # Use `isinstance()` instead.
+        #assert(type(actual_from_value_output) is type(expected_type))
```
telpirion, line 67: I tried changing this to `isinstance()`, but instead get this message:

```
E       AssertionError: assert False
E        +  where False = isinstance(test_str: "Omnia Gallia est divisa"\ntest_int64: 3\ntest_bool: true\n, SomeMessage)

```

google/cloud/aiplatform_helpers/value_converter.py
```diff
@@ -0,0 +1,74 @@
...
+      the message as a :class:`~google.protobuf.struct_pb2.Value` object
+    """
+    def is_prop(prop):
+        if prop[0].isupper():
+            return False
+        if prop.startswith('_'):
+            return False
+        return True
+
+    props = list(filter(is_prop, dir(self._pb)))
```
dizcology, line 40: Looks like the intention here is to collect all the field names - is there a better to do that than relying on attribute name's first character?

google/cloud/aiplatform_helpers/value_converter.py
```diff
@@ -0,0 +1,74 @@
...
+
+    props_dict = {}
+    for prop in props:
+        props_dict[prop] = getattr(self._pb, prop)
+
+    return json_format.ParseDict(props_dict, Value())
```
dizcology, line 46: does this work if some of the values of `props_dict` are nested proto messages?

samples/snippets/create_training_pipeline_image_classification_sample.py
```diff
@@ -14,8 +14,8 @@
 
 # [START aiplatform_create_training_pipeline_image_classification_sample]
 from google.cloud import aiplatform
-from google.protobuf import json_format
-from google.protobuf.struct_pb2 import Value
+from google.cloud.aiplatform.v1beta1.schema.trainingjob import definition
+ModelType = definition.AutoMlImageClassificationInputs().ModelType
```
dizcology, line 18: I think we don't have to create an instance of `AutoMlImageClassificationInputs` here.

samples/snippets/create_training_pipeline_image_classification_sample.py
```diff
@@ -14,8 +14,8 @@
 
 # [START aiplatform_create_training_pipeline_image_classification_sample]
 from google.cloud import aiplatform
-from google.protobuf import json_format
-from google.protobuf.struct_pb2 import Value
+from google.cloud.aiplatform.v1beta1.schema.trainingjob import definition
```
dizcology, line 17: is it possible to remove `v1beta` by exporting schema under cloud.aiplatform? (without moving all the files)

samples/snippets/create_training_pipeline_image_classification_sample.py
```diff
@@ -30,18 +30,18 @@ def create_training_pipeline_image_classification_sample(
...
     client = aiplatform.gapic.PipelineServiceClient(client_options=client_options)
-    training_task_inputs_dict = {
-        "multiLabel": True,
-        "modelType": "CLOUD",
-        "budgetMilliNodeHours": 8000,
-        "disableEarlyStopping": False,
-    }
-    training_task_inputs = json_format.ParseDict(training_task_inputs_dict, Value())
+
+    icn_training_inputs = definition.AutoMlImageClassificationInputs(
```
dizcology, line 34: Note that here we introduce an inconsistent style of using instance of a particular python class, whereas the rest of the sample is in python dicts. I think technically we could keep using python dict here too (with only the change of camelCase to snake_case for dict keys).

 @leahecole let us know if this is fine according to sample style guidelines.  Several more samples will be updated and follow the same pattern as the two samples of this PR.

samples/snippets/create_training_pipeline_image_classification_sample.py
```diff
@@ -30,18 +30,18 @@ def create_training_pipeline_image_classification_sample(
...
+        model_type=ModelType.CLOUD,
+        budget_milli_node_hours=8000,
+        disable_early_stopping=False
+    )
 
     training_pipeline = {
         "display_name": display_name,
         "training_task_definition": "gs://google-cloud-aiplatform/schema/trainingjob/definition/automl_image_classification_1.0.0.yaml",
-        "training_task_inputs": training_task_inputs,
+        "training_task_inputs": icn_training_inputs.to_value(),
```
dizcology, line 44: I rather prefer not to have additional method calls here.

samples/snippets/predict_image_classification_sample.py
```diff
@@ -36,25 +37,29 @@ def predict_image_classification_sample(
...
-    parameters_dict = {"confidence_threshold": 0.5, "max_predictions": 5}
-    parameters = json_format.ParseDict(parameters_dict, Value())
+    instance_obj = instance.ImageClassificationPredictionInstance({
+        "content": encoded_content})
+
+    instance_val = instance_obj.to_value()
+    instances = [instance_val]
+
+    params_obj = params.ImageClassificationPredictionParams({
+        "confidence_threshold": 0.5, "max_predictions": 5})
```
dizcology, line 48: it seems more common to pass these in as parameters as opposed to a dict, as is done in the other sample of this PR.

tests/unit/test_enhanced_types.py
```diff
@@ -0,0 +1,37 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from __future__ import absolute_import
+import unittest
+
+from google.cloud.aiplatform.v1beta1.schema.trainingjob import definition
+ModelType = definition.AutoMlImageClassificationInputs().ModelType
+
+
+class EnhancedTypesTests(unittest.TestCase):
```
dizcology, line 21: The other library unit tests are written in the style of plain pytest functions.  Unless that is not feasible for what you are testing here, please follow the same convention.  Also perhaps add another folder under `tests/unit` to house tests that specifically have to do with enhanced types, and add yourself as codeowner of that folder (if tests are allowed to have codeowners).

tests/unit/test_value_converter.py
dizcology: This feels like an artifact of how the tests are collected and run.

tests/unit/test_value_converter.py
```diff
@@ -0,0 +1,90 @@
...
+
+        # Check property-level ("duck-typing") equivalency
+        assert(actual_from_value_output.test_str == expected_type.test_str)
+        assert(actual_from_value_output.test_bool == expected_type.test_bool)
+        assert(actual_from_value_output.test_int64 == expected_type.test_int64)
+
+    def test_convert_map_to_message(self):
+        message_with_map = SomeInType()
+        message_with_map.test_map['test_int'] = 42
+        map_composite = message_with_map.test_map
```
dizcology, line 81: Does this actually create a MapComposite object?  I thought that is need only when the value of the map is itself a message (as opposed to, say, an int).

samples/snippets/create_training_pipeline_image_classification_sample.py
dizcology: also note that this is simply a style issue with some hidden implication on sample generation.  please feel free to leave it as is for sample review.

tests/unit/test_value_converter.py
telpirion: Yes, that does create a MapComposite object. I tested it a couple of times, both in ipython and debugging this test with ipdb.

tests/unit/test_value_converter.py
telpirion: It could be. I'm inclined to leave this test as-is for now (since we test property-level equivalency). I'll log a GitHub issue and assign it to myself to further investigate this problem.

Does that work for you?

google/cloud/aiplatform/v1beta1/schema/predict/instance/__init__.py
```diff
@@ -14,6 +14,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+from google.cloud.aiplatform_helpers import add_methods_to_classes_in_package
```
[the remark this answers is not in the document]
telpirion: The value_converter.py module is intended to be public. It will be helpful for tabular developers who need to format their prediction instances, for example.

I'll change the name of the methods intended to be private so that they have a leading underscore in their names.

tests/unit/test_value_converter.py
dizcology: Sounds good.

google/cloud/aiplatform/v1beta1/schema/predict/instance/__init__.py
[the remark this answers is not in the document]
dizcology: I see - perhaps `add_methods_to_classes_in_package` should be in a private module, where as `value_converter` a public module.  in that case perhaps a nested submodule `aiplatform.helpers.value_converter` would be preferred over `aiplatform_helpers.value_converter`.

tests/unit/test_value_converter.py
telpirion: Done.

google/cloud/aiplatform/v1beta1/schema/predict/instance/__init__.py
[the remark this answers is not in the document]
telpirion: Done.

google/cloud/aiplatform/v1beta1/schema/predict/prediction_v1beta1/types/text_sentiment.py
```diff
@@ -17,8 +17,8 @@
 
 import proto  # type: ignore
 
-
-from google.cloud.aiplatform.v1beta1.schema.predict.instance import text_sentiment_pb2 as gcaspi_text_sentiment  # type: ignore
+# DO NOT OVERWRITE FOLLOWING LINE: it was manually edited.
+from google.cloud.aiplatform.v1beta1.schema.predict.instance import TextSentimentPredictionInstance
```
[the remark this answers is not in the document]
telpirion: Done.

google/cloud/aiplatform_helpers/__init__.py
```diff
@@ -0,0 +1,72 @@
...
+                    for name, cls in pkg.__dict__.items()
+                    if isinstance(cls, type)])
+
+    for class_name, cls in classes.items():
+        # Add to_value() method to class with docstring
+        setattr(cls, 'to_value', to_value)
+        cls.to_value.__doc__ = to_value.__doc__
+
+        # Add from_value() method to class with docstring
+        cls.from_value = add_from_value_to_class(cls)
```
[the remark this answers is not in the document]
telpirion: Changed.

I was just trying different methods of assigning members dynamically; forgot to standardize on one technique.

google/cloud/aiplatform_helpers/value_converter.py
telpirion: Yeah, that was my hack for trying to get around the "int64s as strings" issue.

However, playing with the Java library the other day, I think that sending int64 values as strings might not be as big a deal as I originally though. I'm going to switch this to a simple call to `json_format.ParseDict()` and make sure that I can still train a model.

telpirion: This apparently isn't needed! I've removed this code.

google/cloud/aiplatform_helpers/value_converter.py
telpirion: Removed this bit.

samples/snippets/create_training_pipeline_image_classification_sample.py
telpirion: Changed.

samples/snippets/predict_image_classification_sample.py
telpirion: Done.

tests/unit/test_enhanced_types.py
telpirion: Both done: moved tests, switched to pytest rather than unittest.

samples/snippets/create_training_pipeline_image_classification_sample.py
leahecole: General consensus amongst the owners was preference to have generated classes for API resources - having spell check and autocomplete as well as knowing where to look in reference docs is helpful

If you're using a user-defined object with arbitrary properties, a dict may be simpler.

samples/snippets/predict_image_classification_sample.py
```diff
@@ -36,25 +37,29 @@ def predict_image_classification_sample(
...
+
     endpoint = client.endpoint_path(
         project=project, location=location, endpoint=endpoint_id
     )
     response = client.predict(
-        endpoint=endpoint, instances=instances, parameters=parameters
+        endpoint=endpoint, instances=instances, parameters=params_obj
     )
     print("response")
     print(" deployed_model_id:", response.deployed_model_id)
```
leahecole, line 57: nit - Is there a reason for the extra space at the beginning of this print statement?

samples/snippets/predict_image_classification_sample_test.py
```diff
@@ -31,4 +31,4 @@ def test_ucaip_generated_predict_image_classification_sample(capsys):
     )
 
     out, _ = capsys.readouterr()
-    assert 'string_value: "daisy"' in out
+    assert 'deployed_model_id:' in out
```
leahecole, line 34: Why did this test case change? Is there any chance this could lead to a false positive if no model ID is returned? Or will the sample straight up fail before it gets to this print statement?

samples/snippets/create_training_pipeline_image_classification_sample.py
dizcology: This may be a place where we need to avoid imposing consistency across the board.  For example explicitly construct instances can make certain samples (e.g. https://github.com/googleapis/python-aiplatform/blob/master/samples/snippets/create_hyperparameter_tuning_job_python_package_sample.py) much more difficult to read than dicts, and in some cases we are forced to use instances (e.g. https://github.com/googleapis/python-aiplatform/blob/master/samples/snippets/upload_model_explain_image_managed_container_sample.py) because of dependency issues.

tests/unit/enhanced_library/CODEOWNERS
```diff
@@ -0,0 +1,2 @@
+# Tests for enhancements to the AI Platform library for Python.
+*       @telpirion
```
leahecole, line 2: @busunkim96 should this instead be at `.github/CODEOWNERS` ?

and then be formatted as 
(this might have whitespace errors but I'm just trying to convey the idea here :) )
`/tests/unit/enhanced_library      @telpirion`

samples/snippets/create_training_pipeline_image_classification_sample.py
telpirion: Removed.

samples/snippets/create_training_pipeline_image_classification_sample.py
telpirion: Not very easily, no.

samples/snippets/create_training_pipeline_image_classification_sample.py
telpirion: Just to clarify: I'm keeping this sample as-is?

samples/snippets/predict_image_classification_sample_test.py
telpirion: A couple of reasons: we want to avoid testing for the output of models, since retraining can cause the output to change. 

No, a model ID must be returned as part of the online prediction--you can't have a prediction without a model! The sample will fail if you attempt to send a prediction request to an endpoint that has no model deployed to it.

samples/snippets/predict_image_classification_sample.py
telpirion: I had a \t character in there earlier, but it got dropped accidentally. Added it back.

tests/unit/enhanced_library/CODEOWNERS
telpirion: Done.

dizcology (approved):

PR #125, opened 2020-12-08
Fix: refactor class constructor for retrieving resource

morgandu:
- Added property and abstract method _getter_method and _resource_noun, implemented method _get_gca_resource to class AiPlatformResourceNoun; 
- Added _resource_noun, _getter_method, to Dataset, Model, Endpoint, subclasses of _Job, _TrainingJob, - Remove/moved (_)get_* and utils.full_resource_name in class constructor to self._get_gca_resource to Dataset, Model, Endpoint, _Job; 
- Updated test

Fixes https://b.corp.google.com/issues/173531466

### DIFF
--- a/google/cloud/aiplatform/base.py
+++ b/google/cloud/aiplatform/base.py
@@ -106,7 +106,10 @@ class AiPlatformResourceNoun(metaclass=abc.ABCMeta):
             prediction_client=cls._is_client_prediction_client,
         )
 
-    def _get_gca_resource(self, resource_name):
+    def _get_gca_resource(
+        self,
+        resource_name: str
+    ):
         """Returns GAPIC service representation of client class resource."""
         """
         Args:
@@ -121,9 +124,15 @@ class AiPlatformResourceNoun(metaclass=abc.ABCMeta):
             location=self.location,
         )
 
-        self._gca_resource = getattr(self.api_client, self._getter_method)(
+        return getattr(self.api_client, self._getter_method)(
             name=resource_name
         )
+
+    def _sync_gca_resource(self):
+        """Sync GAPIC service representation of client class resource."""
+
+        self._gca_resource = self._get_gca_resource(
+            resource_name=self.resource_name)
 
     @property
     def name(self) -> str:
--- a/google/cloud/aiplatform/base.py
+++ b/google/cloud/aiplatform/base.py
@@ -18,6 +18,8 @@
 import abc
 from typing import Optional
 
+import proto
+
 from google.auth import credentials as auth_credentials
 from google.cloud.aiplatform import utils
 from google.cloud.aiplatform import initializer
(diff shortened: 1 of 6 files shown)


google/cloud/aiplatform/base.py
```diff
@@ -94,6 +106,25 @@ def _instantiate_client(
             prediction_client=cls._is_client_prediction_client,
         )
 
+    def _get_gca_resource(self, resource_name):
```
sasha-gitg, line 109: resource_name str type hint

google/cloud/aiplatform/jobs.py
```diff
@@ -46,19 +46,10 @@ class _Job(base.AiPlatformResourceNoun):
...
-
-    def _get_job(self, job_name: str):
-        """Returns GAPIC service representation of Job subclass resource"""
-        return getattr(self.api_client, self._getter_method)(name=job_name)
-
     def __init__(
         self,
-        valid_job_name: str,
+        job_name: str,
+        job_type: str,
```
sasha-gitg, line 52: `job_type` seems unnecessary because this will always be subclassed the that value equals `cls._resource_name`

google/cloud/aiplatform/jobs.py
```diff
@@ -82,7 +76,14 @@ def __init__(
                 aiplatform.init will be used.
         """
         super().__init__(project=project, location=location, credentials=credentials)
-        self._gca_resource = self._get_job(job_name=valid_job_name)
+        if (
```
sasha-gitg, line 79: Why not rely on the `utils.full_resource_name` in `_get_gca_resource`?

google/cloud/aiplatform/base.py
```diff
@@ -94,6 +106,25 @@ def _instantiate_client(
...
+        """Returns GAPIC service representation of client class resource."""
```
sasha-gitg, line 110: This doesn't necessarily return the resource it sets the `self._gca_resource` attribute.

It may be worth breaking this down into two methods.

`_get_gca_resource` that gets the resource and returns it.

and `_sync_gca_resource` which calls `_get_gca_resource` and sets it to `self._gca_resource`.

sasha-gitg (changes requested):
Looks great! Have some minor comments that should be resolved.

google/cloud/aiplatform/base.py
```diff
@@ -94,6 +106,34 @@ def _instantiate_client(
             prediction_client=cls._is_client_prediction_client,
         )
 
+    def _get_gca_resource(
+        self,
+        resource_name: str
+    ):
```
sasha-gitg, line 112: Return type should be `proto.Message`.

sasha-gitg (approved):
LGTM. Please resolve lint issues and resolve conflicts against the `dev` branch.

PR #126, opened 2020-12-08
test: Dataset integration tests

vinnysenthil:
### A summary of changes:
- Added six tests for the Dataset class calling the live service
  - `test_get_existing_dataset`
  - `test_get_nonexistent_dataset`
  - `test_get_new_dataset_and_import`
  - `test_create_and_import_image_dataset`
  - `test_create_tabular_dataset`
  - `test_export_data`
- Fixtures for setup/teardown of empty datasets, staging buckets
- The average runtime is around **10 minutes**, this could be decreased by using text datasets for testing all non-tabular method calls.
- Added a [text metadata schema](https://github.com/googleapis/python-aiplatform/compare/dev...vinnysenthil:dataset-integration-tests#diff-2278d7930d857c9fa1127aa7b82b80b5f1291c0fa5bde2cbf82da033b6bc3305R33)

Fixes [b/172285155](http://b/172285155) 🦕


tests/integration/aiplatform/test_dataset.py
```diff
@@ -0,0 +1,278 @@
...
+            import_schema_uri=_TEST_TEXT_ENTITY_IMPORT_SCHEMA,
+        )
+
+        data_items_post_import = dataset_gapic_client.list_data_items(
+            parent=my_dataset.resource_name
+        )
+
+        assert len(list(data_items_post_import)) > 400  # Expected value 469
+
+    @pytest.mark.usefixtures("delete_new_dataset")
```
sasha-gitg, line 200: Confirming this gets called if the test fails internally?

tests/integration/aiplatform/test_dataset.py
```diff
@@ -0,0 +1,278 @@
...
+            metadata_schema_uri=aiplatform.schema.dataset.metadata.image,
+            gcs_source=_TEST_IMAGE_OBJECT_DETECTION_GCS_SOURCE,
+            import_schema_uri=_TEST_IMAGE_OBJ_DET_IMPORT_SCHEMA,
+        )
+
+        data_items_iterator = dataset_gapic_client.list_data_items(
+            parent=img_dataset.resource_name
+        )
+
+        assert len(list(data_items_iterator)) > 10  # Expected value 14
```
sasha-gitg, line 218: Why not test for exactly 14?

tests/integration/aiplatform/test_dataset.py
```diff
@@ -0,0 +1,278 @@
...
+            metadata_schema_uri=aiplatform.schema.dataset.metadata.tabular,
+            gcs_source=[_TEST_TABULAR_CLASSIFICATION_GCS_SOURCE],
+        )
+
+        gapic_dataset = tabular_dataset._gca_resource
+
+        gapic_metadata = json_format.MessageToDict(gapic_dataset._pb.metadata)
+        gcs_source_uris = gapic_metadata["inputConfig"]["gcsSource"]["uri"]
+
+        assert _TEST_TABULAR_CLASSIFICATION_GCS_SOURCE in gcs_source_uris
```
sasha-gitg, line 240: Can this test be more specific and test that the resource has exactly this one uri?

sasha-gitg (approved):
LGTM!

tests/integration/aiplatform/test_dataset.py
vinnysenthil: Yes, I'll change it to do so

tests/integration/aiplatform/test_dataset.py
vinnysenthil: It does not, I'll look into capturing the `AssertionError` and allowing the teardown to still proceed.

tests/integration/aiplatform/test_dataset.py
vinnysenthil: For both data item checks, I wanted to leave a bit of leeway to prevent flakiness caused by partial failures of the data import service. But I can change these to be exact, since a partial failure would mean our test source data is incorrectly formatted.

tests/integration/aiplatform/test_dataset.py
vinnysenthil: Resolved this by setting `shared_state['dataset_name']` before making the assertions

busunkim96 (approved):
LGTM. Please rename to something like `test:` since this change shouldn't show up in the changelog. https://www.conventionalcommits.org/en/v1.0.0/

PR #128, opened 2020-12-10
feat: Added tabular forecasting samples

ivanmkc:
Added samples and tests for AutoML forecasting

- [x] list_model_evaluations_tabular_forecasting_sample
- [x] create_training_pipeline_tabular_forecasting_sample
- [x] get_model_evaluation_tabular_forecasting
- [ ] predict_tabular_forecasting <- Forecasting doesn't support predict so will do in a subsequent PR.


.sample_configs/param_handlers/create_training_pipeline_tabular_forecasting_sample.py
```diff
@@ -0,0 +1,90 @@
+# Copyright 2020 Google LLC
```
ivanmkc, line 1: @dizcology I had to manually paste this in. Is that expected?

.sample_configs/param_handlers/create_training_pipeline_tabular_forecasting_sample.py
```diff
@@ -0,0 +1,90 @@
...
+        {"auto": {"column_name": "county_fips_code"}},
+        {"auto": {"column_name": "confirmed_cases"}},
+        {"auto": {"column_name": "deaths"}}
+    ]
+
+    period = {"unit": "day", "quantity": 1}
+
+    training_task_inputs_dict = {
+        # required inputs
+        "targetColumn": target_column,
```
ivanmkc, line 46: Curious as to if we should show comments for each of these values in the sample.

samples/snippets/create_training_pipeline_tabular_forecasting_sample_test.py
```diff
@@ -0,0 +1,87 @@
...
+from google.cloud import aiplatform
+import pytest
+
+import cancel_training_pipeline_sample
+import create_training_pipeline_tabular_forecasting_sample
+import delete_training_pipeline_sample
+import helpers
+
+PROJECT_ID = os.getenv("BUILD_SPECIFIC_GCLOUD_PROJECT")
+DATASET_ID = "7172228697192136704"  # iris 1000
```
ivanmkc, line 27: need to fix comment

samples/snippets/get_model_evaluation_tabular_forecasting_sample.py
```diff
@@ -0,0 +1,37 @@
+# Copyright 2020 Google LLC
```
ivanmkc, line 1: Same as corresponding classification sample.

samples/snippets/predict_tabular_forecasting_sample.py
```diff
@@ -0,0 +1,44 @@
+# Copyright 2020 Google LLC
```
ivanmkc, line 1: Same as corresponding classification sample.

samples/snippets/predict_tabular_forecasting_sample_test.py
```diff
@@ -0,0 +1,38 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+
+
+import predict_tabular_classification_sample
+
+ENDPOINT_ID = "TODO"
```
ivanmkc, line 20: WIP

ivanmkc: Blocked by: https://b.corp.google.com/issues/175666946

ivanmkc:
Is the `Samples - Python 3.7 — Build failed` failure normal?

ivanmkc:
Fixing CI errors.

samples/snippets/get_model_evaluation_tabular_forecasting_sample.py
dizcology: I think we don’t really need to have sample variants for get and list methods. I see that we have them for some other kinds of AutoML services and perhaps those need to be removed (in a separate PR).

samples/snippets/predict_tabular_forecasting_sample.py
dizcology: Is it possible that all the predict tabular samples would be identical?  Is so we should consolidate those too perhaps. On the other hand if the sample tests would be very different (Eg you might be looking for different specific fields in the predictions), then let’s keep this.

.sample_configs/param_handlers/create_training_pipeline_tabular_forecasting_sample.py
aribray: @ivanmkc do you mean comments like # display_name: YOUR_DISPLAY_NAME?

.sample_configs/param_handlers/create_training_pipeline_tabular_forecasting_sample.py
```diff
@@ -0,0 +1,90 @@
...
+        "timeSeriesIdentifierColumn": time_series_identifier_column,
+        "timeColumn": time_column,
+        "transformations": transformations,
+        "period": period,
+
+        # Objective function the model is to be optimized towards.
```
dizcology, line 52: Maybe it's better to point to the doc pages instead of including the information below here.

.sample_configs/variants.yaml
```diff
@@ -175,7 +177,7 @@ list_hyperparameter_tuning_jobs:
 list_model_evaluation_slices:
 - ''
 list_model_evaluations:
-- ''
+- tabular_forecasting
```
dizcology, line 180: this would replace the list_model_evaluation sample.  Please keep the "default" variant keyed with the empty string.

dizcology: Also - perhaps we don't need a special `list_model_evaluations` variant unless the call to `list_model_evaluations` is different in the `tabular_forecasting` use case.

ivanmkc: Oh I see. Good catch.

.sample_configs/process_configs.yaml
```diff
@@ -182,6 +183,7 @@ get_model_evaluation_sample:
   - model_explanation
 get_model_evaluation_slice_sample: {}
 get_model_evaluation_tabular_classification_sample: {}
+get_model_evaluation_tabular_forecasting_sample: {}
```
dizcology, line 186: If the `tabular_forcasting` variant for `get_model_evaluation` ends up identical with any `get_model_evaluation`, please skip it.  (and we should separately clean up the others - but perhaps i missed something here)

samples/snippets/predict_tabular_forecasting_sample.py
ivanmkc: Predict cannot be used with 'tabular'. They only support `batch_predict`, I will think about consolidation when doing batch_predict next week.

.sample_configs/param_handlers/create_training_pipeline_tabular_forecasting_sample.py
ivanmkc: Thanks for reviewing, I was wondering if we should add comments for each param in the samples.

Seems like Yuhan is suggesting to just tell them to read the docstrings.

samples/snippets/get_model_evaluation_tabular_forecasting_sample.py
ivanmkc: Sure!

.sample_configs/param_handlers/create_training_pipeline_tabular_forecasting_sample.py
ivanmkc: Do you mean include a link in the comments?

.sample_configs/process_configs.yaml
ivanmkc: I got it, we want to consolidate if possible. Let me do another pass.

.sample_configs/param_handlers/create_training_pipeline_tabular_forecasting_sample.py
dizcology: right - if the information here is available on a particular documentation page.

ivanmkc:
@dizcology can I get another look?

dizcology (approved):

PR #129, opened 2020-12-11
feat: Added optional model args for custom training

ivanmkc:
Added optional args to `CustomTrainingJob.init`:

```
        model_serving_container_command: Optional[Sequence[str]] = None,
        model_serving_container_args: Optional[Sequence[str]] = None,
        model_serving_container_environment_variables: Optional[Dict[str, str]] = None,
        model_serving_container_ports: Optional[Sequence[int]] = None,
        model_description: Optional[str] = None,
        model_instance_schema_uri: Optional[str] = None,
        model_parameters_schema_uri: Optional[str] = None,
        model_prediction_schema_uri: Optional[str] = None,
```

Fixes https://b.corp.google.com/issues/172365796

## TODO
- [x] Added unit test

### DIFF
--- a/google/cloud/aiplatform/training_jobs.py
+++ b/google/cloud/aiplatform/training_jobs.py
@@ -38,6 +38,7 @@ from google.cloud.aiplatform import utils
 from google.cloud.aiplatform_v1beta1.services.pipeline_service import (
     client as pipeline_service_client,
 )
+from google.cloud.aiplatform_v1beta1.types import env_var
 from google.cloud.aiplatform_v1beta1.types import (
     accelerator_type as gca_accelerator_type,
 )
@@ -47,7 +48,6 @@ from google.cloud.aiplatform_v1beta1.types import pipeline_state as gca_pipeline
 from google.cloud.aiplatform_v1beta1.types import (
     training_pipeline as gca_training_pipeline,
 )
-from google.cloud.aiplatform_v1beta1.types import env_var
 
 from google.cloud import storage
 from google.protobuf import json_format
--- a/google/cloud/aiplatform/training_jobs.py
+++ b/google/cloud/aiplatform/training_jobs.py
@@ -923,7 +923,6 @@ class CustomTrainingJob(_TrainingJob):
     in Cloud AI Platform Training.
     """
 
-    # TODO(b/172365796) add remainder of model optional arguments
     def __init__(
         self,
         display_name: str,
@@ -933,6 +932,8 @@ class CustomTrainingJob(_TrainingJob):
         model_serving_container_image_uri: Optional[str] = None,
         model_serving_container_predict_route: Optional[str] = None,
         model_serving_container_health_route: Optional[str] = None,
+        model_etag: Optional[str] = None,
+        model_labels: Optional[Sequence[gca_model.model.Model.LabelsEntry]] = None,
         project: Optional[str] = None,
         location: Optional[str] = None,
         credentials: Optional[auth_credentials.Credentials] = None,
@@ -987,6 +988,20 @@ class CustomTrainingJob(_TrainingJob):
                 send health check requests to the container, and which must be supported
                 by it. If not specified a standard HTTP path will be used by AI
                 Platform.
+            model_etag (str):
+                Used to perform consistent read-modify-write
+                updates. If not set, a blind "overwrite" update
+                happens.
+            model_labels (Sequence[~.model.Model.LabelsEntry]):
+                The labels with user-defined metadata to
+                organize your Models.
+                Label keys and values can be no longer than 64
+                characters (Unicode codepoints), can only
+                contain lowercase letters, numeric characters,
+                underscores and dashes. International characters
+                are allowed.
+                See https://goo.gl/xmQnxf for more information
+                and examples of labels.                 
             project (str):
                 Project to run training in. Overrides project set in aiplatform.init.
             location (str):
@@ -1014,6 +1029,8 @@ class CustomTrainingJob(_TrainingJob):
         self._model_serving_container_health_route = (
             model_serving_container_health_route
         )
+        self._model_etag = model_etag
+        self._model_labels = model_labels
 
         self._script_path = script_path
         self._staging_bucket = (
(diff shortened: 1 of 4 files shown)


google/cloud/aiplatform/training_jobs.py
```diff
@@ -987,6 +995,80 @@ def __init__(
                 send health check requests to the container, and which must be supported
                 by it. If not specified a standard HTTP path will be used by AI
                 Platform.
+            model_serving_container_command: Optional[Sequence[str]]=None,
```
sasha-gitg, line 998: Formatting on the type signatures on these arguments.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -1015,6 +1097,20 @@ def __init__(
             model_serving_container_health_route
         )
 
+        self._model_serving_container_command = model_serving_container_command
```
sasha-gitg, line 1100: Can we take this opportunity to envelope all the Model arguments into an object. Perhaps we can create the instance of `gca_model.Model` here?

sasha-gitg (changes requested):
Looks good. Just one main item on encapsulating the model args.

google/cloud/aiplatform/training_jobs.py
ivanmkc: I'll look into it

google/cloud/aiplatform/training_jobs.py
ivanmkc: Sorry, what do you mean?

sasha-gitg: For example:
`model_serving_container_command: Optional[Sequence[str]]=None,`

should be:

`model_serving_container_command (Sequence[str]):`

google/cloud/aiplatform/training_jobs.py
```diff
@@ -47,6 +47,7 @@
 from google.cloud.aiplatform_v1beta1.types import (
     training_pipeline as gca_training_pipeline,
 )
+from google.cloud.aiplatform_v1beta1.types import env_var
```
sasha-gitg, line 50: nit: this should be in alphabetical order in this import block.

google/cloud/aiplatform/training_jobs.py
ivanmkc: The model needs `model_display_name`, which is passed in the `run` function and not the `init` function, so we can't create it here.

We COULD create the `gca_model.ModelContainerSpec` at `CustomTrainingJob.init` though. I'll go ahead and do that.

google/cloud/aiplatform/training_jobs.py
ivanmkc: Oh ok, I didn't see that it was the comments you were commenting on.

google/cloud/aiplatform/training_jobs.py
sasha-gitg: Can't you create the model up front and set the `display_name` afterwards? ie.

```
self._managed_model = gca_model.Model(...)
```

```
self._managed_model.display_name = display_name
```

ivanmkc: I didn't know you could do that with protos. Let me try it.

ivanmkc:
Still need to fix end-to-end tests

ivanmkc:
> Still need to fix end-to-end tests

Done

ivanmkc:
@sasha-gitg Ready for another look, thanks!

ivanmkc:
I think the Kokoro CI step is broken right?

It's blocking 3 of my PRs from merging.

sasha-gitg:
> I think the Kokoro CI step is broken right?
> 
> It's blocking 3 of my PRs from merging.

Update the PRs with the latest dev branch. That should fix the builds.

ivanmkc:
@sasha-gitg still need a stamp if ready.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -1015,6 +1097,62 @@ def __init__(
...
+            command=model_serving_container_command,
+            args=model_serving_container_args,
+            env=env,
+            ports=ports,
+            predict_route=model_serving_container_predict_route,
+            health_route=model_serving_container_health_route,
+        )
+
+        # create model payload
+        self._managed_model = gca_model.Model(
```
sasha-gitg, line 1150: Since this managed_model is created, I don't think we need to store the fields stored in the managed_model as attributes on this object.

ie:
self._model_serving_container_args 
self._model_serving_container_environment_variables
self._model_serving_container_ports
self._model_description
etc...
can be removed and any downstream instance method that may require these fields can access them through self._managed_model.

sasha-gitg (changes requested):

google/cloud/aiplatform/training_jobs.py
ivanmkc: Done

google/cloud/aiplatform/training_jobs.py
```diff
@@ -987,6 +995,80 @@ def __init__(
                 send health check requests to the container, and which must be supported
                 by it. If not specified a standard HTTP path will be used by AI
                 Platform.
+            model_serving_container_command (Sequence[str]),
```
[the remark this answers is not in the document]
ivanmkc: Done

ivanmkc:
@sasha-gitg test fixed

google/cloud/aiplatform/training_jobs.py
```diff
@@ -1148,26 +1270,16 @@ def run(
...
-                display_name=model_display_name, container_spec=container_spec
-            )
-
         # make and copy package
         python_packager = _TrainingScriptPythonPackager(
             script_path=self._script_path, requirements=self._requirements
         )
 
+        managed_model = self._managed_model
+        if model_display_name:
```
sasha-gitg, line 1279: `managed_model` should be None in the case where `model_display_name` is not passed in to support not exporting a model from the training job.

sasha-gitg: Can you create a bug to capture that as a unit test case?

sasha-gitg (approved):
Looks great! Just needs one more change to support not exporting a managed model.

PR #134, opened 2020-12-15
chore: alpha

dizcology:
This makes the mb-release package show up as version `0.4.0a1`, and allows the user to revert to the public package with `pip install google-cloud-aiplatform==0.3.1`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,7 +21,7 @@ import os
 import setuptools  # type: ignore
 
 name = "google-cloud-aiplatform"
-version = "0.3.1"
+version = "0.3.1-alpha"
 description = "Cloud AI Platform API client library"
 
 package_root = os.path.abspath(os.path.dirname(__file__))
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -21,7 +21,7 @@
 import setuptools  # type: ignore
 
 name = "google-cloud-aiplatform"
-version = "0.3.1"
+version = "0.3.1-alpha"
```
busunkim96, line 24: If this is going to be pushed out to PyPI, please use one of the schemes listed here: https://www.python.org/dev/peps/pep-0440/#examples-of-compliant-version-schemes

sasha-gitg (approved):

setup.py
sasha-gitg: We will not push this to PyPI.

I wonder if the postfix should be `experimental` to align with the current release stage naming.

dizcology: Updated to `0.4.0a1`.

Issue #147, opened 2020-12-22
predict_image_object_detection_sample.py uses wrong parameter format


Crash-GHaun:
- OS type and version:  Debian GNU/ Linux 10 
  - Python version: 3.7
  - pip version: 20.2.4
  - `google-cloud-aiplatform` version: 0.3.1

#### Steps to reproduce

Line 45 in predict_image_object_detection_sample.py 

Parameters have no effect because of "_". Need to change to camalCase 

#### Code example

If available to you: https://github.com/googleapis/python-aiplatform/compare/master...Crash-GHaun:patch-1?expand=1

Following process to create bug before pull request

dizcology:
@Crash-GHaun Thank you for reporting the issue!

Crash-GHaun:
Anytime! Quick question on this, will this update propagate to the [documentation on the website](https://cloud.google.com/ai-platform-unified/docs/predictions/online-predictions-automl#using_the_api)? 

If not I'll make sure it get's changed there too :)

dizcology:
I *think* the documentation should get updated automatically after a few minutes, I will keep an eye on it.

PR #150, opened 2020-12-22
fix: predict image samples params

dizcology:
Fixes https://github.com/googleapis/python-aiplatform/issues/147.


lucaswadedavis (approved):
Golly this breaks my expectations. Thanks for the fix Yu-Han.

PR #152, opened 2020-12-24
feat: Add AutoML Image Training Job class

vinnysenthil:
### Description of changes:
- Adds `AutoMLImageTrainingJob` class, subclass of `_TrainingJob`
  - One public method `run()` 
- 11 unit tests
- Adds constants and schemas specific to AutoML Image training jobs
- Fixes a few typos in `training_jobs.py`

Fixes [b/175726626](http://b/175726626) 🦕


google/cloud/aiplatform/training_jobs.py
```diff
@@ -1753,3 +1754,329 @@ def _model_upload_fail_string(self) -> str:
...
+            model_display_name = self._display_name
+
+        model_tbt = None  # gca Model to be trained
+
+        if not self._base_model:
+            model_tbt = gca_model.Model(display_name=model_display_name)
+        else:
+            # Use provided base_model to pass to model_to_upload causing the
+            # description and tags from base_model to be passed onto the new model
+            model_tbt = self._base_model._gca_resource
```
sasha-gitg, line 2060: This seems like it should be copied or this will mutate the display name in the Model instance's gca_resource.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -1753,3 +1754,329 @@ def _model_upload_fail_string(self) -> str:
...
+        if model_display_name is None:
+            model_display_name = self._display_name
+
+        model_tbt = None  # gca Model to be trained
+
+        if not self._base_model:
+            model_tbt = gca_model.Model(display_name=model_display_name)
+        else:
+            # Use provided base_model to pass to model_to_upload causing the
+            # description and tags from base_model to be passed onto the new model
```
sasha-gitg, line 2059: I really like this but should we copy all the fields? Are there any issues with setting the Output only fields?

google/cloud/aiplatform/training_jobs.py
```diff
@@ -1753,3 +1754,329 @@ def _model_upload_fail_string(self) -> str:
...
+                especially may happen when further model training ceases to provide
+                any improvements.
+                If the budget is set to a value known to be insufficient to train a
+                Model for the given training set, the training won't be attempted and
+                will error.
+                The minimum value is 1000 and the maximum is 72000.
+            model_display_name (str):
+                Optional. The display name of the managed AI Platform Model. The name
+                can be up to 128 characters long and can be consist of any UTF-8
+                characters. If a `base_model` was provided, the display_name in the
```
sasha-gitg, line 1925: My assumption is that when a base model is provided the service doesn't alter that base model resource and creates a new model resource with this `model_display_name` as the output model of the training job. Is that not the case? If it is the case then I don't think we need to include the "display_name is overwritten" statement.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -1753,3 +1754,329 @@ def _model_upload_fail_string(self) -> str:
...
+            training_fraction_split=training_fraction_split,
+            validation_fraction_split=validation_fraction_split,
+            test_fraction_split=test_fraction_split,
+            budget_milli_node_hours=budget_milli_node_hours,
+            model_display_name=model_display_name,
+            disable_early_stopping=disable_early_stopping,
+            sync=sync,
+        )
+
+    @base.optional_sync()
```
sasha-gitg, line 1963: We should force a dependency on `base_model` here to satisfy async workflows. Similar to how this method accepts a `dataset.Dataset` we should also pass in `base_model` here so it gets picked up by `base.optional_sync`.

tests/unit/aiplatform/test_automl_image_training_jobs.py
```diff
@@ -0,0 +1,388 @@
...
+    def test_init_all_parameters(self, mock_dataset_image):
+        """Ensure all private members are set correctly at initalization"""
+
+        aiplatform.init(project=_TEST_PROJECT)
+
+        job = training_jobs.AutoMLImageTrainingJob(
+            display_name=_TEST_DISPLAY_NAME,
+            prediction_type=_TEST_PREDICTION_TYPE_ICN,
+            model_type=_TEST_MODEL_TYPE_MOBILE,
+            base_model=mock_dataset_image,
```
sasha-gitg, line 168: Should this be `mock_model_image`?

tests/unit/aiplatform/test_automl_image_training_jobs.py
```diff
@@ -0,0 +1,388 @@
...
+                display_name=_TEST_DISPLAY_NAME,
+                prediction_type="classification",
+                model_type=_TEST_MODEL_TYPE_IOD,
+            )
+
+        with pytest.raises(ValueError, match=r"`base_model` is only supported"):
+            training_jobs.AutoMLImageTrainingJob(
+                display_name=_TEST_DISPLAY_NAME,
+                prediction_type=_TEST_PREDICTION_TYPE_IOD,
+                base_model=mock_dataset_image,
```
sasha-gitg, line 199: `mock_model_image` here too

sasha-gitg (changes requested):

google/cloud/aiplatform/training_jobs.py
```diff
@@ -1602,10 +1603,10 @@ def run(
         """
 
         if self._is_waiting_to_run():
-            raise RuntimeError("Custom Training is already scheduled to run.")
+            raise RuntimeError("AutoML Tabular Training is already scheduled to run.")
```
dizcology, line 1606: Out of scope for this PR, but in cases like this it might be helpful if the error message includes some additional information such as the training job's full resource name.

google/cloud/aiplatform/constants.py
```diff
@@ -29,3 +29,20 @@
...
+    "MOBILE_TF_VERSATILE_1",
+    "MOBILE_TF_HIGH_ACCURACY_1",
+}
+
+# Defined by gs://google-cloud-aiplatform/schema/trainingjob/definition/automl_image_*
+# Format: "prediction_type": set() of model_type's
+#
+# NOTE: When adding a new prediction_type's, ensure it fits the pattern
+#       "automl_image_{prediction_type}_*" used by the YAML schemas on GCS
+AUTOML_IMAGE_PREDICTION_MODEL_TYPES = {
```
dizcology, line 44: For later: use the same information in schema types: https://github.com/googleapis/python-aiplatform/blob/master/google/cloud/aiplatform/v1beta1/schema/trainingjob/definition_v1beta1/types/automl_image_classification.py#L96

google/cloud/aiplatform/training_jobs.py
```diff
@@ -1753,3 +1754,329 @@ def _model_upload_fail_string(self) -> str:
...
+                        This type is available only to columns that contain
+                        semantically numeric values, i.e. integers or floating
+                        point number, even if stored as e.g. strings.
+            multi_label: bool = False
+                Required. Default is False.
+                If false, a single-label (multi-class) Model will be trained
+                (i.e. assuming that for each image just up to one annotation may be
+                applicable). If true, a multi-label Model will be trained (i.e.
+                assuming that for each image multiple annotations may be applicable).
+            model_type: str = "CLOUD"
```
dizcology, line 1790: but the acceptable model_types depends on prediction_type, and this is validated in the function body below.

would it be cleaner if the public class is specialized, such as `AutoMLImageClassificationTrainingJob`, etc.?

google/cloud/aiplatform/training_jobs.py
vinnysenthil: I agree, while the service does ignore other fields we should only copy what we need. Updated the code to just copy the `description` and `labels`

google/cloud/aiplatform/training_jobs.py
vinnysenthil: Agreed it would be cleaner, but after checking with @sasha-gitg the plan is to get this class out then released the specialized down the road.

google/cloud/aiplatform/training_jobs.py
vinnysenthil: Captured in [b/177077413](http://b/177077413)

google/cloud/aiplatform/training_jobs.py
vinnysenthil: Good catch, added `base_model` as a param to `_run()`which is passed by `run()`

google/cloud/aiplatform/constants.py
vinnysenthil: Captured in [b/177079208](http://b/177079208)

dizcology (approved):

sasha-gitg (approved):

PR #160, opened 2021-01-08
fix: suppress no project id warning

morgandu:
Fixes https://b.corp.google.com/issues/176166878

### DIFF
--- a/google/cloud/aiplatform/initializer.py
+++ b/google/cloud/aiplatform/initializer.py
@@ -65,6 +65,8 @@ class _Config:
         """
         if project:
             self._project = project
+            if not os.environ.get(google.auth.environment_vars.PROJECT):
+                os.environ[google.auth.environment_vars.PROJECT] = self._project
         if location:
             utils.validate_region(location)
             self._location = location
--- a/google/cloud/aiplatform/initializer.py
+++ b/google/cloud/aiplatform/initializer.py
@@ -75,12 +75,6 @@ class _Config:
             self._staging_bucket = staging_bucket
         if credentials:
             self._credentials = credentials
-        else:
-            logger = logging.getLogger("google.auth._default")
-            logging_warning_filter = utils.LoggingWarningFilter()
-            logger.addFilter(logging_warning_filter)
-            self._credentials, _ = google.auth.default()
-            logger.removeFilter(logging_warning_filter)
 
     @property
     def project(self) -> str:
@@ -123,6 +117,12 @@ class _Config:
     @property
     def credentials(self) -> Optional[auth_credentials.Credentials]:
         """Default credentials, if provided."""
+        if not self._credentials:
+            logger = logging.getLogger("google.auth._default")
+            logging_warning_filter = utils.LoggingWarningFilter()
+            logger.addFilter(logging_warning_filter)
+            self._credentials, _ = google.auth.default()
+            logger.removeFilter(logging_warning_filter)
         return self._credentials
 
     def get_client_options(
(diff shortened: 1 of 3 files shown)


google/cloud/aiplatform/initializer.py
```diff
@@ -65,6 +65,8 @@ def init(
         """
         if project:
             self._project = project
+            if not os.environ.get(google.auth.environment_vars.PROJECT):
+                os.environ[google.auth.environment_vars.PROJECT] = self._project
```
sasha-gitg, line 69: This seems like it would cause unintended side effects as other SDKs may reference this environment variable. I think we need to:

1. Determine the source of the warning and the call site in the SDK that leads to warning being logged.
2. See if we can avoid calling that method when we have project or pass project to the method.

It looks like this is the source of the warning:
https://github.com/googleapis/google-auth-library-python/blob/master/google/auth/_default.py#L251

My first guess is that it's called here: https://github.com/googleapis/python-aiplatform/blob/be78cdeaadcf0bcfe7267cb03001211f4714964b/google/cloud/aiplatform/initializer.py#L95

But I suspect that assumption is incorrect because, if the project set, it shouldn't invoke that method. So the issue may lie elsewhere.

sasha-gitg (changes requested):

google/cloud/aiplatform/initializer.py
busunkim96: Yep, `GOOGLE_CLOUD_PROJECT` is one of the standardized environment variables respected across libraries so overriding this may have unintended side effects. [docs](https://g3doc.corp.google.com/company/teams/cloud-client-libraries/dev/reference/client-library-environment-variables.md?cl=head#google-cloud-project)

FWIW the existing behavior makes sense to me - the warning only seems to be raised if the user has both (1) not passed a `project` to `__init__` and (2) not set a project in the environment. https://github.com/googleapis/python-aiplatform/blob/4459fffe3711c92db085b5049e4076b438ac0ad6/google/cloud/aiplatform/initializer.py#L44-L67

https://github.com/googleapis/python-aiplatform/blob/4459fffe3711c92db085b5049e4076b438ac0ad6/google/cloud/aiplatform/initializer.py#L79-L100

morgandu: @sasha-gitg , no, I thought at the beginning the warning was from https://github.com/googleapis/python-aiplatform/blob/be78cdeaadcf0bcfe7267cb03001211f4714964b/google/cloud/aiplatform/initializer.py#L95 too. But using Dataset as example, the warning comes from https://github.com/googleapis/python-aiplatform/blob/be78cdeaadcf0bcfe7267cb03001211f4714964b/google/cloud/aiplatform_v1beta1/services/dataset_service/transports/grpc.py#L130.

morgandu: modified approach: when no credentials is provided during MBSDK initialization, set the credentials initializer.global_config._credentials to google.auth.default's credentials, thus during any service client instantiation, google.auth.default will not be called again.

google/cloud/aiplatform/initializer.py
```diff
@@ -75,6 +75,12 @@ def init(
             self._staging_bucket = staging_bucket
         if credentials:
             self._credentials = credentials
+        else:
+            logger = logging.getLogger("google.auth._default")
```
sasha-gitg, line 79: Please move this logic into the `credentials` property to be consistent with the default project retrieval. Default `project` is pulled at time of accessing the property while, in this case, default `credential` is pulled when initializing the config.

sasha-gitg (approved):
LGTM! Thanks for putting this together. One minor change for consistency.

PR #161, opened 2021-01-08
feat: Added CustomTrainingJob.get

ivanmkc:
This allows users to retrieve an existing CustomTrainingJob using the resource name.
The `get` function will always block as it is a short function.
The `get_model` is now async, so downstream functions can use the future that was created at this stage.

Fixes https://b.corp.google.com/issues/174771897

### TODO
- [x] Add validation that the job is an instance of CustomTrainingJob
- [x] Add get function to abstract class
- [x] Add unit test for incorrect training task definition
- [x] Add doc strings 
- [x] Fix missing future 
- [x] Test TrainingJob.get_model for a variety of use cases

### Test cases

I've tested this against the below potential use cases.

#### Retrieving the state for retrieved job

```
job = aiplatform.CustomTrainingJob(...)
model = job.run(ds, replica_count=1, model_display_name="abalone-model")
 
Session stops...
 
job = aiplatform.CustomTrainingJob.get(
   resource_name="resource_name"
)
 
job.state # Should be pending/success/failure
```
 
#### Running a job on a retrieved job should raise an error

```
job = aiplatform.CustomTrainingJob(...)
model = job.run(ds, replica_count=1, model_display_name="abalone-model")
 
Session stops...
 
job = aiplatform.CustomTrainingJob.get(
   resource_name="resource_name"
)
 
job.state # Should be pending/success/failure
```

# Async case 1
```
new_job = CustomTrainingJob.get(CURRENTLY_RUNNING_JOB)
new_job.get_model(sync=True) # This should block until training finishes
```

# Async case 2
```
new_job = CustomTrainingJob.get(CURRENTLY_RUNNING_JOB)
new_job.get_model(sync=False)
new_job.wait() # This should block until training finishes
```

# Async case 3 (downstream usage)
```
new_job = CustomTrainingJob.get(CURRENTLY_RUNNING_JOB)
model = new_job.get_model(sync=False)
endpoint = model.deploy(sync=False)
endpoint.wait() # This should block until training finishes
```

### DIFF
--- a/google/cloud/aiplatform/training_jobs.py
+++ b/google/cloud/aiplatform/training_jobs.py
@@ -454,8 +454,15 @@ class _TrainingJob(base.AiPlatformResourceNounWithFutureManager):
 
         return self._gca_resource.state
 
+    def get_model(self, sync=True) -> Optional[models.Model]:
+        self._assert_has_run()
+        if not self._gca_resource.model_to_upload:
+            raise RuntimeError(self._model_upload_fail_string)
+
+        return self._force_get_model(sync=sync)
+
     @base.optional_sync()
-    def get_model(self, sync: bool = True,) -> Optional[models.Model]:
+    def _force_get_model(self, sync: bool = True) -> Optional[models.Model]:
         """AI Platform Model produced by this training, if one was produced.
 
         Args:
@@ -468,12 +475,12 @@ class _TrainingJob(base.AiPlatformResourceNounWithFutureManager):
             model: AI Platform Model produced by this training or None if a model was
                 not produced by this training.
         """
-        self._assert_has_run()
+        model = self._get_model()
 
-        if not self._gca_resource.model_to_upload:
+        if model is None:
             raise RuntimeError(self._model_upload_fail_string)
 
-        return self._get_model()
+        return model
 
     def _get_model(self) -> Optional[models.Model]:
         """Helper method to get and instantiate the Model to Upload.
--- a/google/cloud/aiplatform/training_jobs.py
+++ b/google/cloud/aiplatform/training_jobs.py
@@ -454,6 +454,26 @@ class _TrainingJob(base.AiPlatformResourceNounWithFutureManager):
 
         return self._gca_resource.state
 
+    @classmethod
+    @abc.abstractmethod
+    def get(
+        cls,
+        resource_name: str,
+        project: Optional[str] = None,
+        location: Optional[str] = None,
+        credentials: Optional[auth_credentials.Credentials] = None,
+    ):
+        """Get training job for the given resource_name.
+
+        Raises:
+            ValueError: If the retrieved training job's training task definition 
+                doesn't match the training job class type.
+
+        Returns:
+            An AI Platform Training Job
+        """
+        pass
+
     def get_model(self) -> Optional[models.Model]:
         """AI Platform Model produced by this training, if one was produced.
 
@@ -1979,11 +1999,21 @@ class CustomContainerTrainingJob(_CustomTrainingJob):
     @classmethod
     def get(
         cls,
-        training_job_name: str,
+        resource_name: str,
         project: Optional[str] = None,
         location: Optional[str] = None,
         credentials: Optional[auth_credentials.Credentials] = None,
     ):
+        """Get CustomTrainingJob for the given resource_name.
+
+        Raises:
+            ValueError: If the retrieved training job's training task definition 
+                doesn't match the custom training task definition.
+
+        Returns:
+            An AI Platform Training Job
+        """
+
         # Create job with dummy parameters
         # These parameters won't be used as user can not run the job again.
         # If they try, an exception will be raised.
@@ -1997,7 +2027,7 @@ class CustomContainerTrainingJob(_CustomTrainingJob):
             staging_bucket="dummy",
         )
 
-        job._gca_resource = job._get_gca_resource(resource_name=training_job_name)
+        job._gca_resource = job._get_gca_resource(resource_name=resource_name)
 
         if (
             job._gca_resource.training_task_definition
@@ -2005,8 +2035,8 @@ class CustomContainerTrainingJob(_CustomTrainingJob):
         ):
             raise ValueError(
                 f"The retrieved job's training task definition "
-                "is {job._gca_resource.training_task_definition}, "
-                "which is not compatible with CustomTrainingJob."
+                f"is {job._gca_resource.training_task_definition}, "
+                f"which is not compatible with CustomTrainingJob."
             )
 
         return job
--- a/google/cloud/aiplatform/training_jobs.py
+++ b/google/cloud/aiplatform/training_jobs.py
@@ -454,9 +454,16 @@ class _TrainingJob(base.AiPlatformResourceNounWithFutureManager):
 
         return self._gca_resource.state
 
-    def get_model(self) -> Optional[models.Model]:
+    @base.optional_sync()
+    def get_model(self, sync: bool = True,) -> Optional[models.Model]:
         """AI Platform Model produced by this training, if one was produced.
 
+        Args:
+            sync (bool):
+                Whether to execute this method synchronously. If False, this method
+                will be executed in concurrent Future and any downstream object will
+                be immediately returned and synced when the Future has completed.
+                
         Returns:
             model: AI Platform Model produced by this training or None if a model was
                 not produced by this training.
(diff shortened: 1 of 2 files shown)


google/cloud/aiplatform/training_jobs.py
```diff
@@ -1150,6 +1150,31 @@ def __init__(
...
             )
 
+    @classmethod
+    def get(
+        cls,
+        training_job_name: str,
+        project: Optional[str] = None,
+        location: Optional[str] = None,
+        credentials: Optional[auth_credentials.Credentials] = None,
+    ):
```
ivanmkc, line 1160: I couldn't add a type annotation as discussed in https://github.com/python/typing/issues/254

I could add it after adding `from __future__ import annotations` but this broke something else.

ivanmkc: Let me know what you want to use this workaround (from the above link):

```
_T = TypeVar("_T")

class Foo:
    @classmethod
    def create(cls: Type[_T]) -> _T:
        return cls()
```

tests/unit/aiplatform/test_training_jobs.py
```diff
@@ -1069,6 +1090,18 @@ def test_run_call_pipeline_service_create_distributed_training(
...
+        aiplatform.init(project=_TEST_PROJECT)
+        training_jobs.CustomTrainingJob.get(training_job_name=_TEST_NAME)
+        get_training_job_mock.assert_called_once_with(name=_TEST_NAME)
+
+    @pytest.mark.usefixtures("get_training_job_mock")
+    def test_get_training_job_with_id_only(self, get_training_job_mock):
+        aiplatform.init(project=_TEST_PROJECT, location=_TEST_LOCATION)
+        training_jobs.CustomTrainingJob.get(training_job_name=_TEST_ID)
+        get_training_job_mock.assert_called_once_with(name=_TEST_NAME)
+
```
ivanmkc, line 1104: Didn't add the other units tests from `Dataset.init` because they are already covered by those tests.

google/cloud/aiplatform/training_jobs.py
ivanmkc: Ah I just went with ` -> "CustomTrainingJob"`

ivanmkc:
Actually, I'll remove the abstract method since each subclass will have a different type signature.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -358,9 +358,16 @@ def state(self) -> Optional[gca_pipeline_state.PipelineState]:
 
         return self._gca_resource.state
 
-    def get_model(self) -> Optional[models.Model]:
+    @base.optional_sync()
```
ivanmkc, line 361: Is it as simple as adding this annotation?

@sasha-gitg

ivanmkc:
@sasha-gitg I tested all the listed cases and it seems to work as expected.

vinnysenthil:
@ivanmkc the lint failed, run `nox -s lint` in your branch to see what you need to fix

google/cloud/aiplatform/training_jobs.py
sasha-gitg: The flow should look something like this:

```python

def get_model(self, sync=True) -> Optional[models.Model]:
  self._assert_has_run()
  if not self._gca_resource.model_to_upload:
    raise RuntimeError(self._model_upload_fail_string)

  return self._force_get_model(sync=sync)

@base.optional_sync()
def _force_get_model(self, sync=True) -> models.Model:
  model = self._get_model()
  if model is None:
    raise RuntimeError(self._model_upload_fail_string)
  return model
```

This will deal with the scenario where the user calls job.get_model(sync=False) on a job that does not produce a Model.

You can add these changes and preferably a test or to unblock this PR, this can be removed for now and can you open a ticket to add aync support to get_model and add as a TODO (please add this comment to the ticket for posterity if you do).

sasha-gitg (changes requested):

tests/unit/aiplatform/test_training_jobs.py
```diff
@@ -1251,6 +1302,36 @@ def test_run_call_pipeline_service_create_distributed_training(
 
         assert job.state == gca_pipeline_state.PipelineState.PIPELINE_STATE_SUCCEEDED
 
+    @pytest.mark.usefixtures("get_training_job_custom_mock")
+    def test_get_training_job(self, get_training_job_custom_mock):
+        aiplatform.init(project=_TEST_PROJECT)
+        training_jobs.CustomTrainingJob.get(resource_name=_TEST_NAME)
+        get_training_job_custom_mock.assert_called_once_with(name=_TEST_NAME)
+
+    @pytest.mark.usefixtures("get_training_job_custom_mock_no_model_to_upload")
```
ivanmkc, line 1311: Added this test to test a no model case

ivanmkc:
@sasha-gitg Why should this throw a Runtime error instead of the current behaviour of returning `None` (see `_get_model`)?

sasha-gitg:
> @sasha-gitg Why should this throw a Runtime error instead of the current behaviour of returning `None` (see `_get_model`)?

In the `sync=False` case we would end up in a scenario where we would return a speculative Model resource noun object but that Model resource noun would never materialize.

Take this scenario where the TrainingJob is still running but it is not configured to produce a Model:
```Python
my_job = TrainingJob.get(...)  # returns training job that is still running
my_model = TrainingJob.get_model(..., sync=False)  # returns speculative Model resource
endpoint = my_model.deploy(...) # this will block until training is complete 
``` 

In the case where we throw the Exception, that last blocking command will raise an Exception describing why the Model is unavailable.

In the case where we pass back `None`, a cryptic exception about `None` not having a name attribute will be raised.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -1201,6 +1232,43 @@ def __init__(
...
+    @classmethod
+    def get(
+        cls,
+        resource_name: str,
+        project: Optional[str] = None,
+        location: Optional[str] = None,
+        credentials: Optional[auth_credentials.Credentials] = None,
+    ) -> "CustomTrainingJob":
+        """Get CustomTrainingJob for the given resource_name.
+
```
sasha-gitg, line 1244: Needs args section.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -1201,6 +1232,43 @@ def __init__(
...
+                doesn't match the custom training task definition.
+
+        Returns:
+            An AI Platform Training Job
+        """
+
+        # Create job with dummy parameters
+        # These parameters won't be used as user can not run the job again.
+        # If they try, an exception will be raised.
+        self = cls._empty_constructor()
```
sasha-gitg, line 1256: `project`, `location`, and `credentials` are retrieved from the global config but this can cause a failure if there's a mismatch with the passed in `resource name`

It looks like `resource_name` can be a resource name or a resource id.

If it's a resource name then the project and location should be extracted and validated against the project and location passed in if not None.

If it's a resource id then pass in project, location, credentials.

sasha-gitg (changes requested):

google/cloud/aiplatform/training_jobs.py
```diff
@@ -1201,6 +1232,43 @@ def __init__(
                 "set using aiplatform.init(staging_bucket='gs://my-bucket')"
             )
 
+    @classmethod
+    def get(
```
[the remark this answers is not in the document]
ivanmkc: I thought about that but given the task description, it looked like we wanted CustomTrainingJob to have this first.

I can move it to the superclass.

google/cloud/aiplatform/training_jobs.py
ivanmkc: Got it.

google/cloud/aiplatform/training_jobs.py
ivanmkc: Added a test

google/cloud/aiplatform/training_jobs.py
ivanmkc: Added

google/cloud/aiplatform/training_jobs.py
ivanmkc: @sasha-gitg I see the potential for conflict with the passed in project vs the project in the `resource_name`. 

However, isn't that  issue in `Dataset.__init__` and `Model.__init__` as well?

sasha-gitg: That's a good point. For now, we can keep this consistent with Dataset/Model/Endpoint and pass in project, location, credentials as passed in to the method. Can you please open a ticket to extract and initialize the client with the extracted project and location when receiving a fully qualified resource name in all cases in the SDK?

sasha-gitg (approved):
LGTM! Thanks!

PR #164, opened 2021-01-08
feat: Add custom container support

sasha-gitg:
Fixes [b/175147058](b/175147058) 🦕


tests/unit/aiplatform/test_training_jobs.py
```diff
@@ -1070,6 +1071,611 @@ def test_run_call_pipeline_service_create_distributed_training(
         assert job.state == gca_pipeline_state.PipelineState.PIPELINE_STATE_SUCCEEDED
 
 
+class TestCustomContainerTrainingJob:
+    def setup_method(self):
+        importlib.reload(initializer)
+        importlib.reload(aiplatform)
+
+    def teardown_method(self):
+        initializer.global_pool.shutdown(wait=True)
```
vinnysenthil, line 1080: Should we include this teardown in future tests?

vinnysenthil (approved):
LGTM!

tests/unit/aiplatform/test_training_jobs.py
sasha-gitg: Yes it should be included if the tests include `sync=False` and used the global_pool.

PR #165, opened 2021-01-09
feat: Add delete methods, add list_models and undeploy_all for Endpoint class

vinnysenthil:
### Summary of changes:
- Add a delete method to the `Dataset`, `Model`, and `Endpoint` classes
- Add a `list_models` and `undeploy_all` methods to the `Endpoint` class
- Six unit tests
- Changed the `Endpoint._undeploy()` to handle cases where a `deployed_model_id` is not in the `traffic_split` map returned by the service but is still deployed to the Endpoint

Fixes [b/175781695](http://b/175781695), [b/175782496](http://b/175782496), [b/175783505](http://b/175783505) 🦕


vinnysenthil:
Facing 3 hour timeout for Kokoro build again, adding `initializer.global_pool.shutdown(wait=True)` to two more test class teardowns as a potential fix.

google/cloud/aiplatform/models.py
```diff
@@ -1535,3 +1572,12 @@ def batch_predict(
         return jobs.BatchPredictionJob(
             batch_prediction_job_name=new_batch_prediction_job_name
         )
+
+    def delete_model(self):
```
sasha-gitg, line 1576: `delete` and optional `sync` here as well.

google/cloud/aiplatform/models.py
```diff
@@ -852,7 +854,42 @@ def predict(self, instances: List, parameters: Optional[Dict] = None) -> Predict
...
+        """Returns a list of the models deployed to this Endpoint.
+
+        Returns:
+            deployed_models (Sequence[aiplatform.gapic.DeployedModel]):
+                A list of the models deployed in this Endpoint.
+        """
+        self._sync_gca_resource()
+        return self._gca_resource.deployed_models
+
+    def undeploy_all(self):
```
sasha-gitg, line 869: Calling the `_undeploy` method skips unnecessary argument validation for this case.

Additionally, it looks like this can be made optionally sync, by adding a sync parameter and passing it to `_undeploy`. No need to decorate the method. The undeploys will queue on the endpoint object.

google/cloud/aiplatform/models.py
```diff
@@ -852,7 +854,42 @@ def predict(self, instances: List, parameters: Optional[Dict] = None) -> Predict
...
+
+        for deployed_model in self._gca_resource.deployed_models:
+            self.undeploy(deployed_model_id=deployed_model.id)
+
+    def delete_endpoint(self, force: bool = False):
+        """Deletes this AI Platform Endpoint resource. If force is set to True,
+        all models on this Endpoint will be undeployed prior to deletion.
+
+        Args:
+            force: bool = False
```
sasha-gitg, line 881: I think the format should be `force (bool):`

google/cloud/aiplatform/models.py
```diff
@@ -852,7 +854,42 @@ def predict(self, instances: List, parameters: Optional[Dict] = None) -> Predict
...
+        return self._gca_resource.deployed_models
+
+    def undeploy_all(self):
+        """Undeploys every model deployed to this Endpoint."""
+        self._sync_gca_resource()
+
+        for deployed_model in self._gca_resource.deployed_models:
+            self.undeploy(deployed_model_id=deployed_model.id)
+
+    def delete_endpoint(self, force: bool = False):
```
sasha-gitg, line 876: To make this optional `sync` we can add a sync argument and forward it to `undeploy_all`. The delete endpoint call should be a private method `_delete` with `@base.optional_sync` and this method should call that private method and forward the `sync` flag.

sasha-gitg (changes requested):
Added comments.

sasha-gitg (approved):
LGTM!

dizcology (approved):

PR #167, opened 2021-01-12
fix: default credentials config related test failures

morgandu:
fix and follow up on https://github.com/googleapis/python-aiplatform/pull/160

### DIFF
--- a/tests/unit/aiplatform/test_end_to_end.py
+++ b/tests/unit/aiplatform/test_end_to_end.py
@@ -186,7 +186,7 @@ class TestEndToEnd:
         mock_python_package_to_gcs.assert_called_once_with(
             gcs_staging_dir=test_training_jobs._TEST_BUCKET_NAME,
             project=test_training_jobs._TEST_PROJECT,
-            credentials=initializer.global_config.credentials,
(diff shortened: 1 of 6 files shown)


tests/unit/aiplatform/test_end_to_end.py
```diff
@@ -370,7 +370,7 @@ def test_dataset_create_to_model_predict_with_pipeline_fail(
         mock_python_package_to_gcs.assert_called_once_with(
             gcs_staging_dir=test_training_jobs._TEST_BUCKET_NAME,
             project=test_training_jobs._TEST_PROJECT,
-            credentials=initializer.global_config.credentials,
+            credentials=None,
```
sasha-gitg, line 373: Training Job passes self.credentials to this method: https://github.com/googleapis/python-aiplatform/blob/dev/google/cloud/aiplatform/training_jobs.py#L1597

So it seems like this should be the credentials and not None.

sasha-gitg (approved):
LGTM. One question about the tests. Thanks!

vinnysenthil (approved):
Rubber stamping for yoshi-python

PR #169, opened 2021-01-12
feat: add async support to batch predict

morgandu:
https://b.corp.google.com/issues/176178652

### DIFF
--- a/google/cloud/aiplatform/jobs.py
+++ b/google/cloud/aiplatform/jobs.py
@@ -121,11 +121,16 @@ class _Job(base.AiPlatformResourceNounWithFutureManager):
         """Helper method to compose the dashboard uri where job can be viewed."""
         if self._job_type:
             fields = utils.extract_fields_from_resource_name(self.resource_name)
-            url = f"https://pantheon.corp.google.com/ai/platform/locations/{fields.location}/{self._job_type}/{fields.id}?project={fields.project}"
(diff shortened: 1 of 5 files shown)


google/cloud/aiplatform/jobs.py
```diff
@@ -127,6 +185,279 @@ def __init__(
...
+
+        Raises:
+            ValueError:
+                If no or multiple source or destinations are provided. Also, if
+                provided instances_format or predictions_format are not supported
+                by AI Platform.
+
+        """
+
+        utils.validate_display_name(job_display_name)
```
sasha-gitg, line 333: Argument validation should be decoupled and run before launching the async method so we fail early. Something like:

```python
def create(cls, ..., sync=True):
  # validate arguments
  return self._create(..., sync=sync)

@base.optional_sync()
def _create(cls, ..., sync=True):
  # call service and wait
  return batch_prediction_job
```

sasha-gitg (changes requested):

tests/unit/aiplatform/test_jobs.py
```diff
@@ -205,7 +205,7 @@ def test_batch_prediction_job_status(self, get_batch_prediction_job_bq_output_mo
         )
```
sasha-gitg, line 205: There should be additional unit tests in this class to test `BatchPredictionJob.create` alone because the other tests are calling through `Model.batch_predict`.

sasha-gitg (approved):
LGTM. Added two more comments. Thanks!

vinnysenthil (approved):

PR #199, opened 2021-01-30
Fix: custom training jobs using managed non tabular dataset requiring annotation_schema_uri 

morgandu:
Fixes https://b.corp.google.com/issues/178834223


google/cloud/aiplatform/training_jobs.py
```diff
@@ -1496,6 +1557,30 @@ def run(
                 Additionally the dataset format is passed in as:
 
                 os.environ["AIP_DATA_FORMAT"]
+            annotation_schema_uri (str):
+                Only applicable to custom training.
```
sasha-gitg, line 1561: This line can be removed because this argument should only be made available in the Custom Training classes.

sasha-gitg (approved):
LGTM!

vinnysenthil (approved):

PR #210, opened 2021-02-04
feat: add filter and timestamp splits for custom training

sirtorry:
Fixes b/172365904 and b/172368070 respectively.


google/cloud/aiplatform/training_jobs.py
```diff
@@ -217,6 +251,13 @@ def _create_input_data_config(
                 test_fraction=test_fraction_split,
             )
 
+            # Create filter split spec
+            filter_split = gca_training_pipeline.FilterSplit(
```
sasha-gitg, line 255: What's the expected behavior of the service if both FractionSplit and FilterSplit are passed in? 

It seems to me that only one should be passed in. Under that assumption it seems like we shouldn't default to creating FractionSplit and FilterSplit as that will automatically cause this service call to fail. Or does the service maintain a precedence ordering for Splits?

sasha-gitg (changes requested):

PR #207, opened 2021-02-03
feat: Add accelerator options to Model/Endpoint deploy()

vinnysenthil:
Addition of `accelerator_type` and `accelerator_count` to `Endpoint.deploy()` and `Model.deploy()` methods. Also added a util method to validate the `accelerator_type` against the GAPIC Enum.

Manually tested that deploying with GPUs work. Does not affect existing tests.

Fixes [b/179069467](http://b/179069467) 🦕

PR #211, opened 2021-02-05
fix: Add tests and final touches for PR #207

vinnysenthil:
- Update `deploy()` tests to include passing `accelerator_type` and `accelerator_count`
- Add test for `validate_accelerator_type()`
- Include passing accelerator params in a call to Endpoint._deploy_call() that was missed in #207 
-  Add return type to util method

Fixes [b/179294827](http://b/179294827) 🦕


sasha-gitg (approved):
LGTM!

PR #219, opened 2021-02-11
feat: Add SDK-wide delete method, cancel methods to Job and TrainingJob

vinnysenthil:
## Summary of changes
- Add `_delete_method` abstract property to `AiPlatformResourceNoun` class, requiring a resource noun to define it's delete method in GAPIC (i.e. `Dataset` would set this property to `delete_dataset`)
- Add abstract property `_cancel_method` to `_Job` class to force children to define it's cancel method (i.e. `BatchPredictionJob` would set `cancel_batch_prediction_job`
- Add high-level `delete()` method to `AiPlatformResourceNounWithFutureManager` that calls subclass deletion method
- Shifted the `optional_sync` decorator higher in `base.py` to be referenced by `delete()` for async capability
- Removed `Dataset.delete()` and `Model.delete()` but left their tests; they now test the high-level delete method
- Add `_TrainingJob.cancel()` and `_Job.cancel()` along with unit tests

Fixes [b/179510193](http://b/179510193) and [b/179945882](http://b/179945882) 🦕

### DIFF
--- a/google/cloud/aiplatform/models.py
+++ b/google/cloud/aiplatform/models.py
@@ -920,19 +920,6 @@ class Endpoint(base.AiPlatformResourceNounWithFutureManager):
 
         return self
 
-    @base.optional_sync()
-    def _delete(self, sync: bool = True) -> None:
-        """Private helper method to delete this endpoint via GAPIC.
-
-        Args:
-            sync (bool):
-                Whether to execute this method synchronously. If False, this method
-                will be executed in concurrent Future and any downstream object will
-                be immediately returned and synced when the Future has completed.
-        """
-        lro = self.api_client.delete_endpoint(name=self.resource_name)
-        lro.result()
-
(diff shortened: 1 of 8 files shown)


google/cloud/aiplatform/models.py
```diff
@@ -950,7 +951,7 @@ def delete(self, force: bool = False, sync: bool = True) -> None:
         if force:
             self.undeploy_all(sync=sync)
 
-        self._delete(sync=sync)
```
sasha-gitg, line 953: Can this method be removed?

sasha-gitg (approved):
LGTM

google/cloud/aiplatform/models.py
vinnysenthil: Yup, good call. Dropping it

ivanmkc (approved):
LGTM

PR #220, opened 2021-02-11
feat: Added AutoMLTextTrainingJob and tests

ivanmkc:
This adds the AutoMLTextTrainingJob subclass:

- [x] Added Added AutoMLTextTrainingJob
- [x] Added unit tests
- [x] Tested CUJ for Text classification: [uCAIP MB SDK - AutoML Text Training Demo - Classification](https://colab.research.google.com/drive/1uO3VvWfr1dz0YqFBeJhFeXoHlsWT8sUf?resourcekey=0-KHj-QHOZZd2jOSv7oSn6Ew&usp=sharing)
- [x] Tested CUJ for Text extraction: [uCAIP MB SDK - AutoML Text Training Demo - Extraction](https://colab.research.google.com/drive/1kZMNPIyBgTeYvvjyAfMiyie8OtjayNOn?resourcekey=0-ru5W2u1Rv6uhIAxLPfjVcw&usp=sharing)
- [x] Tested CUJ for Text sentiment: [uCAIP MB SDK - AutoML Text Training Demo - Sentiment](https://colab.research.google.com/drive/1YTR57-UWWdyrXLLsUTDVg-1gELHghaFo?resourcekey=0-IOIH12h1dnMhHWxBDDnhJw&usp=sharing)

Fixes #https://b.corp.google.com/issues/179106243

### DIFF
--- a/google/cloud/aiplatform/training_jobs.py
+++ b/google/cloud/aiplatform/training_jobs.py
@@ -3578,8 +3578,6 @@ class AutoMLTextTrainingJob(_TrainingJob):
                 and what is compatible should be described in the used
                 TrainingPipeline's [training_task_definition]
                 [google.cloud.aiplatform.v1beta1.TrainingPipeline.training_task_definition].
-                For Text Datasets, all their data is exported to
-                training, to pick and choose from.
             training_fraction_split: float = 0.8
                 Required. The fraction of the input data that is to be
                 used to train the Model. This is ignored if Dataset is not provided.
@@ -3595,12 +3593,6 @@ class AutoMLTextTrainingJob(_TrainingJob):
                 of any UTF-8 characters.
 
                 If not provided upon creation, the job's display_name is used.
-            disable_early_stopping (bool):
-                Required. If true, the entire budget is used. This disables the early stopping
-                feature. By default, the early stopping feature is enabled, which means
-                that training might stop before the entire training budget has been
-                used, if further training does no longer brings significant improvement
-                to the model.
             sync (bool):
                 Whether to execute this method synchronously. If False, this method
                 will be executed in concurrent Future and any downstream object will
@@ -3665,16 +3657,6 @@ class AutoMLTextTrainingJob(_TrainingJob):
             test_fraction_split (float):
                 Required. The fraction of the input data that is to be
                 used to evaluate the Model. This is ignored if Dataset is not provided.
-            predefined_split_column_name (str):
-                Optional. The key is a name of one of the Dataset's data
-                columns. The value of the key (either the label's value or
-                value in the column) must be one of {``training``,
-                ``validation``, ``test``}, and it defines to which set the
-                given piece of data is assigned. If for a piece of data the
-                key is not present or has an invalid value, that piece is
-                ignored by the pipeline.
-
-                Supported only for Text Datasets.
             model_display_name (str):
                 Optional. If the script produces a managed AI Platform Model. The display name of
                 the Model. The name can be up to 128 characters long and can be consist
--- a/google/cloud/aiplatform/training_jobs.py
+++ b/google/cloud/aiplatform/training_jobs.py
@@ -50,10 +50,8 @@ from google.cloud.aiplatform_v1beta1.types import (
     training_pipeline as gca_training_pipeline,
 )
 
-from google.cloud.aiplatform.v1beta1.schema.trainingjob.definition_v1beta1 import (
-    AutoMlTextClassificationInputs,
-    AutoMlTextExtractionInputs,
-    AutoMlTextSentimentInputs,
+from google.cloud.aiplatform.v1beta1.schema.trainingjob import (
+    definition_v1beta1 as training_job_inputs,
 )
 
 from google.cloud import storage
@@ -268,7 +266,7 @@ class _TrainingJob(base.AiPlatformResourceNounWithFutureManager):
     def _run_job(
         self,
         training_task_definition: str,
-        training_task_inputs: proto.Message,
+        training_task_inputs: Union[Dict, proto.Message],
         dataset: Optional[datasets.Dataset],
         training_fraction_split: float,
         validation_fraction_split: float,
@@ -296,7 +294,7 @@ class _TrainingJob(base.AiPlatformResourceNounWithFutureManager):
                 point to a location where the user only has a
                 read access.
             training_task_inputs (proto.Message):
-                Required. The training task's input that correspons to the training_task_definition parameter.
+                Required. The training task's input that corresponds to the training_task_definition parameter. 
             dataset (datasets.Dataset):
                 The dataset within the same Project from which data will be used to train the Model. The
                 Dataset must use schema compatible with Model being trained,
@@ -3508,15 +3506,11 @@ class AutoMLTextTrainingJob(_TrainingJob):
             location=location,
             credentials=credentials,
         )
-        self._prediction_type = prediction_type
-        self._multi_label = multi_label
-        self._sentiment_max = sentiment_max
-        self._model_hint = model_hint
 
         training_task_definition: str
         training_task_inputs_dict: proto.Message
 
-        if self._prediction_type == "classification":
+        if prediction_type == "classification":
             training_task_definition = (
                 schema.training_job.definition.automl_text_classification
             )
@@ -3526,27 +3520,29 @@ class AutoMLTextTrainingJob(_TrainingJob):
                     "The multi_label parameter must be provided for a prediction_type of 'classification'"
                 )
 
-            training_task_inputs_dict = AutoMlTextClassificationInputs(
-                multi_label=self._multi_label
+            training_task_inputs_dict = training_job_inputs.AutoMlTextClassificationInputs(
+                multi_label=multi_label
             ).to_value()
         elif self._prediction_type == "extraction":
             training_task_definition = (
                 schema.training_job.definition.automl_text_extraction
             )
 
-            training_task_inputs_dict = AutoMlTextExtractionInputs().to_value()
-        elif self._prediction_type == "sentiment":
+            training_task_inputs_dict = (
+                training_job_inputs.AutoMlTextExtractionInputs().to_value()
+            )
+        elif prediction_type == "sentiment":
             training_task_definition = (
                 schema.training_job.definition.automl_text_sentiment
             )
 
-            if self._sentiment_max is None:
+            if sentiment_max is None:
                 raise ValueError(
                     "Tge sentiment_max parameter must be provided for a prediction_type of 'sentiment'"
                 )
 
-            training_task_inputs_dict = AutoMlTextSentimentInputs(
-                sentiment_max=self._sentiment_max
+            training_task_inputs_dict = training_job_inputs.AutoMlTextSentimentInputs(
+                sentiment_max=sentiment_max
             ).to_value()
         else:
             raise ValueError(
--- a/google/cloud/aiplatform/training_jobs.py
+++ b/google/cloud/aiplatform/training_jobs.py
@@ -50,10 +50,18 @@ from google.cloud.aiplatform_v1beta1.types import (
     training_pipeline as gca_training_pipeline,
 )
 
+from google.cloud.aiplatform.v1beta1.schema.trainingjob.definition_v1beta1 import (
+    AutoMlTextClassificationInputs,
+    AutoMlTextExtractionInputs,
+    AutoMlTextSentimentInputs,
+)
+
 from google.cloud import storage
 from google.protobuf import json_format
 from google.protobuf import struct_pb2
 from google.rpc import code_pb2
+
+import proto
 
 logging.basicConfig(level=logging.INFO, stream=sys.stdout)
 _LOGGER = logging.getLogger(__name__)
@@ -262,7 +270,7 @@ class _TrainingJob(base.AiPlatformResourceNounWithFutureManager):
     def _run_job(
         self,
         training_task_definition: str,
-        training_task_inputs: dict,
+        training_task_inputs: proto.Message,
         dataset: Optional[datasets.Dataset],
         training_fraction_split: float,
         validation_fraction_split: float,
@@ -406,9 +414,7 @@ class _TrainingJob(base.AiPlatformResourceNounWithFutureManager):
         training_pipeline = gca_training_pipeline.TrainingPipeline(
             display_name=self._display_name,
             training_task_definition=training_task_definition,
-            training_task_inputs=json_format.ParseDict(
-                training_task_inputs, struct_pb2.Value()
-            ),
+            training_task_inputs=training_task_inputs,
             model_to_upload=model,
             input_data_config=input_data_config,
         )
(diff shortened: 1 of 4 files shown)


google/cloud/aiplatform/training_jobs.py
```diff
@@ -50,11 +50,19 @@
     training_pipeline as gca_training_pipeline,
 )
 
+from google.cloud.aiplatform.v1beta1.schema.trainingjob.definition_v1beta1 import (
```
ivanmkc, line 53: Uses EPCL types

google/cloud/aiplatform/training_jobs.py
```diff
@@ -50,11 +50,19 @@
...
+    AutoMlTextExtractionInputs,
+    AutoMlTextSentimentInputs,
+)
+
 from google.cloud import storage
 from google.protobuf import json_format
 from google.protobuf import struct_pb2
 from google.rpc import code_pb2
 
+import proto
```
ivanmkc, line 64: Superclass for EPCL's

google/cloud/aiplatform/training_jobs.py
```diff
@@ -3346,3 +3352,265 @@ def _run(
...
+        credentials: Optional[auth_credentials.Credentials] = None,
+    ):
+        """Constructs a AutoML Text Training Job.
+
+        Args:
+            display_name (str):
+                Required. The user-defined name of this TrainingPipeline.
+            prediction_type (str):
+                The type of prediction the Model is to produce, one of:
+            TODO
```
ivanmkc, line 3376: TODO

google/cloud/aiplatform/training_jobs.py
```diff
@@ -3346,3 +3352,265 @@ def _run(
...
+        )
+        self._prediction_type = prediction_type
+        self._multi_label = multi_label
+        self._sentiment_max = sentiment_max
+        self._model_hint = model_hint
+
+        training_task_definition: str
+        training_task_inputs_dict: proto.Message
+
+        if self._prediction_type == "classification":
```
ivanmkc, line 3416: Since there are no switch statements, is there a better way than to do this in python?

I don't quite like this as it couples each prediction_type with each other into one if-else statement.
May refactor to decouple them.

ivanmkc: Well, since we've flattened the hierarchy, the coupling is hard to avoid.

tests/unit/aiplatform/test_automl_text_training_jobs.py
```diff
@@ -0,0 +1,532 @@
...
+                sync=sync,
+            )
+
+            if not sync:
+                job.wait()
+
+        with pytest.raises(RuntimeError):
+            job.get_model()
+
+    def test_raises_before_run_is_called(self, mock_pipeline_service_create):
```
ivanmkc, line 516: @sasha-gitg Perhaps remove this as it doesn't add code coverage

tests/unit/aiplatform/test_automl_text_training_jobs.py
```diff
@@ -0,0 +1,532 @@
...
+            job.run(
+                dataset=mock_dataset_text,
+                model_display_name=_TEST_MODEL_DISPLAY_NAME,
+                training_fraction_split=_TEST_FRACTION_SPLIT_TRAINING,
+                validation_fraction_split=_TEST_FRACTION_SPLIT_VALIDATION,
+                test_fraction_split=_TEST_FRACTION_SPLIT_TEST,
+                sync=sync,
+            )
+
+    @pytest.mark.parametrize("sync", [True, False])
```
ivanmkc, line 487: @sasha-gitg Perhaps remove this as it doesn't add code coverage

google/cloud/aiplatform/training_jobs.py
```diff
@@ -288,11 +294,8 @@ def _run_job(
                 than the one given on input. The output URI will
                 point to a location where the user only has a
                 read access.
-            training_task_inputs (dict):
-                Required. The training task's parameter(s), as specified in
-                the
-                ``training_task_definition``'s
-                ``inputs``.
+            training_task_inputs (proto.Message):
```
ivanmkc, line 297: Since we pass in the proto, it'd be nice if we could just remove the `training_task_definition yaml file` as the proto basically represents the same thing. 

Since the service still needs this passed in, we could use a dictionary to infer the correct `training_task_definition yaml file URI`.

@sasha-gitg thoughts?

tests/unit/aiplatform/test_automl_text_training_jobs.py
sasha-gitg: Does it add branch coverage?

tests/unit/aiplatform/test_automl_text_training_jobs.py
sasha-gitg: Does it add branch coverage?

google/cloud/aiplatform/training_jobs.py
sasha-gitg: Nice

tests/unit/aiplatform/test_automl_text_training_jobs.py
ivanmkc: You know, actually it tests for this one line, so let's keep it:

```
if self._has_run:
            raise RuntimeError("AutoML Text Training has already run.")
```

tests/unit/aiplatform/test_automl_text_training_jobs.py
ivanmkc: No

google/cloud/aiplatform/training_jobs.py
dinagraves: May I suggest something like

```
        mapping = {
            "classification": {
                "task": schema.training_job.definition.automl_text_classification_task,
                "func": AutoMlTextClassificationInputs,
                "required_field": "multi_label"
                },
           "extraction": {
                "task": schema.training_job.definition.automl_text_extraction_task,
                "func": AutoMlTextExtractionInputs,
                "required_field": ""
                },
            "sentiment": {
                "task": schema.training_job.definition.automl_text_sentiment_task,
                "func": AutoMlTextSentimentInputs,
                "required_field": "sentiment_max"
            }
        }

        if self._prediction_type not in mapping:
            raise ValueError(
                "Prediction type must be one of 'classification', 'extraction', or 'sentiment'."
            )

        prediction_type = mapping[self._prediction_type]
        training_task_definition = prediction_type["task"]
        training_func = prediction_type["func"]
        required_field = prediction_type["required_field"]
        args = {}

        if required_field:
            self_required_field = self.__getattribute__("_" + required_field)

            # If there is a field required and it is missing
            if self_required_field is None:               
                raise ValueError("Required parameter missing:" + prediction_type["required_field"]
                    )
            args = {required_field: self_required_field}

        training_task_inputs_dict = training_func(**args).to_value()
```

google/cloud/aiplatform/training_jobs.py
```diff
@@ -3346,3 +3347,255 @@ def _run(
...
+
+            if self._sentiment_max is None:
+                raise ValueError(
+                    "Tge sentiment_max parameter must be provided for a prediction_type of 'sentiment'"
+                )
+
+            training_task_inputs_dict = AutoMlTextSentimentInputs(
+                sentiment_max=self._sentiment_max
+            ).to_value()
+        else:
```
dinagraves, line 3451: We should check for this first and exit early if it's not one of the required types.

google/cloud/aiplatform/training_jobs.py
ivanmkc: Thanks for contributing! There are some limitations to this particular solution though, such as only supporting a single `self_required_field` and requiring certain variables being named a certain way ('the underscore prefix').

I think in this case, the if-else makes things most explicit whereas preserving flexibility.

ivanmkc: The 'correct' approach would just be to unflatten the function signature and just encapsulate all required parameters into its own class (i.e. use the protos). However, for UX reasons it's been decided that flattening the various prediction types (and respective parameters) was preferable.

google/cloud/aiplatform/training_jobs.py
ivanmkc: Not sure I understand correctly, but I can move this check higher up in the function. I'd refrain from adding an additional check that repeats this logic (to reduce potential for bugs in the future if someone forgets to sync that logic with this if-else statement).

google/cloud/aiplatform/training_jobs.py
```diff
@@ -3346,3 +3347,255 @@ def _run(
...
+
+
+class AutoMLTextTrainingJob(_TrainingJob):
+    def __init__(
+        self,
+        display_name: str,
+        prediction_type: str,
+        multi_label: Optional[bool] = None,
+        sentiment_max: Optional[str] = None,
+        model_hint: str = "default",
```
sasha-gitg, line 3359: This seems to be missing in the docstring.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -3346,3 +3347,255 @@ def _run(
...
+                        prevailing emotional opinion within it, especially to determine a writer's attitude
+                        as positive, negative, or neutral.
+            multi_label (bool):
+                Required and only applicable for text classification task. If false, a single-label (multi-class) Model will be trained (i.e.
+                assuming that for each text snippet just up to one annotation may be
+                applicable). If true, a multi-label Model will be trained (i.e.
+                assuming that for each text snippet multiple annotations may be
+                applicable).
+            sentiment_max (int):
+                Required and only applicable for sentiment task. A sentiment is expressed as an integer
```
sasha-gitg, line 3387: This looks optional.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -3346,3 +3347,255 @@ def _run(
...
+                        returns a list of categories that apply to the text found in the data.
+                        AI Platform offers both single-label and multi-label text classification models.
+                    "extraction" - An entity extraction model inspects text data
+                        for known entities referenced in the data and
+                        labels those entities in the text.
+                    "sentiment" - A sentiment analysis model inspects text data and identifies the
+                        prevailing emotional opinion within it, especially to determine a writer's attitude
+                        as positive, negative, or neutral.
+            multi_label (bool):
+                Required and only applicable for text classification task. If false, a single-label (multi-class) Model will be trained (i.e.
```
sasha-gitg, line 3381: Does it make sense to default this to False and ignore the argument if "classification" is not passed in.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -3346,3 +3347,255 @@ def _run(
...
+                Optional. Custom credentials to use to run call training service. Overrides
+                credentials set in aiplatform.init.
+        """
+        super().__init__(
+            display_name=display_name,
+            project=project,
+            location=location,
+            credentials=credentials,
+        )
+        self._prediction_type = prediction_type
```
sasha-gitg, line 3411: These attributes don't seem necessary since they are enveloped in training_task_input_dict.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -3346,3 +3347,255 @@ def _run(
...
+        self._multi_label = multi_label
+        self._sentiment_max = sentiment_max
+        self._model_hint = model_hint
+
+        training_task_definition: str
```
sasha-gitg, line 3416: No need to define these just for the type signature. The type signatures can be inlined with assignment but generally in cases where the type is obvious, it's not necessary.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -3346,3 +3347,255 @@ def _run(
...
+                Required. The fraction of the input data that is to be
+                used to train the Model. This is ignored if Dataset is not provided.
+            validation_fraction_split: float = 0.1
+                Required. The fraction of the input data that is to be
+                used to validate the Model. This is ignored if Dataset is not provided.
+            test_fraction_split: float = 0.1
+                Required. The fraction of the input data that is to be
+                used to evaluate the Model. This is ignored if Dataset is not provided.
+            model_display_name (str):
+                Optional. If the script produces a managed AI Platform Model. The display name of
```
sasha-gitg, line 3494: I think the "If the script produces..." can be removed because this will always create a managed model.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -3346,3 +3347,255 @@ def _run(
...
+                the Model. The name can be up to 128 characters long and can be consist
+                of any UTF-8 characters.
+
+                If not provided upon creation, the job's display_name is used.
+            sync (bool):
+                Whether to execute this method synchronously. If False, this method
+                will be executed in concurrent Future and any downstream object will
+                be immediately returned and synced when the Future has completed.
+        Returns:
+            model: The trained AI Platform Model resource or None if training did not
```
sasha-gitg, line 3504: Similarly, the return None can be removed.

sasha-gitg (changes requested):
Added comments. Thanks!

google/cloud/aiplatform/training_jobs.py
dinagraves: To your first comment -- it would be pretty simple to support a list of fields: 
`"required_fields": ["multi_label"]"`. 

If there are no fields, it would simply be an empty list.  

If we cannot use protos in the solution, I think having some kind of mapping is the next best option.  I think the value of using an approach like this is that it is more readable - both in terms of logic and whitespace.  It can be hard to follow so many if/else statements, especially if they have nested if statements.   It also results in having a large amount of code indented far to the right, which means you have to start breaking up strings and functions, etc, to keep it under 80 characters.

ivanmkc: Thanks, those are great points. I'll discuss it with the team as this issue appears across a variety of our training jobs and might be useful in other places.

google/cloud/aiplatform/training_jobs.py
ivanmkc: Okay good to know. It will be cleaner to keep the concepts separate then.

google/cloud/aiplatform/training_jobs.py
ivanmkc: Done

google/cloud/aiplatform/training_jobs.py
```diff
@@ -261,7 +269,7 @@ def _create_input_data_config(
     def _run_job(
         self,
         training_task_definition: str,
-        training_task_inputs: dict,
+        training_task_inputs: proto.Message,
```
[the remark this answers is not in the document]
ivanmkc: For the custom jobs, I wrapped the dict like this before passing to `_run_job`. Are you okay with that?

```
json_format.ParseDict(
                training_task_inputs, struct_pb2.Value()
            )
```

google/cloud/aiplatform/training_jobs.py
```diff
@@ -405,9 +408,7 @@ def _run_job(
         training_pipeline = gca_training_pipeline.TrainingPipeline(
             display_name=self._display_name,
             training_task_definition=training_task_definition,
-            training_task_inputs=json_format.ParseDict(
-                training_task_inputs, struct_pb2.Value()
-            ),
+            training_task_inputs=training_task_inputs,
```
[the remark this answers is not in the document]
ivanmkc: I moved the conversion upstream into the CustomTrainingJob variants.

google/cloud/aiplatform/training_jobs.py
ivanmkc: We should do that if that's the convention across MBSDK. Will change.

google/cloud/aiplatform/training_jobs.py
ivanmkc: Good catch. Missed these after refactoring.

google/cloud/aiplatform/training_jobs.py
ivanmkc: I just wanted to make it explicit what the scope of this variable was.

google/cloud/aiplatform/training_jobs.py
ivanmkc: Done

google/cloud/aiplatform/training_jobs.py

google/cloud/aiplatform/training_jobs.py
ivanmkc: Removed

google/cloud/aiplatform/training_jobs.py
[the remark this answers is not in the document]
ivanmkc: The conversion is moved upstream to the callers in the CustomJob.

google/cloud/aiplatform/training_jobs.py
ivanmkc: It's not optional for the sentiment case.

google/cloud/aiplatform/training_jobs.py
ivanmkc: Done.

google/cloud/aiplatform/training_jobs.py
ivanmkc: Done

google/cloud/aiplatform/training_jobs.py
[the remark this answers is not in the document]
sasha-gitg: That works for me, but the other AutoML classes are passing in dicts as well: https://github.com/googleapis/python-aiplatform/blob/4f5797f65af03fc781c778e6f2d99c0f2d468920/google/cloud/aiplatform/training_jobs.py#L2643

So it may be cleaner and less code to accept Union[dict, proto.message] and conditionally create the struct_pb2.Value if passed a dict. Though, we may be able to bypass `json_format.ParseDict` and directly pass in the dict based on this: https://github.com/googleapis/proto-plus-python/blob/ccf903e3cddfcb1ff539e853594b4342914b7d61/docs/messages.rst#usage

tests/unit/aiplatform/test_automl_text_training_jobs.py
```diff
@@ -0,0 +1,509 @@
...
+
+class TestAutoMLTextTrainingJob:
+    def setup_method(self):
+        importlib.reload(initializer)
+        importlib.reload(aiplatform)
+
+    def teardown_method(self):
+        initializer.global_pool.shutdown(wait=True)
+
+    def test_init_all_parameters_classification(self, mock_model):
```
sasha-gitg, line 147: Does the `mock_model` fixture need to be passed in these tests?

sasha-gitg (approved):
LGTM! Added two comments.

google/cloud/aiplatform/training_jobs.py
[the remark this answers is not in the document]
ivanmkc: I imagined we would replace them with EPCL's at one point.

I will do this:
1. Change the docstring to allow a `Union of Dict + EPCL types`. 

I've also tested that passing in the `training_task_inputs_dict` as a dict WITHOUT conversion works fine.

sasha-gitg (approved):
LGTM!

sasha-gitg:
Thank you for including the Notebooks!

ivanmkc:
> Thank you for including the Notebooks!

No problem, I captured a bug for final cleanup before ready for the next bug bash (https://b.corp.google.com/issues/181027098)

PR #237, opened 2021-02-19
feat: Added AutoMLForecastingTrainingJob and tests

thehardikv:
This adds the `AutoMLForecastingTrainingJob` subclass:

- [x]  Added `AutoMLForecastingTrainingJob`
- [x]  Added unit tests
- [x]  Tested [CUJ for Forecasting](http://doc/1jKpbfHHxQK_V5XjCGmdoslpPyxp3yowQePuR5q-O1MA)

Fixes [b/170511584](http://b/170511584) 🦕


google/cloud/aiplatform/__init__.py
```diff
@@ -32,6 +32,7 @@
     CustomContainerTrainingJob,
```
thehardikv, line 32: Are the changes in this file needed?

thehardikv:
Should we consider Forecasting to be a type of Tabular? So instead of referencing "Forecasting" as a standalone product in the SDK, we could use "TabularForecasting"?

vinnysenthil (approved):
LGTM! Added a few suggestions and a question

google/cloud/aiplatform/__init__.py
vinnysenthil: Yes, it allows the user to directly access the class via `aiplatform.AutoMLForecastingTrainingJob` instead of `aiplatform.training_jobs.AutoMLForecastingTrainingJob`

google/cloud/aiplatform/schema.py
```diff
@@ -22,6 +22,7 @@ class training_job:
     class definition:
         custom_task = "gs://google-cloud-aiplatform/schema/trainingjob/definition/custom_task_1.0.0.yaml"
         automl_tabular = "gs://google-cloud-aiplatform/schema/trainingjob/definition/automl_tabular_1.0.0.yaml"
+        automl_forecasting = "gs://google-cloud-aiplatform/schema/trainingjob/definition/automl_forecasting_1.0.0.yaml"
```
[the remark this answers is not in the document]
thehardikv: It does exist with a different name, [automl_time_series_forecasting_1.0.0.yaml](http://google3/google/cloud/aiplatform/publicfiles/trainingjob/definition/automl_time_series_forecasting_1.0.0.yaml), but the API validation is based off, http://google3/java/com/google/cloud/ai/platform/boq/common/constants/ApiConstants.java?l=337. I tried fixing the latter but it ended up breaking the UI. This change will require more coordination and will come later.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -2673,6 +2673,455 @@ def _model_upload_fail_string(self) -> str:
...
+        time_column: str,
+        time_series_identifier_column: str,
+        time_variant_past_only_columns: List[str],
+        time_variant_past_and_future_columns: List[str],
+        forecast_window_end: int,
+        period_unit: str,
+        period_count: int,
+        predefined_split_column_name: Optional[str] = None,
+        weight_column: Optional[str] = None,
+        static_columns: List[str] = [],
```
sasha-gitg, line 2741: We should avoid mutable default values. Perhaps set this to Optional and default to None.

sasha-gitg (changes requested):
LGTM! Just one requested change on a default value.

vinnysenthil:
@thehardikv looks like the lint is failing, run `nox -s lint` in the root folder to check why. I think we might need to run [Black](https://github.com/psf/black) on `training_jobs.py`

google/cloud/aiplatform/training_jobs.py
```diff
@@ -2673,6 +2673,455 @@ def _model_upload_fail_string(self) -> str:
...
+                will be executed in concurrent Future and any downstream object will
+                be immediately returned and synced when the Future has completed.
+        Returns:
+            model: The trained AI Platform Model resource or None if training did not
+                produce an AI Platform Model.
+        """
+
+        training_task_definition = schema.training_job.definition.automl_forecasting
+
+        training_task_inputs_dict = {
```
ivanmkc, line 3069: You should use the EPCL Python types now instead of a dictionary, unless Forecasting has some special restrictions.

```
from google.cloud.aiplatform.v1beta1.schema.trainingjob import (
    definition_v1beta1 as training_job_inputs,
)

...

training_job_inputs.AutoMlForecastingInputs(...)
```

thehardikv: Hmm...how would that work? Seems the `_run_job` method expects a `dict`: https://github.com/googleapis/python-aiplatform/blob/dev/google/cloud/aiplatform/training_jobs.py#L265

google/cloud/aiplatform/datasets/_datasources.py
```diff
@@ -222,6 +222,11 @@ def create_datasource(
             raise ValueError("tabular dataset does not support data import.")
         return TabularDatasource(gcs_source, bq_source)
 
+    if metadata_schema_uri == schema.dataset.metadata.time_series:
```
ivanmkc, line 225: LGTM

google/cloud/aiplatform/training_jobs.py
ivanmkc: Yesterday's merge of https://github.com/googleapis/python-aiplatform/pull/220 changed it to `training_task_inputs: Union[dict, proto.Message]`.

However, I think Forecasting has had the EPCL removed so you might be restricted to using the dict for now.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -2673,6 +2673,459 @@ def _model_upload_fail_string(self) -> str:
...
+            "optimizationObjective": self._optimization_objective,
+        }
+
+        final_export_eval_bq_uri = export_evaluated_data_items_bigquery_destination_uri
+        if final_export_eval_bq_uri and not final_export_eval_bq_uri.startswith(
+            "bq://"
+        ):
+            final_export_eval_bq_uri = f"bq://{final_export_eval_bq_uri}"
+
+        if export_evaluated_data_items:
```
ivanmkc, line 3098: @sasha-gitg Seems like `definition/automl_tabular_1.0.0.yaml` also defines `exportEvaluatedDataItemsConfig`. This is unimplemented for `AutoMLTabularTrainingJob`.

I'll add a bug to track this.

google/cloud/aiplatform/__init__.py
```diff
@@ -52,6 +53,7 @@
     "init",
     "AutoMLImageTrainingJob",
     "AutoMLTabularTrainingJob",
```
sasha-gitg, line 55: TimeSeriesDataset should also be added here.

sasha-gitg (approved):

PR #238, opened 2021-02-19
feat: Made get function available to all training jobs

ivanmkc:
Lifted `get` function to `_TrainingJob` so that all subclasses can use this function to get an existing TrainingJob.

Fixes #https://b.corp.google.com/issues/180746020

### DIFF
--- a/google/cloud/aiplatform/training_jobs.py
+++ b/google/cloud/aiplatform/training_jobs.py
@@ -3465,6 +3465,12 @@ class CustomPythonPackageTrainingJob(_CustomTrainingJob):
 
 
 class AutoMLTextTrainingJob(_TrainingJob):
+    _supported_training_schemas = (
+        schema.training_job.definition.automl_text_classification,
+        schema.training_job.definition.automl_text_extraction,
+        schema.training_job.definition.automl_text_sentiment,
+    )
+
     def __init__(
         self,
         display_name: str,
--- a/google/cloud/aiplatform/training_jobs.py
+++ b/google/cloud/aiplatform/training_jobs.py
@@ -107,9 +107,10 @@ class _TrainingJob(base.AiPlatformResourceNounWithFutureManager):
         self._project = project
         self._gca_resource = None
 
+    @property
     @classmethod
     @abc.abstractmethod
-    def _supported_training_schemas(self) -> Tuple[str]:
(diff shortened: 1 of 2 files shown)


google/cloud/aiplatform/training_jobs.py
```diff
@@ -103,6 +103,65 @@ def __init__(
...
+
+        Raises:
+            ValueError: If the retrieved training job's training task definition
+                doesn't match the custom training task definition.
+
+        Returns:
+            An AI Platform Training Job
+        """
+
+        # Create job with dummy parameters
```
sasha-gitg, line 144: I created a bug to extract project and location from resource name (b/180923329) if given using this parser: https://github.com/googleapis/python-aiplatform/blob/b7aeeb637b7c36792a2de0b65c868121a4a7e412/google/cloud/aiplatform_v1beta1/services/pipeline_service/client.py#L197

We can add as a follow up for all classes.

sasha-gitg (changes requested):
Added a few comments. Thanks!

google/cloud/aiplatform/training_jobs.py
```diff
@@ -103,6 +103,65 @@ def __init__(
         self._project = project
         self._gca_resource = None
 
+    @classmethod
+    @abc.abstractmethod
+    def _supported_training_schemas(self) -> Tuple[str]:
```
sasha-gitg, line 108:
```suggestion
    def _supported_training_schemas(cls) -> Tuple[str]:
```

sasha-gitg: The property decorator is appropriate here and is consistent with the abstract class property pattern we use in the SDK:

https://github.com/googleapis/python-aiplatform/blob/dev/google/cloud/aiplatform/base.py#L223

Concrete class usage:
https://github.com/googleapis/python-aiplatform/blob/dev/google/cloud/aiplatform/training_jobs.py#L72

sasha-gitg (changes requested):

google/cloud/aiplatform/training_jobs.py
ivanmkc: Ok I was wondering if they were compatible and took it off just to be safe as it didn't change too much.

I will add it back.

sasha-gitg (approved):
LGTM!

PR #256, opened 2021-03-03
feat: Added encryption_spec_key_name to all relevant resources

ivanmkc:
The encryption key can be passed into aiplatform.init which will be applied to all applicable resources:

```
aiplatform.init(
            project=_TEST_PROJECT,
            encryption_spec_key_name=_TEST_DEFAULT_ENCRYPTION_KEY_NAME,
        )
```

Alternatively, it can be passed directly into the resource itself on creation. This will override the aiplatform default.

1. Added CMEK support to:
- [x] Dataset
- [x] ImageDataset
- [x] TabularDataset
- [x] TextDataset
- [x] VideoDataset
- [x] Endpoint
- [x] Batch Prediction Job
- [x] Model
- [x] Training Pipeline
- [x] _TrainingJob
- [x] _CustomTrainingJob
- [x] CustomPythonPackageTrainingJob
- [x] CustomContainerTrainingJob
- [x] CustomTrainingJob
- [x] AutoMLTabularTrainingJob
- [x] AutoMLImageTrainingJob
- [x] AutoMLTextTrainingJob

2. Added tests:
- [x] Dataset
- [x] ImageDataset
- [x] TabularDataset
- [x] TextDataset
- [x] VideoDataset
- [x] Endpoint
- [x] Batch Prediction Job
- [x] Model
- [x] Training Pipeline
- [x] _TrainingJob
- [x] _CustomTrainingJob
- [x] CustomPythonPackageTrainingJob
- [x] CustomContainerTrainingJob
- [x] CustomTrainingJob
- [x] AutoMLTabularTrainingJob
- [x] AutoMLImageTrainingJob
- [x] AutoMLTextTrainingJob

Currently Unused:
- [ ] Custom Job
- [ ] Hyper-parameter Tuning

Fixes [b/178500930](https://b.corp.google.com/issues/178500930)

### DIFF
--- a/google/cloud/aiplatform/datasets/dataset.py
+++ b/google/cloud/aiplatform/datasets/dataset.py
@@ -67,6 +67,7 @@ class Dataset(base.AiPlatformResourceNounWithFutureManager):
             credentials (auth_credentials.Credentials):
                 Custom credentials to use to upload this model. Overrides
                 credentials set in aiplatform.init.
+
         """
 
         super().__init__(
--- a/google/cloud/aiplatform/datasets/dataset.py
+++ b/google/cloud/aiplatform/datasets/dataset.py
@@ -30,6 +30,7 @@ from google.cloud.aiplatform_v1beta1.types import dataset as gca_dataset
 from google.cloud.aiplatform_v1beta1.services.dataset_service import (
     client as dataset_service_client,
 )
+from google.cloud.aiplatform_v1beta1.types.encryption_spec import EncryptionSpec
 
 
 class Dataset(base.AiPlatformResourceNounWithFutureManager):
@@ -106,6 +107,7 @@ class Dataset(base.AiPlatformResourceNounWithFutureManager):
         location: Optional[str] = None,
         credentials: Optional[auth_credentials.Credentials] = None,
         request_metadata: Optional[Sequence[Tuple[str, str]]] = (),
+        encryption_spec_key_name: Optional[str] = None,
         sync: bool = True,
     ) -> "Dataset":
         """Creates a new dataset and optionally imports data into dataset when
@@ -166,6 +168,15 @@ class Dataset(base.AiPlatformResourceNounWithFutureManager):
                 credentials set in aiplatform.init.
             request_metadata (Sequence[Tuple[str, str]]):
                 Strings which should be sent along with the request as metadata.
+            encryption_spec_key_name (Optional[str]):
+                Optional. The Cloud KMS resource identifier of the customer
+                managed encryption key used to protect a resource. Has the
+                form:
+                ``projects/my-project/locations/my-region/keyRings/my-kr/cryptoKeys/my-key``.
+                The key needs to be in the same region as where the compute
+                resource is created.
+
+                If set, this Dataset and all sub-resources of this Dataset will be secured by this key.
             sync (bool):
                 Whether to execute this method synchronously. If False, this method
                 will be executed in concurrent Future and any downstream object will
@@ -201,6 +212,7 @@ class Dataset(base.AiPlatformResourceNounWithFutureManager):
             location=location or initializer.global_config.location,
             credentials=credentials or initializer.global_config.credentials,
             request_metadata=request_metadata,
+            encryption_spec_key_name=encryption_spec_key_name,
             sync=sync,
         )
 
@@ -217,6 +229,7 @@ class Dataset(base.AiPlatformResourceNounWithFutureManager):
         location: str,
         credentials: Optional[auth_credentials.Credentials],
         request_metadata: Optional[Sequence[Tuple[str, str]]] = (),
+        encryption_spec_key_name: Optional[str] = None,
         sync: bool = True,
     ) -> "Dataset":
         """Creates a new dataset and optionally imports data into dataset when
@@ -253,6 +266,15 @@ class Dataset(base.AiPlatformResourceNounWithFutureManager):
                 credentials set in aiplatform.init.
             request_metadata (Sequence[Tuple[str, str]]):
                 Strings which should be sent along with the request as metadata.
+            encryption_spec_key_name (Optional[str]):
+                Optional. The Cloud KMS resource identifier of the customer
+                managed encryption key used to protect a resource. Has the
+                form:
+                ``projects/my-project/locations/my-region/keyRings/my-kr/cryptoKeys/my-key``.
+                The key needs to be in the same region as where the compute
+                resource is created.
+
+                If set, this Dataset and all sub-resources of this Dataset will be secured by this key.
             sync (bool):
                 Whether to execute this method synchronously. If False, this method
                 will be executed in concurrent Future and any downstream object will
@@ -264,6 +286,7 @@ class Dataset(base.AiPlatformResourceNounWithFutureManager):
         """
 
         create_dataset_lro = cls._create(
+            encryption_spec_key_name=encryption_spec_key_name,
             api_client=api_client,
             parent=parent,
             display_name=display_name,
@@ -291,6 +314,7 @@ class Dataset(base.AiPlatformResourceNounWithFutureManager):
     @classmethod
     def _create(
         cls,
+        encryption_spec_key_name: Optional[str],
         api_client: dataset_service_client.DatasetServiceClient,
         parent: str,
         display_name: str,
@@ -301,6 +325,15 @@ class Dataset(base.AiPlatformResourceNounWithFutureManager):
         """Creates a new managed dataset by directly calling API client.
 
         Args:
+            encryption_spec_key_name (Optional[str]):
+                Optional. The Cloud KMS resource identifier of the customer
+                managed encryption key used to protect a resource. Has the
+                form:
+                ``projects/my-project/locations/my-region/keyRings/my-kr/cryptoKeys/my-key``.
+                The key needs to be in the same region as where the compute
+                resource is created.
+
+                If set, this Dataset and all sub-resources of this Dataset will be secured by this key.
             api_client (dataset_service_client.DatasetServiceClient):
                 An instance of DatasetServiceClient with the correct api_endpoint
                 already set based on user's preferences.
@@ -328,10 +361,16 @@ class Dataset(base.AiPlatformResourceNounWithFutureManager):
             operation (Operation):
                 An object representing a long-running operation.
         """
+
+        encryption_spec: Optional[EncryptionSpec] = None
+        if encryption_spec_key_name:
+            encryption_spec = EncryptionSpec(kms_key_name=encryption_spec_key_name)
+
         gapic_dataset = gca_dataset.Dataset(
             display_name=display_name,
             metadata_schema_uri=metadata_schema_uri,
             metadata=datasource.dataset_metadata,
+            encryption_spec=encryption_spec,
         )
 
         return api_client.create_dataset(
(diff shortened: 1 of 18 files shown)


google/cloud/aiplatform/datasets/dataset.py
```diff
@@ -197,10 +212,11 @@ def create(
             display_name=display_name,
             metadata_schema_uri=metadata_schema_uri,
             datasource=datasource,
-            project=project or initializer.global_config.project,
```
ivanmkc, line 200: Removed as this is redundant. The global_config's are used downstream if not already set.

ivanmkc:
I've modified existing tests to handle CMEK for the cases where it's passed into aiplatform.init and where it's passed into the job itself. Let me know if any of that is weird or could be improved.

google/cloud/aiplatform/initializer.py
```diff
@@ -59,9 +61,18 @@ def init(
             experiment (str): The experiment to assign
             staging_bucket (str): The default staging bucket to use to stage artifacts
                 when making API calls. In the form gs://...
-            credentials (google.auth.crendentials.Crendentials): The default custom
+            credentials (google.auth.crendentials.Credentials): The default custom
```
ivanmkc, line 64: Just fixed this as I came across it.

google/cloud/aiplatform/datasets/dataset.py
sasha-gitg: Double check this because this is a synchronous method calling into an optionally asynchronous method meaning the value of global_config could have changed between the call time of `create` and the execution time of `_create_and_import`.

google/cloud/aiplatform/datasets/dataset.py
```diff
@@ -323,15 +353,39 @@ def _create(
...
+                Overrides encryption_spec_key_name set in aiplatform.init.
         Returns:
             operation (Operation):
                 An object representing a long-running operation.
         """
+
+        # Use provided encryption key name or else use one from global config
+        kms_key_name = (
+            encryption_spec_key_name
+            or initializer.global_config.encryption_spec_key_name
```
sasha-gitg, line 375: Similar to the above comment. This method can potentially be called asynchronously so defaulting should happen in the synchronous caller.

This seems applicable to all changes in this PR.

google/cloud/aiplatform/datasets/dataset.py
```diff
@@ -30,6 +30,7 @@
 from google.cloud.aiplatform_v1beta1.services.dataset_service import (
     client as dataset_service_client,
 )
+from google.cloud.aiplatform_v1beta1.types import encryption_spec as gca_encryption_spec
```
sasha-gitg, line 33: Please reorganize the import blocks to be alphabetical.

google/cloud/aiplatform/datasets/dataset.py
```diff
@@ -323,15 +353,39 @@ def _create(
...
+                If set, this Dataset and all sub-resources of this Dataset will be secured by this key.
 
+                Overrides encryption_spec_key_name set in aiplatform.init.
         Returns:
             operation (Operation):
                 An object representing a long-running operation.
         """
+
+        # Use provided encryption key name or else use one from global config
+        kms_key_name = (
```
sasha-gitg, line 373: Since this code block is repeated multiple times it may be worth centralizing it in `initializer._Config`. Something like:

```python
def get_encryption_spec(self, encryption_spec_key_name: str): -> Optional(gca_encryption_spec.EncryptionSpec):
  kms_key_name = (
            encryption_spec_key_name
            or self.encryption_spec_key_name
        )
  encryption_spec = None
  if kms_key_name:
      encryption_spec = gca_encryption_spec.EncryptionSpec(
          kms_key_name=kms_key_name
      )
   return encryption_spec
```

google/cloud/aiplatform/training_jobs.py
```diff
@@ -1240,6 +1310,28 @@ def __init__(
             credentials (auth_credentials.Credentials):
                 Custom credentials to use to run call training service. Overrides
                 credentials set in aiplatform.init.
+            training_pipeline_encryption_spec_key_name (Optional[str]):
```
sasha-gitg, line 1313: I'd possibly remove `pipeline` from this name so it's consistent with the class name. If we end up changing this the XTrainingPipeline instead of XTrainignJob, we can add it back. You can leave the docstring as is.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -2662,7 +2835,9 @@ def _run(
         if model_display_name is None:
             model_display_name = self._display_name
 
-        model = gca_model.Model(display_name=model_display_name)
+        model = gca_model.Model(
+            display_name=model_display_name, encryption_spec=self.model_encryption_spec,
```
sasha-gitg, line 2839: This has a potential similar issue to where we reference this property that may pull the encryption spec key name from the global config at execution time of this potentially asynchronous method.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -99,14 +103,76 @@ def __init__(
...
         self._display_name = display_name
         self._project = project
+        self._training_pipeline_encryption_spec_key_name = (
+            training_pipeline_encryption_spec_key_name
+        )
+        self._model_encryption_spec_key_name = model_encryption_spec_key_name
         self._gca_resource = None
 
+    @property
+    def training_pipeline_encryption_spec(
```
sasha-gitg, line 141: I'm not sure of the purpose of making this a public method. It doesn't seem like it would be useful for the user to access. If all this of this logic is centralized in the global config you can lock in this spec at construction time.

sasha-gitg (changes requested):

google/cloud/aiplatform/jobs.py
```diff
@@ -420,7 +433,11 @@ def create(
                 f"type. Please choose from: {constants.BATCH_PREDICTION_OUTPUT_STORAGE_FORMATS}"
             )
 
-        gapic_batch_prediction_job = gca_bp_job.BatchPredictionJob()
+        gapic_batch_prediction_job = gca_bp_job.BatchPredictionJob(
+            encryption_spec=initializer.global_config.get_encryption_spec(
```
sasha-gitg, line 437: We may want to mutate `gapic_batch_prediction_job` instead of passing in the encryption spec into the constructor to be consistent with the rest of the composition below. Or compose it all through the constructor.

google/cloud/aiplatform/models.py
```diff
@@ -234,8 +260,22 @@ def _create(
             project=project, location=location
         )
 
+        # Use provided encryption key name or else use one from global config
+        kms_key_name = (
```
sasha-gitg, line 264: It looks like this block wasn't updated to use the global config. Also this should be set in `create` instead of this method which can be executed asynchronously.

google/cloud/aiplatform/models.py
```diff
@@ -1075,6 +1115,7 @@ def __init__(
         project: Optional[str] = None,
         location: Optional[str] = None,
         credentials: Optional[auth_credentials.Credentials] = None,
+        encryption_spec_key_name: Optional[str] = None,
```
sasha-gitg, line 1118: Why does this need encryption spec passed in?

google/cloud/aiplatform/models.py
```diff
@@ -1570,6 +1638,7 @@ def batch_predict(
         ] = None,
         labels: Optional[dict] = None,
         credentials: Optional[auth_credentials.Credentials] = None,
+        encryption_spec_key_name: Optional[str] = None,
```
sasha-gitg, line 1641: It seems like `Model.deploy` also needs support.

sasha-gitg (changes requested):

google/cloud/aiplatform/datasets/dataset.py
ivanmkc: Fixed

google/cloud/aiplatform/datasets/dataset.py
ivanmkc: Fixed

google/cloud/aiplatform/training_jobs.py
ivanmkc: Removed

google/cloud/aiplatform/datasets/dataset.py
```diff
@@ -323,15 +352,29 @@ def _create(
...
         Returns:
             operation (Operation):
                 An object representing a long-running operation.
         """
+
         gapic_dataset = gca_dataset.Dataset(
             display_name=display_name,
             metadata_schema_uri=metadata_schema_uri,
             metadata=datasource.dataset_metadata,
+            encryption_spec=initializer.global_config.get_encryption_spec(
```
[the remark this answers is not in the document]
ivanmkc: Done

google/cloud/aiplatform/initializer.py
```diff
@@ -75,6 +87,19 @@ def init(
             self._staging_bucket = staging_bucket
         if credentials:
             self._credentials = credentials
+        if encryption_spec_key_name:
+            self._encryption_spec_key_name = encryption_spec_key_name
+
+    def get_encryption_spec(
+        self, encryption_spec_key_name: str
```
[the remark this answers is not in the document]
ivanmkc: Done

google/cloud/aiplatform/jobs.py
ivanmkc: Done

google/cloud/aiplatform/initializer.py
```diff
@@ -75,6 +87,19 @@ def init(
...
+    ) -> Optional[gca_encryption_spec.EncryptionSpec]:
```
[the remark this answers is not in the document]
ivanmkc: Done

google/cloud/aiplatform/models.py
ivanmkc: Done

google/cloud/aiplatform/models.py
ivanmkc: I think I left this from an earlier version, removing.

google/cloud/aiplatform/models.py
ivanmkc: Good catch, adding it.

google/cloud/aiplatform/training_jobs.py

ivanmkc:
@sasha-gitg Fixed according to comments and then improved some docstrings.

sasha-gitg (approved):
Looks great. Thanks Ivan!

google/cloud/aiplatform/initializer.py
```diff
@@ -75,6 +87,22 @@ def init(
...
         if credentials:
             self._credentials = credentials
+        if encryption_spec_key_name:
+            self._encryption_spec_key_name = encryption_spec_key_name
+
+    def get_encryption_spec(
+        self, encryption_spec_key_name: Optional[str]
+    ) -> Optional[gca_encryption_spec.EncryptionSpec]:
+        """Creates a gca_encryption_spec.EncryptionSpec instance from the given key name.
+        If the provided key name is None, it uses the default key name if provided.
```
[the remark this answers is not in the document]
ivanmkc: I need to improve my docstring game. Thanks

vinnysenthil (approved):

PR #259, opened 2021-03-04
samples: adds text prediction samples to MBSDK

telpirion:
This adds the following samples for the MBSDK:

* aiplatform_predict_text_entity_extraction_mbsdk_sample
* aiplatform_predict_text_classification_single_label_mbsdk_sample
* aiplatform_predict_text_sentiment_analysis_mbsdk_sample

### DIFF
--- a/samples/model-builder/conftest.py
+++ b/samples/model-builder/conftest.py
@@ -1,0 +1,113 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import pytest
+from unittest import mock
+from unittest.mock import MagicMock, patch
+
+from google.cloud import aiplatform
+
+
+@pytest.fixture
+def mock_sdk_init():
+    with patch.object(aiplatform, "init") as mock:
+        yield mock
+
+
+# ----------------------------------------------------------------------------
+# Dataset Fixtures
+# ----------------------------------------------------------------------------
+
+
+@pytest.fixture
+def mock_dataset():
+    mock = MagicMock(aiplatform.datasets.Dataset)
+    yield mock
+
+
+@pytest.fixture
+def mock_new_dataset(mock_dataset):
+    with patch.object(aiplatform.datasets.Dataset, "__new__") as mock_new_dataset:
+        mock_new_dataset.return_value = mock_dataset
+        yield mock_new_dataset
+
+
+@pytest.fixture
+def mock_init_dataset(mock_new_dataset):
+    with patch.object(aiplatform.datasets.Dataset, "__init__") as mock_init_dataset:
+        mock_init_dataset.return_value = None
+        yield mock_init_dataset
+
+
+@pytest.fixture
+def mock_create_dataset():
+    with patch.object(aiplatform.datasets.Dataset, "create") as mock:
+        mock.return_value = MagicMock(aiplatform.Dataset)
+        yield mock
+
+
+@pytest.fixture
+def mock_create_image_dataset():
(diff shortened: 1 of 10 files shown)


samples/model_builder/predict_text_classification_single_label.py
```diff
@@ -0,0 +1,38 @@
...
+    endpoint = aiplatform.Endpoint(endpoint_id)
+
+    instance = TextClassificationPredictionInstance(content=content)
+
+    response = endpoint.predict(
+        instances=[instance], parameters={}
+    )
+
+    for prediction_ in response.predictions:
+        prediction = ClassificationPredictionResult.from_map(prediction_)
```
sasha-gitg, line 32: I also really like this usage of the EPCL but instead of exposing this through a sample we should consider designing this to be a native integration in the future where the prediction automatically returns the EPCL type. You can print or pprint the predictions directly.

sasha-gitg (changes requested):
Requested changes on the first sample that should apply to the remainder as well. Thank you!

samples/model_builder/predict_text_classification_single_label.py
```diff
@@ -0,0 +1,38 @@
...
+from google.cloud import aiplatform
+from google.cloud.aiplatform.v1beta1.schema.predict.instance import TextClassificationPredictionInstance
+from google.cloud.aiplatform.v1beta1.schema.predict.prediction import ClassificationPredictionResult
+
+def predict_text_classification_single_label_mbsdk(project, location, endpoint_id, content):
+    aiplatform.init(project=project, location=location)
+
+    endpoint = aiplatform.Endpoint(endpoint_id)
+
+    instance = TextClassificationPredictionInstance(content=content)
```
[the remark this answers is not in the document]
telpirion: Ack.

EPCL == GAPIC for the purposes of customers. The EPCL should be our default technique for demonstrating how to work with the library. They allow customers to inspect objects using their IDEs. A list of dictionaries doesn't do that for them.

samples/model_builder/predict_text_classification_single_label.py
```diff
@@ -0,0 +1,38 @@
...
+
+    response = endpoint.predict(
+        instances=[instance], parameters={}
+    )
+
+    for prediction_ in response.predictions:
+        prediction = ClassificationPredictionResult.from_map(prediction_)
+        print(f'Predictions:\n\n')
+        for x in range(0, len(prediction.ids)):
```
dizcology, line 34: here instead of `range` i would suggest either `for i, prediction_id in enumerate(prediction.ids):` or `for prediction_id, display_name, confidence in zip(prediction.ids, prediction.display_names, prediction.confidences):`

samples/model_builder/predict_text_classification_single_label.py
```diff
@@ -0,0 +1,38 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+#[START aiplatform_predict_text_classification_single_label_mbsdk_sample]
+from google.cloud import aiplatform
+from google.cloud.aiplatform.v1beta1.schema.predict.instance import TextClassificationPredictionInstance
+from google.cloud.aiplatform.v1beta1.schema.predict.prediction import ClassificationPredictionResult
+
+def predict_text_classification_single_label_mbsdk(project, location, endpoint_id, content):
```
dizcology, line 20: i think the sample function name and region tag should not say `mbsdk`, and just `aiplatform_sdk_..._sample`.

telpirion:
> Requested changes on the first sample that should apply to the remainder as well. Thank you!

The PR was in **Draft** form; will resend for review when done.

samples/model_builder/predict_text_classification_single_label.py
telpirion: Done

samples/model_builder/predict_text_classification_single_label.py
telpirion: Done

samples/model_builder/predict_text_classification_single_label.py
```diff
@@ -0,0 +1,38 @@
...
+    aiplatform.init(project=project, location=location)
+
+    endpoint = aiplatform.Endpoint(endpoint_id)
+
+    instance = TextClassificationPredictionInstance(content=content)
+
+    response = endpoint.predict(
+        instances=[instance], parameters={}
```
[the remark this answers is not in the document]
telpirion: Done

PR #262, opened 2021-03-15
chore: Modify integration tests to support VPC-SC environment

vinnysenthil:
Fixes part of [b/178500935](http://b/178500935) and unblocks [cl/362114575](http://cl/362114575) 🦕

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -123,10 +123,6 @@ def system(session):
     if not os.environ.get("GOOGLE_APPLICATION_CREDENTIALS", ""):
         session.skip("Credentials must be set via environment variable")
 
-    # Set test artifact bucket env var for system tests
-    if not os.environ.get("GCLOUD_TEST_SAMPLES_BUCKET"):
-        session.env["GCLOUD_TEST_SAMPLES_BUCKET"] = "cloud-samples-data-us-central1"
-
     system_test_exists = os.path.exists(system_test_path)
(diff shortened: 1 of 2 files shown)


noxfile.py
```diff
@@ -123,6 +123,10 @@ def system(session):
     if not os.environ.get("GOOGLE_APPLICATION_CREDENTIALS", ""):
         session.skip("Credentials must be set via environment variable")
 
+    # Set test artifact bucket env var for system tests
```
busunkim96, line 126: Synthtool will attempt to overwrite these lines since the `noxfile.py` is [templated](https://github.com/googleapis/synthtool/blob/master/synthtool/gcp/templates/python_library/noxfile.py.j2).

Options:
- Add a `s.replace(...)` in `synth.py` that adds in these lines ([example](https://github.com/googleapis/python-aiplatform/blob/116a29b1efcebb15bad14c3c36d3591c09ef10be/synth.py#L174))
- Move the env var setting to the top of `tests/system/aiplatform/test_dataset.py`

busunkim96 (approved):

sasha-gitg (approved):

noxfile.py
vinnysenthil: Thanks for catching this, taking option 2 and moving to `tests/system/aiplatform/test_dataset.py`

PR #263, opened 2021-03-17
feat: add support for multiple client versions

morgandu:
- [x] add **compat** module

> - [x] centralize **types** and **servcies** (client) in **compat** for _v1_ and _v1beta1_, set default to _v1beta1_
> - [x] update all types in google.cloud.aiplatform modules to point to **compat.types**
> - [x] update all services (client) in google.cloud.aiplatform modules to point to **compat.services**

- [x] In utils.py

> - [x] change the **WrappedClient** to an abstract **ClientWithOverride** class
> - [x] add **(Service)ClientWithOverride** for each AiPlatformService to predefine __default_version_ and __version_map_
> - [x] add a TypeVar **AiPlatformServiceClientWithOverride**, and update TypeVar AiPlatformServiceClient
- [x] For each AiPlatformResourceNoun class, change **client_class** to **(Service)ClientWithOverride**
- [x] add _client version selection logic_ in **BatchPredictionJob**, **Endpoint**, and **Model** class
- [x] update existing tests to make sure all tests pass
- [x] add unit tests to make sure **AiPlatformServiceClientWithOverride** works for _v1beta1_ and  _v1_

Fixes #<https://b.corp.google.com/issues/180711319> 🦕


google/cloud/aiplatform/compat/__init__.py
```diff
@@ -0,0 +1,24 @@
+# -*- coding: utf-8 -*-
+
+# Copyright 2020 Google LLC
```
sasha-gitg, line 3: 2021

google/cloud/aiplatform/compat/services/__init__.py
```diff
@@ -0,0 +1,44 @@
...
+  client as prediction_service_client,
+)
+
+__all__ = (
+    dataset_service_client,
+    endpoint_service_client,
+    job_service_client,
+    model_service_client,
+    pipeline_service_client,
+    prediction_service_client,
```
sasha-gitg, line 43: I like this idea. It seems like we should abstract away the client version handling away from the resource noun implementation as there's only one client. We have a few options there:

1. Is include v1beta1 and v1 clients in there which may defeat the purpose of this module
2. The other is wrap both clients and forward the request to the correct client. I created this quick POC implementation here: https://github.com/sasha-gitg/python-aiplatform/commit/93ab9766e11eac82e36498fad9923fc64f2d5baa

That way each resource noun has a single client without the need to manage versions. ie:

```python
client = prediction_service_client()
client.predict(...)  # goes to v1
client.explain(...)  # goes to v1beta1
```

google/cloud/aiplatform/models.py
```diff
@@ -961,8 +995,12 @@ def predict(self, instances: List, parameters: Optional[Dict] = None) -> Predict
 
         """
         self.wait()
-
-        prediction_response = self._prediction_client.predict(
+        self._prediction_api_client = self._instantiate_prediction_client(
```
sasha-gitg, line 998: Similar to the above comment about treating the `ClientWithOverride` directly as a client we wouldn't need to instantiate the client we can just directly call predict.

google/cloud/aiplatform/models.py
```diff
@@ -1013,8 +1051,12 @@ def explain(
             prediction: Prediction with returned predictions, explanations and Model Id.
         """
         self.wait()
-
-        explain_response = self._prediction_client.explain(
+        self._prediction_api_client = self._instantiate_prediction_client(
```
sasha-gitg, line 1054: And here this can be called directly:
```python
explain_response = self._prediction_client.select_version(compat.V1BETA1).explain(...)
```

sasha-gitg (changes requested):
Left some comments. Let me know if you have questions. Thanks!

google/cloud/aiplatform/compat/compat.py
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+  def default_version(self):
+    return self._clients[self._default_client_version]
+
+  def select_version(self, version: str):
+    return self._clients[version]
+
+class DatasetClientWithOverride(ClientWithOverride):
+  _default_client_version = V1BETA1
```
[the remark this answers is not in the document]
morgandu: I would vote for change default to V1 in another PR, and this PR to focus on working V1BETA1 based on the ClientWithOverride

google/cloud/aiplatform/utils.py
```diff
@@ -296,20 +306,29 @@ def extract_bucket_and_prefix_from_gcs_path(gcs_path: str) -> Tuple[str, Optiona
...
+    def _default_version(self) -> str:
+        pass
+
+    @property
+    @classmethod
+    @abc.abstractmethod
+    def _version_map(self) -> Tuple:
+        pass
 
     def __init__(
```
sasha-gitg, line 324: We should provide the option to not call through the temporary client for prediction because it is latency sensitive.  The lines for checking were removed from the `_Config.create_client`. 

I think this can done in a few ways. One option is to nest the WrappedClient class in this class like this and construct the client at instantiation:

```python
class ClientWithOverride:
  class WrappedClient:
    """Wrapper class for client that creates client at API invocation time."""

    def __init__(
        self,
        client_class: Type[AiPlatformServiceClient],
        client_options: client_options.ClientOptions,
        client_info: gapic_v1.client_info.ClientInfo,
    ):
        self._client_class = client_class
        self._client_options = client_options
        self._client_info = client_info

    def __getattr__(self, name: str) -> Any:
        """Instantiates client and returns attribute of the client."""
        temporary_client = self._client_class(
            credentials=self._credentials,
            client_options=self._client_options,
            client_info=self._client_info,
        )
        return getattr(temporary_client, name)

  @abstract class property
  _is_temporary: Bool

  def __init__(
      self, credentials, client_options,  client_info):
      self._clients = {
          version: WrappedClient(client, credentials, client_options, client_info) if self._is_temporary else client(...)
          for version, client in self._version_map
      }

  def __getattr__(self, name: str) -> Any:
      """Instantiates client and returns attribute of the client."""
      return getattr(self._clients[self._default_version], name)

  def select_version(self, version: str) -> AiPlatformServiceClient:
      return self._clients[version]
```

The main goal is to avoid the additional check to see if we should use the temporary client at every API invocation.

google/cloud/aiplatform/utils.py
```diff
@@ -318,20 +337,79 @@ def __init__(
...
         )
-        return getattr(temporary_client, name)
+        return temporary_client
+
+    def __getattr__(self, name: str) -> Any:
+        """Instantiates client and returns attribute of the client."""
+        return getattr(self._temporary_client(self._default_version), name)
+
+    def select_version(self, version: str) -> AiPlatformServiceClient:
+        return self._temporary_client(version)
```
sasha-gitg, line 360: One potential issue with this approach is that we are not enforcing that the client is temporary like we are through the `__getattr__` path.

sasha-gitg (changes requested):

google/cloud/aiplatform/jobs.py
```diff
@@ -533,7 +542,8 @@ def _create(
         cls,
         api_client: job_service_client.JobServiceClient,
         parent: str,
-        batch_prediction_job: gca_bp_job.BatchPredictionJob,
+        batch_prediction_job: gca_bp_job_default.BatchPredictionJob,
```
sasha-gitg, line 545: This seems like the type should be a Union of v1 and v1beta1 BatchPredictionJob.

google/cloud/aiplatform/models.py
```diff
@@ -354,6 +353,35 @@ def _unallocate_traffic(
 
         return new_traffic_split
 
+    @staticmethod
+    def _validate_explanation_args(
+        explanation_metadata: Optional["aiplatform.explain.ExplanationMetadata"] = None,
```
sasha-gitg, line 358: It seems like the forward reference is not required here. Can this be:
`Optional[explain.ExplanationMetadata]`.

Same for the line below.

google/cloud/aiplatform/models.py
```diff
@@ -748,20 +776,24 @@ def _deploy_call(
             raise ValueError(
                 "Both `accelerator_type` and `accelerator_count` should be specified or None."
             )
-        if bool(explanation_metadata) != bool(explanation_parameters):
-            raise ValueError(
-                "Both `explanation_metadata` and `explanation_parameters` should be specified or None."
-            )
+        generate_explanation = cls._validate_explanation_args(
```
sasha-gitg, line 779: I think this validation should be moved to _validate_deployment_args so it fails early.

google/cloud/aiplatform/models.py
```diff
@@ -1352,8 +1393,8 @@ def upload(
             parent=initializer.global_config.common_location_path(project, location),
             model=managed_model,
         )
-
         managed_model = lro.result()
+        print(f"MANAGED MODEL PRINT: {managed_model.model}")
```
sasha-gitg, line 1397: Debug statement

sasha-gitg (changes requested):
This is looking great. I have a few more comments and questions. Thanks!

PR #265, opened 2021-03-19
feat: Add initial Model Builder SDK samples

vinnysenthil:
Adds an initial set of Model Builder SDK samples and their tests, along with a `conftest.py` file to hold mocks.

Here are the samples added:
- `init_sample`
- `create_and_import_dataset_image_sample`
- `create_batch_prediction_job_sample`
- `create_training_pipeline_image_classification_sample`


samples/model-builder/create_and_import_dataset_image_sample.py
```diff
@@ -0,0 +1,34 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from google.cloud import aiplatform
+
+#  [START aiplatform_sdk_create_and_import_dataset_image_sample]
+def create_and_import_dataset_image_sample(
+    project: str, location: str, display_name: str, src_uris: list
```
sasha-gitg, line 19:
```suggestion
    project: str, location: str, display_name: str, src_uris: List[str]
```

Not sure if this is the standard for samples.

samples/model-builder/create_batch_prediction_job_sample.py
```diff
@@ -0,0 +1,45 @@
...
+    sync: bool = True,
+):
+    aiplatform.init(project=project, location=location)
+
+    batch_prediction_job = aiplatform.BatchPredictionJob.create(
+        model_name=model_resource_name,
+        job_display_name=job_display_name,
+        gcs_source=gcs_source,
+        gcs_destination_prefix=gcs_destination,
+        sync=sync,
```
sasha-gitg, line 36: We should call `batch_prediction_job.wait()` if we allow the user to call this async. I'm not sure if it adds value here since we will have to wait to be able to print at least some of the information below.

Also, I don't have an opinion on which is a better option but I think we should try to keep usage consistent between samples. ie. add to the Dataset.create example above

samples/model-builder/create_batch_prediction_job_sample.py
```diff
@@ -0,0 +1,45 @@
...
+    location: str,
+    model_resource_name: str,
+    job_display_name: str,
+    gcs_source: Union[str, Sequence[str]],
+    gcs_destination: str,
+    sync: bool = True,
+):
+    aiplatform.init(project=project, location=location)
+
+    batch_prediction_job = aiplatform.BatchPredictionJob.create(
```
sasha-gitg, line 31: Slight preference here is to call this through `Model`.

samples/model-builder/create_training_pipeline_image_classification_sample.py
```diff
@@ -0,0 +1,55 @@
...
+    test_fraction_split: float = 0.1,
+    budget_milli_node_hours: int = 8000,
+    disable_early_stopping: bool = False,
+    sync: bool = True,
+):
+    aiplatform.init(project=project)
+
+    job = aiplatform.AutoMLImageTrainingJob(display_name=display_name)
+
+    my_image_ds = aiplatform.Dataset(dataset_id)
```
sasha-gitg, line 34: Since this is expecting dataset_id, we also need location or this could fail if the dataset is not in the default location ("us-central1")

samples/model-builder/create_and_import_dataset_image_sample_test.py
```diff
@@ -0,0 +1,28 @@
...
+
+def test_create_and_import_dataset_image_sample(
+    mock_sdk_init, mock_create_image_dataset
+):
+
+    create_and_import_dataset_image_sample.create_and_import_dataset_image_sample(
+        project="abc", location="us-central1", src_uris=["123"], display_name="test_dn"
+    )
+
+    mock_sdk_init.assert_called_once()
```
sasha-gitg, line 27: Possibly use `assert_called_once_with` throughout to be more strict.

samples/model-builder/conftest.py
```diff
@@ -0,0 +1,81 @@
...
+# TrainingJob Fixtures
+# ----------------------------------------------------------------------------
+
+
+@pytest.fixture
+def mock_init_automl_image_training_job():
+    with patch.object(
+        aiplatform.training_jobs.AutoMLImageTrainingJob, "__init__"
+    ) as mock:
+        mock.return_value = None
```
sasha-gitg, line 63: This seems either superfluous or possibly will cause an error because we would call `run` on `None`.

sasha-gitg (approved):
LGTM! Added some comments.

samples/model-builder/create_and_import_dataset_image_sample.py
vinnysenthil: Updated to `Union[str, List[str]]`

samples/model-builder/create_batch_prediction_job_sample.py
vinnysenthil: Good catch, swapping this

samples/model-builder/create_batch_prediction_job_sample.py
vinnysenthil: - Adding `wait()` call to this sample
- Adding `sync=sync` + `ds.wait()` to `create_and_import_dataset_image_sample`

samples/model-builder/create_training_pipeline_image_classification_sample.py
vinnysenthil: Added location param to be passed to init

samples/model-builder/conftest.py
vinnysenthil: Not returning `None` when patching the `__init__` caused the following exception:
```
TypeError: __init__() should return None, not 'MagicMock'
```

The downstream calls would also be patched so it shouldn't cause an error. Is this alright or should we look for alternative approaches?

sasha-gitg: Ah, right. I was wrong on that. Leave it as is. That makes sense, `__new__` returns the instance, not `__init__`.

dizcology (approved):

samples/model-builder/create_and_import_dataset_image_sample_test.py
vinnysenthil: Updated to use `assert_called_once_with` everywhere

vinnysenthil:
PTAL when you get a chance @busunkim96 @tmatsuo — thanks!

busunkim96 (commented):
I recommend turning on Autospeccing (https://docs.python.org/3/library/unittest.mock.html#autospeccing) where possible. It makes it easier to catch incorrect method calls.

The `conftest.py` approach seems good for sharing code needed by the tests. For very simple mocks IMO it is easier for folks to follow when the mock is right next to the test using the patch() decorator syntax. ([example in google-auth](https://github.com/googleapis/google-auth-library-python/blob/333cb765b52028329ec3ca04edf32c5764b1db68/tests_async/test__default_async.py#L310-L322))

PR #267, opened 2021-03-20
fix: Fixed exception when the python executable path has a space in it

ivanmkc:
In lines (line 768-778) in training_jobs.py:
```
# Run setup.py to create the source distribution.
setup_cmd = self._SETUP_PY_SOURCE_DISTRIBUTION_CMD.format(
    python_executable=_get_python_executable()
).split()

p = subprocess.Popen(
    args=setup_cmd,
    cwd=trainer_root_path,
    stdout=subprocess.PIPE,
    stderr=subprocess.PIPE,
)
```

If your python executable path has spaces in it, you get an exception:

`FileNotFoundError: [Errno 2] No such file or directory: '/Volumes/GoogleDrive/Shared'`

which is the part of your executable path before the first space.

This fixes avoids combining the executable and args together, which fixes the problem.

### DIFF
--- a/google/cloud/aiplatform/training_jobs.py
+++ b/google/cloud/aiplatform/training_jobs.py
@@ -832,9 +832,7 @@ setup(
     description='My training application.'
 )"""
 
-    _SETUP_PY_SOURCE_DISTRIBUTION_CMD = (
-        "{python_executable} setup.py sdist --formats=gztar"
-    )
+    _SETUP_PY_SOURCE_DISTRIBUTION_CMD = "setup.py sdist --formats=gztar"
 
     # Module name that can be executed during training. ie. python -m
     module_name = f"{_ROOT_MODULE}.{_TASK_MODULE_NAME}"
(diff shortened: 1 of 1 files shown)


google/cloud/aiplatform/training_jobs.py
```diff
@@ -908,9 +908,9 @@ def make_package(self, package_directory: str) -> str:
         shutil.copy(self.script_path, script_out_path)
 
         # Run setup.py to create the source distribution.
-        setup_cmd = self._SETUP_PY_SOURCE_DISTRIBUTION_CMD.format(
-            python_executable=_get_python_executable()
-        ).split()
+        setup_cmd = [
```
ivanmkc, line 911: Let me know if it's more Pythonic to use append, extend or list comprehension instead.

sasha-gitg, line 911: The key word `{python_executable}` should be removed in the constant above(line 836) because it looks like it goes unused.

sasha-gitg (approved):
LGTM. One comment. Thanks!

google/cloud/aiplatform/training_jobs.py
sasha-gitg: Looks good to me.

ivanmkc: Ah yes, I was testing on a different repo and didn't copy it over.

ivanmkc: I fixed it and now addressing some test failures.

PR #292, opened 2021-04-06
fix: add v1 conversion value rule

sasha-gitg:


telpirion (commented):
(I'm asking this because I don't know ...) Does Python do pass-by-value by default? If not, do you need a new variable for the second Marshal rule object?

sasha-gitg:
> (I'm asking this because I don't know ...) Does Python do pass-by-value by default? If not, do you need a new variable for the second Marshal rule object?

Taking a quick look here: https://proto-plus-python.readthedocs.io/en/stable/_modules/proto/marshal/marshal.html

It seems that the variable is unused as the class maintains a mapping of model name to BaseMarshal with rules.

telpirion:
Okay👍

telpirion (approved):

PR #294, opened 2021-04-07
feat: Add list() method to all resource nouns

vinnysenthil:
### Summary of Changes
- Added a `_list_method` property to `AiPlatformResourceNoun` to store GAPIC method name for each noun
- Added a `create_time` and `update_time` property to `AiPlatformResourceNoun`
- Added a single `list()` method that takes four optional fields and returns a list of SDK types
    - All of the fields except `order_by` are available in every GAPIC list methods
    - Added local sorting for GAPIC list methods that do not take `order_by`
- Added 3 unit tests to check correct GAPIC calls and local sorting
- Added `aiplatform.init()` to test class setup and dropped it from some unit tests

Fixes [b/183498826](http://b/183498826) 🦕


google/cloud/aiplatform/base.py
```diff
@@ -561,6 +579,123 @@ def _sync_object_with_future_result(
...
+        return sdk_resource
+
+    # TODO(b/144545165) - Improve documentation for list filtering once available
+    @classmethod
+    def list(
+        cls,
+        filter: Optional[str] = None,
+        order_by: Optional[str] = None,
+        page_size: Optional[int] = None,
+        read_mask: Optional[field_mask.FieldMask] = None,
```
vinnysenthil, line 599: `page_size` is not respected due to the list comprehension, should we drop it for now?
Also `read_mask` has little use since the user is not getting the `_gca_resource`, think we can hide that as well.

sasha-gitg: I agree that we should remove read_mask for now and we can add it back in later.

google/cloud/aiplatform/base.py
```diff
@@ -561,6 +579,123 @@ def _sync_object_with_future_result(
...
+                Supported fields: `display_name`, `create_time`, `update_time`
+            page_size (int):
+                Optional. The standard list page size.
+            read_mask (field_mask.FieldMask):
+                Optional. Mask specifying which fields to read.
+
+        Returns:
+            Sequence[AiPlatformResourceNoun] - A list of SDK resource objects
+        """
+        _UNSUPPORTED_LIST_ORDER_BY_TYPES = (
```
sasha-gitg, line 627: Strong preference to break down this conditional logic at the subclass level instead. 

One option is to have two private list classes and pass in the filtering attributes.
```Python
class AiPlatformResourceNoun:
  def _list(..., 
                order_by: Optional[str],
                gapic_field_filter_name: Optional[str]:None,
                cls_field_filter_name: Optional[str]: None):
    ...
    cls_filter_schema = getattr(cls, cls_field_filter_schema, None) if cls_field_filter_name else set([])
    final_list = [
                self._construct_sdk_resource_from_gapic(
                    gapic_resource, credentials=creds
                )
                for gapic_resource in resource_list
                if gapic_field_filter_key and getattr(gapic_resource, gapic_field_filter_key)
                in cls_filter_schema
            ]
  def _list_with_local_order(...,order_by: Optional[str]):
     li = cls._list(..., order_by=None)
     # order here
     return li
     
  def list(...):
    return cls._list(...)
     
class _TrainingJob:
   def list(...):
      return cls._list_with_local_order(
        ...,
        order,
        gapic_field_filter_name='training_task_definition',
        cls_field_filter_name='_supported_training_schemas') # could just pass in the cls attribute as well
 ```

This will be more extensible in the future and allow external teams to use the list functionality without needing to change the base class.

File: google/cloud/aiplatform/base.py [hunk 3]
```diff
@@ -561,6 +579,123 @@ def _sync_object_with_future_result(
...
+            "read_mask": read_mask,
+        }
+
+        # If list method does not offer `order_by` field, order locally
+        if order_by and issubclass(type(self), _UNSUPPORTED_LIST_ORDER_BY_TYPES):
+            order_locally = True
+        elif order_by:
+            list_request["order_by"] = order_by
+
+        resource_list = resource_list_method(request=list_request) or []
```
sasha-gitg, line 654: Why is the empty list needed here?

tests/unit/aiplatform/test_datasets.py
```diff
@@ -106,6 +106,21 @@
 # misc
 _TEST_OUTPUT_DIR = "gs://my-output-bucket"
 
+_TEST_TABULAR_DATASET_LIST = [
```
sasha-gitg, line 109: To test the filtering, I think we should add at least one item in the mock that isn't tabular.

google/cloud/aiplatform/datasets/dataset.py
```diff
@@ -40,9 +40,10 @@ class Dataset(base.AiPlatformResourceNounWithFutureManager):
     _is_client_prediction_client = False
     _resource_noun = "datasets"
     _getter_method = "get_dataset"
+    _list_method = "list_datasets"
     _delete_method = "delete_dataset"
 
-    _supported_metadata_schema_uris: Optional[Tuple[str]] = None
+    _supported_metadata_schema_uris: Optional[Tuple[str]] = ()
```
sasha-gitg, line 46: Should we change the type here as well?

google/cloud/aiplatform/base.py
```diff
@@ -561,6 +579,123 @@ def _sync_object_with_future_result(
...
+
+    # TODO(b/144545165) - Improve documentation for list filtering once available
+    @classmethod
+    def list(
+        cls,
+        filter: Optional[str] = None,
+        order_by: Optional[str] = None,
+        page_size: Optional[int] = None,
+        read_mask: Optional[field_mask.FieldMask] = None,
+    ) -> Sequence[AiPlatformResourceNoun]:
```
sasha-gitg, line 600: It's probably OK to specify this to a List since we always return a list.

google/cloud/aiplatform/base.py
vinnysenthil: I ran into a few rare instances where the service returned a `None` instead of an empty list, this just serves as a fail safe against that.

google/cloud/aiplatform/datasets/dataset.py
vinnysenthil: Dropping `Optional` portion of type

google/cloud/aiplatform/base.py
```diff
@@ -343,6 +351,16 @@ def display_name(self) -> str:
         """Display name of this resource."""
         return self._gca_resource.display_name
 
+    @property
+    def create_time(self) -> datetime.datetime:
+        """Time this resource was created."""
+        return self._gca_resource.create_time
+
+    @property
+    def update_time(self) -> datetime.datetime:
```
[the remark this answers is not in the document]
vinnysenthil: Good point, adding the call

google/cloud/aiplatform/base.py
vinnysenthil: Dropping both fields and and filed [b/184910159](http://b/184910159) to eventually expose `page_size`

google/cloud/aiplatform/base.py
```diff
@@ -561,6 +579,123 @@ def _sync_object_with_future_result(
             if value:
                 setattr(self, attribute, value)
 
+    def _construct_sdk_resource_from_gapic(
+        self,
+        gapic_resource: proto.Message,
+        credentials: Optional[auth_credentials.Credentials] = None,
+    ) -> AiPlatformResourceNoun:
+        """Given a GAPIC object, return the SDK representation."""
+        sdk_resource = self._empty_constructor(credentials=credentials)
```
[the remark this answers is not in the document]
vinnysenthil: Added `location` and `project` parameters to `_construct_sdk_resource_from_gapic()` and now forward those values to `self._empty_constructor`

google/cloud/aiplatform/base.py
```diff
@@ -561,6 +579,123 @@ def _sync_object_with_future_result(
...
+        sdk_resource._gca_resource = gapic_resource
+        return sdk_resource
+
+    # TODO(b/144545165) - Improve documentation for list filtering once available
+    @classmethod
+    def list(
```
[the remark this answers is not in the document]
vinnysenthil: Done.  Also now forward the three args to the first call to `self._empty_constructor()` and to any calls to `_sync_object_with_future_result()`

tests/unit/aiplatform/test_datasets.py
vinnysenthil: Great call out, added a text and image dataset into the mix.

google/cloud/aiplatform/base.py
vinnysenthil: Thanks for putting together this example! Adopting that option but will have the private `_list` method take a single `cls_filter` Callable that takes a `gca_resource` returns a bool that decides whether to include or exclude a given GAPIC object. Lmk if you have any concerns with that approach.

vinnysenthil:
@sasha-gitg I've made changes that reflect [the structure you suggested](https://github.com/googleapis/python-aiplatform/pull/294#discussion_r609938470), PTAL at [the commit](https://github.com/googleapis/python-aiplatform/pull/294/commits/cabef1b2a743cd8476aec5b3128d688ff62871f8) and lmk if it looks good. If so, I'll add doc strings to the private methods and a bit more test coverage 🙂

google/cloud/aiplatform/base.py
```diff
@@ -561,6 +578,141 @@ def _sync_object_with_future_result(
...
                 setattr(self, attribute, value)
 
+    def _construct_sdk_resource_from_gapic(
+        self,
+        gapic_resource: proto.Message,
+        project: Optional[str] = None,
+        location: Optional[str] = None,
+        credentials: Optional[auth_credentials.Credentials] = None,
+    ) -> AiPlatformResourceNoun:
+        """Given a GAPIC object, return the SDK representation."""
```
sasha-gitg, line 588: docstring

google/cloud/aiplatform/base.py
```diff
@@ -561,6 +578,141 @@ def _sync_object_with_future_result(
...
+            project=project, location=location, credentials=credentials
+        )
+
+        # Fetch credentials once and re-use for all `_empty_constructor()` calls
+        creds = initializer.global_config.credentials
+
+        resource_list_method = getattr(self.api_client, self._list_method)
+
+        list_request = {
+            "parent": initializer.global_config.common_location_path(),
```
sasha-gitg, line 618: Project and location need to be passed here if given.

google/cloud/aiplatform/base.py
```diff
@@ -561,6 +578,141 @@ def _sync_object_with_future_result(
...
+    @classmethod
+    def _list(
+        cls,
+        cls_filter: Callable[[proto.Message], bool] = lambda _: True,
+        filter: Optional[str] = None,
+        order_by: Optional[str] = None,
+        project: Optional[str] = None,
+        location: Optional[str] = None,
+        credentials: Optional[auth_credentials.Credentials] = None,
+    ) -> List[AiPlatformResourceNoun]:
```
sasha-gitg, line 606: docstring

google/cloud/aiplatform/base.py
```diff
@@ -561,6 +578,141 @@ def _sync_object_with_future_result(
...
+    def _list_with_local_order(
+        cls,
+        cls_filter: Callable[[proto.Message], bool] = lambda _: True,
+        filter: Optional[str] = None,
+        order_by: Optional[str] = None,
+        project: Optional[str] = None,
+        location: Optional[str] = None,
+        credentials: Optional[auth_credentials.Credentials] = None,
+    ) -> List[AiPlatformResourceNoun]:
+        """Client-side sorting when list API doesn't support `order_by`"""
```
sasha-gitg, line 645: docstring

google/cloud/aiplatform/base.py
```diff
@@ -561,6 +578,141 @@ def _sync_object_with_future_result(
...
+
+        aiplatform.Model.list(order_by="create_time desc, display_name")
+
+        Args:
+            filter (str):
+                Optional. An expression for filtering the results of the request.
+                For field names both snake_case and camelCase are supported.
+            order_by (str):
+                Optional. A comma-separated list of fields to order by, sorted in
+                ascending order. Use "desc" after a field name for descending.
```
sasha-gitg, line 692: Is something like `'display_name, desc, create_time, update_time, desc'` valid here or it only when `desc` is at the end of the list? It doesn't seem like the local logic handles that case.

sasha-gitg (approved):
LGTM! Thanks for this one. I have one main question on ordering.

google/cloud/aiplatform/base.py
vinnysenthil: Done, forwards the project and location args to `common_location_path()`

google/cloud/aiplatform/base.py
vinnysenthil: Added

google/cloud/aiplatform/base.py
vinnysenthil: Done

google/cloud/aiplatform/base.py
vinnysenthil: Done

google/cloud/aiplatform/base.py
vinnysenthil: `desc` is only considered valid by the service when it follows one of the three fields and remains inside the comma delim. Manually tried other combinations, I think the local logic handles all cases that are valid service-side.

```python
In [5]: endpoint_list = aiplatform.Model.list(order_by='display_name, desc, create_time, update_time, desc')
   ...: print_list(endpoint_list)
---------------------------------------------------------------------------
... long stack trace here ...

~/mbsdk/vinny-aip/venv/lib/python3.7/site-packages/google/api_core/grpc_helpers.py in error_remapped_callable(*args, **kwargs)
     57             return callable_(*args, **kwargs)
     58         except grpc.RpcError as exc:
---> 59             six.raise_from(exceptions.from_grpc_error(exc), exc)
     60 
     61     return error_remapped_callable

~/mbsdk/vinny-aip/venv/lib/python3.7/site-packages/six.py in raise_from(value, from_value)

InvalidArgument: 400 Provided `order_by` field has invalid format.
```

morgandu (approved):

PR #296, opened 2021-04-07
feat: Make aiplatform.Dataset private

vinnysenthil:
- Removes all use of `aiplatform.Dataset` 
  - Replaces internal references with `aiplatform._Dataset`
  - Replaces public references with specific dataset sub-class or TypeVar `datasets.SupportedDatasets`

## TODO after merging
- [x] @sasha-gitg PTAL [at this list of updates](https://docs.google.com/document/d/1mKqldyCZaLgtYrgxpd6F3TJ35Ap3l6MhOxyiWTmZ4NQ/edit) to external notebooks and Colabs to no longer reference `aiplatform.Dataset` 

Fixes [b/184156060](http://b/184156060) 🦕


google/cloud/aiplatform/training_jobs.py
```diff
@@ -1701,7 +1701,7 @@ def __init__(
     # TODO(b/172368070) add timestamp split, training_pipeline.TimestampSplit
     def run(
         self,
-        dataset: Optional[datasets.Dataset] = None,
+        dataset: Optional[datasets.SupportedDatasets] = None,
```
sasha-gitg, line 1704: I like using the TypeVar for our internal implementations but using the private `_Dataset` class is also sufficient for that case. For public APIs we may need a different approach. If this doesn't expand in the docs it's probably more informative to put the explicit unions `Union[datasets.TabularDataset, datasets.ImageDataset, datasets.VideoDataset, datasets.TextDataset]`. This would just be for custom training jobs.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -2585,7 +2585,7 @@ def __init__(
 
     def run(
         self,
-        dataset: datasets.Dataset,
+        dataset: datasets.SupportedDatasets,
```
sasha-gitg, line 2588: This should be scoped down to `TabularDataset`.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -2986,7 +2986,7 @@ def __init__(
...
```
sasha-gitg, line 2989: This should be scoped down to `ImageDatasets`.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -3797,7 +3797,7 @@ def __init__(
...
```
sasha-gitg, line 3800: Scope down to `VideoDatasets`.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -4063,7 +4063,7 @@ def __init__(
...
```
sasha-gitg, line 4066: Scope down to `TextDataset`

sasha-gitg (changes requested):

google/cloud/aiplatform/training_jobs.py
vinnysenthil: Removed `SupportedDatasets` and defaulted to `_Dataset` in private methods and the explicit unions in the three custom training job classes.

sasha-gitg (approved):
LGTM! Thanks!

morgandu (approved):

busunkim96 (approved):

PR #297, opened 2021-04-08
Feat: parse project location when passed full resource name to get apis

morgandu:
- [x] add __get_and_validate_project_location_ for resource_name in **AiPlatformResourceNoun** in _base.py_
- [x] pass in resource_name for AiPlatformResourceNoun subclasses.

_unit tests:_
- [x] add tests through **dataset.Dataset**
- [x] add tests through **training_jobs.CustomTrainingJob.get**

Fixes <[b/180923329](https://b.corp.google.com/issues/180923329)>

### DIFF
--- a/google/cloud/aiplatform/base.py
+++ b/google/cloud/aiplatform/base.py
@@ -20,7 +20,7 @@ from concurrent import futures
 import functools
 import inspect
 import threading
-from typing import Any, Callable, Dict, Optional, Sequence, Type, Union
+from typing import Any, Callable, Dict, Optional, Sequence, Tuple, Type, Union
 
 import proto
 
@@ -266,6 +266,7 @@ class AiPlatformResourceNoun(metaclass=abc.ABCMeta):
         project: Optional[str] = None,
         location: Optional[str] = None,
         credentials: Optional[auth_credentials.Credentials] = None,
+        resource_name: Optional[str] = None,
     ):
         """Initializes class with project, location, and api_client.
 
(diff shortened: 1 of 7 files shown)


google/cloud/aiplatform/base.py
```diff
@@ -320,6 +320,16 @@ def _get_gca_resource(self, resource_name: str) -> proto.Message:
             project=self.project,
             location=self.location,
         )
+        (
```
sasha-gitg, line 323: I think this check needs to happen earlier in the stack than `_get_gca_resource` because the project/location of the object get set before this:

https://github.com/googleapis/python-aiplatform/blob/dev/google/cloud/aiplatform/datasets/dataset.py#L73
-> 
https://github.com/googleapis/python-aiplatform/blob/dev/google/cloud/aiplatform/base.py#L279

In this current implementation the location attribute and project attribute on the object may match the resource and client.

I think the most direct way to support this is to to add something like this to the AIPlatformResourceNoun base class: 

```Python
def _get_and_validate_project_location(cls, project, location, resource_name_or_id):
  if not project and not location:   # no issue
    return project, location
  
  fields = extract_fields_from_resource_name(resource_name_or_id, cls._resource_noun)
  if not fields:  # not a resource name
        return project, location
 
  # skip validating project because it's not needed and would require we correlate project no. with project ids 

  if location and field.location != location:  # this is a problem
      raise runtimeerror
  
  return fields.project, fields.location
```

And update the constructor with the something like:

```Python
def __init__(self, project: Optional, location: Optional, creds: Optional, resource_name_or_id: Optional[str]):
    if resource_name_or_id:
      project, location = self._get_and_validate_project_location(project, location, resource_name_or_id)
    ...
```

And update subclass call sites to pass in the resource noun in the constructor.

sasha-gitg (changes requested):

sasha-gitg (approved):
LGTM! Thanks Morgan!

vinnysenthil (approved):

PR #298, opened 2021-04-08
Adding Metadata Context resource

nachocano:
Adding metadata context class as well as its create call. Added some basic UTs.
Modified utils to support non-integer IDs and nested resource names (both needed for metadata). 

This is based off https://github.com/googleapis/python-aiplatform/pull/293, so it contains extra information.

Fixes b/170334193: support for resource names with non-integer IDs
Fixes b/170334098: support nested resource names. 

TODO in a follow up: our other resources, Artifacts and Executions, look pretty much the same as Context, might try to dedupe code using an abstract base class


nachocano:
Closing in favor of https://github.com/googleapis/python-aiplatform/pull/299

PR #303, opened 2021-04-09
feat: adding MetadataService API and new classes for Context, Artifact and Execution

nachocano:
Add new Context, Artifact and Execution classes. Using a base abstract class to avoid duplicating code.
Added basic UTs for the three classes.

Adding a MetadataService object and publishing our APIs, namely:

```
aiplatform.set_experiment (can be unexposed if not required)
aiplatform.set_run (can be unexposed if not required)
aiplatform.log_param (can be unexposed if not required)
aiplatform.log_params
aiplatform.log_metric (can be unexposed if not required)
aiplatform.log_metrics
aiplatform.get_experiment (unimplemented)
aiplatform.get_run (unimplemented, might not be needed)

```

I was able to test the APIs from a notebook and they work. See https://github.com/googleapis/python-aiplatform/pull/303#issuecomment-817497907

Still pending (for follow up PRs):
- Need to add support for connecting contexts with executions and artifacts.
- Add UTs to get and update calls as well as metadata.py with the "service" logic.
- Implement get_experiment, (get_run?)


google/cloud/aiplatform/metadata/metadata.py
```diff
@@ -0,0 +1,126 @@
...
+            _MetadataStore()
+        except exceptions.NotFound:
+            logging.info(
+                f"Creating a default MetadataStore for experiment {experiment_name}"
+            )
+            _MetadataStore.create()
+
+        try:
+            context = _Context(context_name=experiment_name)
+        except exceptions.NotFound:
```
nachocano, line 48: @sasha-gitg I cannot catch this exception. I'm getting something like:

```---------------------------------------------------------------------------
_InactiveRpcError                         Traceback (most recent call last)
~/Library/Python/3.6/lib/python/site-packages/google/api_core/grpc_helpers.py in error_remapped_callable(*args, **kwargs)
     72         try:
---> 73             return callable_(*args, **kwargs)
     74         except grpc.RpcError as exc:

~/Library/Python/3.6/lib/python/site-packages/grpc/_channel.py in __call__(self, request, timeout, metadata, credentials, wait_for_ready, compression)
    922                                       wait_for_ready, compression)
--> 923         return _end_unary_response_blocking(state, call, False, None)
    924 

~/Library/Python/3.6/lib/python/site-packages/grpc/_channel.py in _end_unary_response_blocking(state, call, with_call, deadline)
    825     else:
--> 826         raise _InactiveRpcError(state)
    827 

_InactiveRpcError: <_InactiveRpcError of RPC that terminated with:
	status = StatusCode.NOT_FOUND
	details = "Resource not found.; GetContext is unable to read context resource from auxiliary table with name: projects/354090969940/locations/us-central1/metadataStores/default/contexts/experiment"
	debug_error_string = "{"created":"@1617954685.100137000","description":"Error received from peer ipv6:[2607:f8b0:400a:807::200a]:443","file":"src/core/lib/surface/call.cc","file_line":1068,"grpc_message":"Resource not found.; GetContext is unable to read context resource from auxiliary table with name: projects/354090969940/locations/us-central1/metadataStores/default/contexts/experiment","grpc_status":5}"
```

I believe this is being thrown from `getattr(self.api_client, self._getter_method)` in resources.py.
I used to have something similar before for the create call, and then end up calling `.result()` and it worked. Is it because it is returning a future? How can I make it sync?

nachocano: Ohh ok, indeed, calling `.result()` works...

nachocano: hmm, not really, still getting this exception

nachocano: OK, figured it out... resolving...

google/cloud/aiplatform/initializer.py
```diff
@@ -88,8 +90,13 @@ def init(
             utils.validate_region(location)
             self._location = location
         if experiment:
-            logging.warning("Experiments currently not supported.")
-            self._experiment = experiment
+            metadata.metadata_service.set_experiment(experiment)
+        if run:
+            if not experiment:
+                raise ValueError(
+                    "No experiment found. Provide and experiment for this run."
```
sasha-gitg, line 97: I would add a way to correct here: ie: ```aiplatform.init(experiment='my-experiment')```

sasha-gitg, line 97:
```suggestion
                    "No experiment set. Provide an experiment for this run."
```

google/cloud/aiplatform/initializer.py
```diff
@@ -44,7 +45,6 @@ class _Config:
 
     def __init__(self):
```
sasha-gitg, line 46: This class most likely needs a helper method `_reset_metadata_tracking` that sets `experiment` and `run` to `None`. This should be called if `project` or `location` is changed.

google/cloud/aiplatform/metadata/artifact.py
```diff
@@ -0,0 +1,285 @@
...
+
+    @classmethod
+    def create(
+        cls,
+        artifact_id: str,
+        schema_title: str,
+        display_name: Optional[str] = None,
+        schema_version: Optional[str] = None,
+        description: Optional[str] = None,
+        metadata: Optional[Dict] = {},
```
sasha-gitg, line 80: This should be set to `None` since the default here is mutable.

google/cloud/aiplatform/metadata/artifact.py
```diff
@@ -0,0 +1,285 @@
...
+    ) -> proto.Message:
+        return client.create_artifact(
+            parent=parent, artifact=resource, artifact_id=resource_id,
+        )
+
+    @classmethod
+    def update(
+        cls,
+        artifact_id: str,
+        metadata: Optional[Dict] = {},
```
sasha-gitg, line 167: `None` here as well.

google/cloud/aiplatform/metadata/artifact.py
```diff
@@ -0,0 +1,285 @@
...
+        parent: str,
+        resource: proto.Message,
+        resource_id: str,
+    ) -> proto.Message:
+        return client.create_artifact(
+            parent=parent, artifact=resource, artifact_id=resource_id,
+        )
+
+    @classmethod
+    def update(
```
sasha-gitg, line 164: This seems like it should be an instance method and should act on the resource of the instance.

google/cloud/aiplatform/metadata/artifact.py
```diff
@@ -0,0 +1,285 @@
...
+            resource_noun=cls._resource_noun,
+            gapic_resource=gapic_artifact,
+            metadata_store_id=metadata_store_id,
+            project=project,
+            location=location,
+            credentials=credentials,
+        )
+
+        if not resource_name:
+            raise ValueError("Error while updating artifact")
```
sasha-gitg, line 223: It's better to propagate the service error.

google/cloud/aiplatform/metadata/artifact.py
```diff
@@ -0,0 +1,285 @@
...
+            metadata_store_id=metadata_store_id,
+            project=project,
+            location=location,
+            credentials=credentials,
+        )
+
+    @classmethod
+    def create_resource(
+        cls,
+        client: utils.AiPlatformServiceClientWithOverride,
```
sasha-gitg, line 154: This should be specified to the `utils.MetadataClientWithOverride`

google/cloud/aiplatform/metadata/artifact.py
```diff
@@ -0,0 +1,285 @@
...
+        return cls(
+            artifact_name=resource_name,
+            metadata_store_id=metadata_store_id,
+            project=project,
+            location=location,
+            credentials=credentials,
+        )
+
+    @classmethod
+    def update_resource(
```
sasha-gitg, line 234: Why do we need two update methods?

google/cloud/aiplatform/initializer.py
nachocano: Done

nachocano: Done, thanks!

google/cloud/aiplatform/metadata/artifact.py
nachocano: Done, here and other places, thanks!

google/cloud/aiplatform/metadata/artifact.py
nachocano: I'm doing something similar than the create methods. As each resource needs to call its create_artifact, create_context, create_execution, I was basically having a base _Resource class that does pretty much everything, and the children have to implement the abstract create_resource (in this case update_resource) with their particular implementations (update_artifact, update_context, update_execution)

Might be a better way of doing this in python. Let me know your thoughts

google/cloud/aiplatform/metadata/artifact.py
nachocano: Done! thanks!

google/cloud/aiplatform/metadata/artifact.py
nachocano: This is the only method that I haven't tested manually, and will probably fail. Yes, it might be useful to have as an instance one. Let me see if I can do some changes

google/cloud/aiplatform/metadata/artifact.py
nachocano: Done, thanks for the pointer!

google/cloud/aiplatform/metadata/artifact.py
nachocano: Done

tests/unit/aiplatform/test_initializer.py
```diff
@@ -69,9 +70,17 @@ def test_init_location_with_invalid_location_raises(self):
         with pytest.raises(ValueError):
             initializer.global_config.init(location=_TEST_INVALID_LOCATION)
 
-    def test_init_experiment_sets_experiment(self):
-        initializer.global_config.init(experiment=_TEST_EXPERIMENT)
-        assert initializer.global_config.experiment == _TEST_EXPERIMENT
+    # TODO(nachocano) Mock the call to set_experiment
+    # def test_init_experiment_calls_metadata_service(self):
```
nachocano, line 74: @sasha-gitg how can I mock the call to set_experiment?

google/cloud/aiplatform/__init__.py
```diff
@@ -47,10 +48,24 @@
 """
 init = initializer.global_config.init
 
+log_params = metadata.metadata_service.log_params
+log_metrics = metadata.metadata_service.log_metrics
+set_experiment = metadata.metadata_service.set_experiment
+get_experiment = metadata.metadata_service.get_experiment
+set_run = metadata.metadata_service.set_run
```
jialuzh, line 55: we can discuss more offline on this, but based on the initial feedback from Nate, it is better to just use `init` to change or set experiment and run, rather than having separated `set` methods. The init call can take just an experiment or run as input parameter, which should not reset the other parameters.

google/cloud/aiplatform/initializer.py
```diff
@@ -44,7 +45,6 @@ class _Config:
...
         self._project = None
-        self._experiment = None
```
jialuzh, line 47: I think we need to keep this field so that the `init` call can reset `run` w/o changing experiment info.

google/cloud/aiplatform/metadata/metadata.py
```diff
@@ -0,0 +1,133 @@
...
+    def set_experiment(self, experiment: str):
+        if not experiment:
+            raise ValueError(f"Invalid experiment {experiment}.")
+
+        store = _MetadataStore.get()
+        if not store:
+            logging.info(
+                f"Creating a default MetadataStore for experiment {experiment}"
+            )
+            _MetadataStore.create()
```
jialuzh, line 44: As we have handle for already exist exception for create, we probably can just call create here directly w/o calling get.

google/cloud/aiplatform/metadata/metadata.py
```diff
@@ -0,0 +1,133 @@
...
+        execution = _Execution.get(execution_name=self._run)
+        if not execution:
+            execution = _Execution.create(
+                execution_id=self._run,
+                schema_title="system.Run",
+                schema_version="0.0.1",
+                metadata=params,
+            )
+        else:
+            execution = _Execution.update(execution_name=self._run, metadata=params)
```
jialuzh, line 94: I believe this would override existing parameters, we would need to append new parameters to existing execution before update.

google/cloud/aiplatform/initializer.py
```diff
@@ -88,8 +90,14 @@ def init(
             utils.validate_region(location)
             self._location = location
         if experiment:
-            logging.warning("Experiments currently not supported.")
-            self._experiment = experiment
+            metadata.metadata_service.set_experiment(experiment)
+        if run:
+            if not experiment:
```
jialuzh, line 95: Again as mentioned above, we should allow this per our current design.

google/cloud/aiplatform/initializer.py
```diff
@@ -88,8 +90,14 @@ def init(
...
+                raise ValueError(
+                    "No experiment set. Provide an experiment for this run, e.g., aiplatform.init("
+                    "experiment='my-experiment')."
+                )
+            metadata.metadata_service.set_run(run)
```
jialuzh, line 100: I could be wrong on this as I am not very family how python would handle calls to class methods w/o a created class, but wouldn't this call create a new metadata_service class every time we call the class method and the experiment information is not stored? I thought we would need to add metadata_service(or maybe call it metadata_config) class as a parameter in the global config class so that the metadata status would be stored globally and would be used for check for "log_param/metric" calls. Like what I was trying to do in https://github.com/jialuzh/python-aiplatform/blob/dev-test/google/cloud/aiplatform/initializer.py#L47.

google/cloud/aiplatform/metadata/metadata.py
```diff
@@ -0,0 +1,133 @@
...
+        if not context:
+            logging.info(f"Creating a Context for experiment {experiment}")
+            context = _Context.create(
+                context_id=experiment,
+                schema_title="system.Experiment",
+                schema_version="0.0.1",
+            )
+        self._experiment = context.name
+
+    def set_run(self, run: str):
```
jialuzh, line 56: As I am thinking we do no need to expose a separate set_run methods to end user, I think it is probably more straightforward to combine all the experiment/run/metric artifact initialization steps into one init function so that we can check and create all resources, and also connect them together.

jialuzh (changes requested):

google/cloud/aiplatform/metadata/artifact.py
nachocano: Forgot to push the changes.. Done, and cleaned up other unnecessary methods in child classes. Most of the work is being done in the base _Resource class.

google/cloud/aiplatform/metadata/artifact.py
nachocano: Remove update from child classes. Only needed in the parent _Resource. Did some other cleanups as well, to remove duplicated code, PTAL

google/cloud/aiplatform/__init__.py
nachocano: We can talk offline. The init does that, takes experiment and run. And ends up calling set_experiment and set_run, underneath. We might not expose the latter ones. 
IMO we should. I basically prefer this:

```
aiplatform.init(experiment='my-experiment', run='first-run')
...
aiplatform.set_run(run='second-run')
```

instead of this:

```
aiplatform.init(experiment='my-experiment', run='first-run')
...
aiplatform.init(run='second-run')
```

The code is already in place, so I think we can keep it. Let's discuss tomorrow

google/cloud/aiplatform/initializer.py
nachocano: This is kept on the _MetadataService class, and the way of doing that would be to just call `set_run`. 
We can discuss further, but I think the set_run way has cleaner semantics

google/cloud/aiplatform/metadata/metadata.py
nachocano: Already changed this to use an instance method. Indeed, we do append new parameters, or new metrics... Tested out and worked. Will share the notebook I used for my experiments shortly

google/cloud/aiplatform/initializer.py
nachocano: Think of the metadata_service as your Config. 
It's a singleton, so is created only once, in the last line of metadata.py. We use that object all over. Similar to what the initializer does

google/cloud/aiplatform/metadata/metadata.py
nachocano: The init currently calls set_experiment and set_run. So is basically creating everything, but it's still missing the piece of connecting them together. We can try to merge this asap, and then you can keep on adding the missing bits.

google/cloud/aiplatform/initializer.py
nachocano: Same comment as before. I think we can still change the design. We can talk more tmrw

google/cloud/aiplatform/metadata/metadata.py
nachocano: I think is better to try to get and then create if is not there. That will reflect API metrics more accurately. IIRC pipelines are doing something similar, because they don't want create calls to appear with an error code in stackdriver.

nachocano:
Run the following and worked:

```
!pip3 install --user -e /Users/nachocano/src/python-aiplatform

PROJECT = 'PROJECT_NAME' # replace me
EXPERIMENT = "experiment1"
RUN="run1"

aiplatform.constants.API_BASE_PATH = 'AUTOPUSH_ENDPOINT' # replace me

aiplatform.init(project=PROJECT, experiment=EXPERIMENT, run=RUN)
aiplatform.log_metrics({'rmse':123, 'accuracy': 0.9})
aiplatform.log_metrics({'rmse':222, 'accuracy': 1})
aiplatform.log_metrics({'log_loss': 0.04})
aiplatform.log_params({'learning_rate': 0.01, 'dropout': 0.2})
aiplatform.log_params({'learning_rate': 0.1, 'dropout': 0.3, 'lambda': 0.001})
```

google/cloud/aiplatform/initializer.py
nachocano: Synced offline with @jialuzh, he is planning on taking care of these things in follow up PRs. Would that be ok?

nachocano:
@sasha-gitg this is ready for another look. It'd be great if we can get this in today, so that @jialuzh can resume the work, add more UTs, and continue adding more functionality

nachocano:
Unexposed log_param and log_metric and we wanted to keep the api simpler. Removing get_pipeline as well, as we won't expose it in the first cut

google/cloud/aiplatform/__init__.py
nachocano: Discussed offline. We can get this in, and then un-expose if not needed

google/cloud/aiplatform/initializer.py
nachocano: Discussed offline. We can get this in, and then un-expose if not needed

google/cloud/aiplatform/metadata/metadata.py
nachocano: Discussed offline. We can get this in, and then un-expose if not needed

google/cloud/aiplatform/metadata/metadata.py
nachocano: Let's leave it as is for now, and then we can revisit

google/cloud/aiplatform/initializer.py
sasha-gitg: SGTM

google/cloud/aiplatform/metadata/resource.py
```diff
@@ -0,0 +1,300 @@
...
+                Custom credentials to use to upload this model. Overrides
+                credentials set in aiplatform.init.
+        """
+
+        super().__init__(
+            project=project, location=location, credentials=credentials,
+        )
+
+        # If we receive a full resource name, we extract the metadata_store_id and use that
+        if resource_name.find("/") != -1:
```
sasha-gitg, line 72:
```suggestion
        if "/" in resource_name:
```

nachocano: Done, thanks!

nachocano:
Comment addressed, PTAL

google/cloud/aiplatform/metadata/resource.py
```diff
@@ -0,0 +1,300 @@
...
+        return cls(
+            resource_name=f"{parent}/{cls._resource_noun}/{resource_id}",
+            metadata_store_id=metadata_store_id,
+            project=project,
+            location=location,
+            credentials=credentials,
+        )
+
+    @classmethod
+    def get(
```
sasha-gitg, line 172: This seems redundant with the constructor. Why do we need a get API with the constructor which also gets?

google/cloud/aiplatform/metadata/resource.py
```diff
@@ -0,0 +1,300 @@
...
+        Args:
+            metadata (Dict):
+                Optional. metadata contains the updated metadata information.
+            credentials (auth_credentials.Credentials):
+                Custom credentials to use to update this resource. Overrides
+                credentials set in aiplatform.init.
+
+        """
+
+        if not self._gca_resource:
```
sasha-gitg, line 235: It seems like this shouldn't occur. In what scenario do we not have the local _gca_resource available?

google/cloud/aiplatform/metadata/resource.py
```diff
@@ -0,0 +1,300 @@
...
+        gca_resource.metadata.update(metadata)
+        api_client = self._instantiate_client(credentials=credentials)
+
+        try:
+            update_gca_resource = self.__class__._update_resource(
+                client=api_client, resource=gca_resource,
+            )
+            self._gca_resource = update_gca_resource
+        except exceptions.NotFound:
+            logging.info(f"Resource to update '{self.name}' not found.")
```
sasha-gitg, line 248: I think we shouldn't catch this and instead allow the service exception to propagate up. Or is there a scenario where we would expect this to occur?

google/cloud/aiplatform/metadata/resource.py
```diff
@@ -0,0 +1,300 @@
...
+
+        if not self._gca_resource:
+            raise ValueError("Invalid resource to update")
+
+        gca_resource = deepcopy(self._gca_resource)
+        gca_resource.metadata.update(metadata)
+        api_client = self._instantiate_client(credentials=credentials)
+
+        try:
+            update_gca_resource = self.__class__._update_resource(
```
sasha-gitg, line 243:
```suggestion
            update_gca_resource = self._update_resource(
```

google/cloud/aiplatform/metadata/resource.py
```diff
@@ -0,0 +1,300 @@
...
+    @classmethod
+    @abc.abstractmethod
+    def _update_resource(
+        cls, client: utils.MetadataClientWithOverride, resource: proto.Message,
+    ) -> proto.Message:
+        """Update resource method."""
+        pass
+
+    @classmethod
+    def _extract_metadata_store_id(cls, resource_name, resource_noun) -> str:
```
sasha-gitg, line 275: This seems like it can be a staticmethod.

google/cloud/aiplatform/metadata/metadata.py
```diff
@@ -0,0 +1,145 @@
...
+
+class _MetadataService:
+    """Contains the exposed APIs to interact with the Managed Metadata Service."""
+
+    def __init__(self):
+        self._experiment = None
+        self._run = None
+
+    def set_experiment(self, experiment: str):
+        if not experiment:
```
sasha-gitg, line 35: This check shouldn't be needed.

google/cloud/aiplatform/metadata/metadata.py
```diff
@@ -0,0 +1,145 @@
...
+            )
+        self._experiment = context.name
+
+    def set_run(self, run: str):
+        if not self._experiment:
+            raise ValueError(
+                "No experiment set for this run. Make sure to call aiplatform.init(experiment='my-experiment') or "
+                "aiplatform.set_experiment(experiment='my-experiment') before trying to set_run. "
+            )
+        if not run:
```
sasha-gitg, line 61: Can also skip this check.

google/cloud/aiplatform/metadata/metadata.py
```diff
@@ -0,0 +1,145 @@
...
+            raise ValueError(f"Invalid run {run}.")
+
+        execution = _Execution.get(resource_name=run)
```
sasha-gitg, line 64: I thought Runs were going to be Contexts with a single Execution underneath. Did that change? @jialuzh

google/cloud/aiplatform/metadata/metadata.py
```diff
@@ -0,0 +1,145 @@
...
+                f"Creating a default MetadataStore for experiment {experiment}"
+            )
+            _MetadataStore.create()
+
+        context = _Context.get(resource_name=experiment)
+        if not context:
+            logging.info(f"Creating a Context for experiment {experiment}")
+            context = _Context.create(
+                resource_id=experiment,
+                schema_title="system.Experiment",
```
sasha-gitg, line 50: Preference to store schema titles as constants in a centralized location.

google/cloud/aiplatform/metadata/metadata.py
```diff
@@ -0,0 +1,145 @@
...
+                schema_version="0.0.1",
```
sasha-gitg, line 51: And map the schema titles to versions.

google/cloud/aiplatform/metadata/metadata.py
```diff
@@ -0,0 +1,145 @@
...
+            raise ValueError(
+                "No experiment set for this run. Make sure to call aiplatform.init(experiment='my-experiment') or "
+                "aiplatform.set_experiment(experiment='my-experiment') before trying to set_run. "
+            )
+        if not run:
+            raise ValueError(f"Invalid run {run}.")
+
+        execution = _Execution.get(resource_name=run)
+        if not execution:
+            logging.info(f"Creating an Execution for run {run}")
```
sasha-gitg, line 66: Preference for "Creating a run" since Execution isn't surfaced to the user.

google/cloud/aiplatform/metadata/metadata.py
```diff
@@ -0,0 +1,145 @@
...
+            execution = _Execution.create(
+                resource_id=run, schema_title="system.Run", schema_version="0.0.1",
+            )
+        self._run = execution.name
+
+    def log_param(self, name: str, value):
```
sasha-gitg, line 72: Value should be `Union[float, int, str]`

google/cloud/aiplatform/metadata/metadata.py
```diff
@@ -0,0 +1,145 @@
...
+        if not isinstance(value, (float, int, str)):
+            raise ValueError(
+                f"Invalid type for parameter value {value}. It should be one of [int, float, str]"
+            )
+        return self.log_params({name: value})
+
+    def log_params(self, params: Dict):
+        if not self._experiment:
```
sasha-gitg, line 80: Capturing that the preference is to automatically create experiment and run if one has not been set.

google/cloud/aiplatform/metadata/metadata.py
```diff
@@ -0,0 +1,145 @@
...
+        self._run = execution.name
+
+    def log_param(self, name: str, value):
+        if not isinstance(value, (float, int, str)):
+            raise ValueError(
+                f"Invalid type for parameter value {value}. It should be one of [int, float, str]"
+            )
+        return self.log_params({name: value})
+
+    def log_params(self, params: Dict):
```
sasha-gitg, line 79: Type on params should be `Dict[str, Union[float, in, str]]`.

google/cloud/aiplatform/metadata/metadata.py
```diff
@@ -0,0 +1,145 @@
...
+                "to log_params. "
+            )
+        if not self._run:
+            raise ValueError(
+                "No run set for logging parameters. Make sure to call aiplatform.init(experiment='my-experiment', "
+                "run='my-run') or aiplatform.set_run('my-run') before trying to log_params. "
+            )
+        execution = _Execution.get(resource_name=self._run)
+        if not execution:
+            logging.info(f"Creating an Execution for run {self._run}")
```
sasha-gitg, line 93: Same comment about logging about Execution in this block.

google/cloud/aiplatform/metadata/metadata.py
```diff
@@ -0,0 +1,145 @@
...
+                schema_title="system.Run",
+                schema_version="0.0.1",
+                metadata=params,
+            )
+        else:
+            logging.info(f"Updating Execution for run {self._run}")
+            execution.update(metadata=params)
+        self._run = execution.name
+
+    def log_metric(self, name: str, value):
```
sasha-gitg, line 105: Value type should be `Union[str, float, int]`

google/cloud/aiplatform/metadata/metadata.py
```diff
@@ -0,0 +1,145 @@
...
+        if not isinstance(value, (float, int, str)):
```
sasha-gitg, line 106: Can skip this check.

google/cloud/aiplatform/metadata/metadata.py
```diff
@@ -0,0 +1,145 @@
...
+            raise ValueError(
+                f"Invalid type for metric value {value}. It should be one of [int, float, str]"
+            )
+        return self.log_metrics({name: value})
+
+    def log_metrics(self, metrics: Dict):
```
sasha-gitg, line 112: `Dict[str, Union[str, float, int]]`

google/cloud/aiplatform/metadata/metadata.py
```diff
@@ -0,0 +1,145 @@
...
+        if not self._experiment:
```
sasha-gitg, line 113: I would dedup these validation blocks and create a helper method `_validate_experiment_and_run`

google/cloud/aiplatform/metadata/resource.py
```diff
@@ -0,0 +1,300 @@
...
+
+    @classmethod
+    def get(
+        cls,
+        resource_name: str,
+        metadata_store_id: Optional[str] = "default",
+        project: Optional[str] = None,
+        location: Optional[str] = None,
+        credentials: Optional[auth_credentials.Credentials] = None,
+    ) -> "_Resource":
```
sasha-gitg, line 179: `Optional['_Resource']`

google/cloud/aiplatform/metadata/resource.py
sasha-gitg: After reviewing the code up the stack, I think this API should be a `get_or_create` instead of pushing that logic to the caller.

google/cloud/aiplatform/metadata/resource.py
```diff
@@ -0,0 +1,300 @@
...
+                project=project,
+                location=location,
+                credentials=credentials,
+            )
+        except exceptions.NotFound:
+            logging.info(f"Resource {resource_name} not found.")
+
+    def update(
+        self,
+        metadata: Optional[Dict] = None,
```
sasha-gitg, line 221: This should strictly be a `Dict` because passing Non would result in a TypeError on line 239.

tests/unit/aiplatform/test_initializer.py
sasha-gitg: Preferably mock the service calls that get called underneath.

google/cloud/aiplatform/metadata/metadata.py
nachocano: Done, thanks!

google/cloud/aiplatform/metadata/metadata.py
nachocano: Done, thanks!

google/cloud/aiplatform/metadata/metadata.py
nachocano: Done, thanks!

google/cloud/aiplatform/metadata/metadata.py
nachocano: Done!

google/cloud/aiplatform/metadata/metadata.py
nachocano: Yes, it changed. Runs are executions that will store parameters in their metadata. Experiments are contexts. And metrics is a single artifact. Let me see if I can share the latest doc
We are still missing adding the edges, which Jialu will be working on.

nachocano: Sent the doc offline

google/cloud/aiplatform/metadata/metadata.py
nachocano: Makes sense. Done! 
These were just for me testing stuff, we can leave them for now and then see.
It'd be nice to only log if setting a verbose flag or something. Is that something you folks already have in place?

google/cloud/aiplatform/metadata/metadata.py
nachocano: neat, thanks!

google/cloud/aiplatform/metadata/metadata.py
nachocano: Done, thanks!

google/cloud/aiplatform/metadata/metadata.py
nachocano: @jialuzh mentioned the same. He is going to clean this up a bit. Seems that Nate doesn't want the set_run, nor other stuff. If you don't mind, we can leave it as is for now, and further discuss?
I'm more of the idea of actually not automagically doing things. IMO the user should have done an init or set_experiment, set_run before... But I have no strong opinion..

google/cloud/aiplatform/metadata/metadata.py
nachocano: Done, here and other places, thanks, makes sense

google/cloud/aiplatform/metadata/metadata.py
nachocano: Thanks!

google/cloud/aiplatform/metadata/metadata.py
nachocano: Thanks, I'm starting to like python 3

google/cloud/aiplatform/metadata/metadata.py
nachocano: Done, thanks!

google/cloud/aiplatform/metadata/metadata.py
nachocano: Done, thanks!

google/cloud/aiplatform/metadata/resource.py
nachocano: I tried using the constructor but I wasn't able to properly catch exceptions... Also from the caller, it felt strange to do something like:

`Execution(execution_name="my-execution")`, which was actually doing an RPC call. 

Let me see if I can change it to get_or_create, and make the create and get private...

google/cloud/aiplatform/metadata/resource.py
nachocano: Done, thanks!

google/cloud/aiplatform/metadata/resource.py
nachocano: Gotcha. 
Yes, I was doing some ugly stuff in the _update_resource because of that...
`metadata=metadata if metadata else {}`, as it was failing marshalling.
Thanks!

google/cloud/aiplatform/metadata/resource.py
nachocano: we should have it, let me remove this. Thanks!

google/cloud/aiplatform/metadata/resource.py
nachocano: Done, thanks!

google/cloud/aiplatform/metadata/resource.py
nachocano: Done, you are right, I think it shouldn't. Removed

google/cloud/aiplatform/metadata/resource.py
nachocano: Done! thanks!

google/cloud/aiplatform/metadata/resource.py
nachocano: Only exposing get_or_create and update now. Thanks for the suggestion, caller code is much cleaner now

google/cloud/aiplatform/metadata/metadata.py
nachocano: Let's actually leave this for a follow up once we agree with Nate tomorrow... SG?

tests/unit/aiplatform/test_initializer.py
nachocano: OK, ended up mocking the the set calls. Once we create the UTs for the metadata_service, I think we can mock what they actually end up calling...

nachocano:
Thanks for all the comments @sasha-gitg, all have been addressed. 
Ready for another look. Thanks!

google/cloud/aiplatform/metadata/metadata.py
sasha-gitg: SGTM

sasha-gitg (approved):

nachocano:
Thanks for approving @sasha-gitg ! Can we get this merged or need to wait for CI to do it?

sasha-gitg:
Merged. Thanks Nacho!

PR #310, opened 2021-04-13
fix: make logging more informative during training

sasha-gitg:


google/cloud/aiplatform/training_jobs.py
```diff
@@ -646,12 +646,12 @@ def _block_until_complete(self):
...
                 )
                 log_wait = min(log_wait * multiplier, max_wait)
-            previous_time = current_time
+                previous_time = current_time
             time.sleep(wait)
 
         self._raise_failure()
 
-        _LOGGER.log_action_completed_against_resource("", "run", self)
+        _LOGGER.log_action_completed_against_resource("run", "completed", self)
```
ivanmkc, line 654: From the docstring of `log_action_completed_against_resource`, which is:
```
        Args:
            noun (str): Noun the action acts on against the resource. Can be empty string.
            action (str): Action started against resource. ie: "Deployed". Can be empty string.
            resource_noun_obj (AiPlatformResourceNoun):
                Resource noun object the action is acting against
```

I can't tell if these args are correct or not.

What does `Noun the action acts on against the resource.` mean?

google/cloud/aiplatform/training_jobs.py
```diff
@@ -646,12 +646,12 @@ def _block_until_complete(self):
                     )
                 )
                 log_wait = min(log_wait * multiplier, max_wait)
-            previous_time = current_time
```
ivanmkc, line 649: LGTM

google/cloud/aiplatform/training_jobs.py
sasha-gitg: The noun-action combo is for nouns in the following:
- _run_ completed
- _data_ imported
- _model_ deployed 

Sometimes these nouns are separate from the resource they act on. Like data/dataset and model/endpoint. Happy to improve the wording in the comment.

ivanmkc: Okay I see!

I was having trouble parsing the sentence. I get it now.

Perhaps could be changed to:
`The noun that the action acts on against the resource.`,
`The action that was started against the resource.`, etc.

The examples would actually be the most helpful.

ivanmkc (approved):
LGTM

PR #314, opened 2021-04-14
Api

jialuzh:
(Feat) Add logics to connect experiment, run and metrics together.
Update run and metrics resource name.
Add display name and experiment context UI distinguish field.
Add UTs for metadata_service methods
Add UTs for resource update methods


jialuzh:
@sasha-gitg 
@nachocano

google/cloud/aiplatform/metadata/context.py
```diff
@@ -57,3 +58,82 @@ def _update_resource(
         cls, client: utils.MetadataClientWithOverride, resource: proto.Message,
     ) -> proto.Message:
         return client.update_context(context=resource)
+
+    @classmethod
+    def add_artifacts_or_executions(
+        cls,
+        context_id: str,
+        artifact_ids: Optional[Sequence[str]] = None,
```
sasha-gitg, line 66: For these methods that take nodes as parameters, it's much more preferable to:
1. Pass in the object representations so we are not duplicating the resource naming code at every call and can just rely on the constructor of the class.
2. They should be instance methods instead of class methods so we are not duplicating the api client instantiation at every call.

It seems like that would reduce the amount of code needed for these methods and avoid the requirement for passing in project/location/creds at every call.

The source in this method would reduce to:

```python
self.api_client.add_context_artifacts_and_executions(
            context=self.resource_name,
            artifacts=[node.resource_name for node in artifacts],
            executions=[node.resource_name for node in executions],
        )
```

google/cloud/aiplatform/metadata/execution.py
```diff
@@ -57,3 +59,64 @@ def _update_resource(
         cls, client: utils.MetadataClientWithOverride, resource: proto.Message,
     ) -> proto.Message:
         return client.update_execution(execution=resource)
+
+    @classmethod
+    def add_artifact(
```
sasha-gitg, line 64: Same comment about converting these into instance methods.

google/cloud/aiplatform/metadata/metadata.py
```diff
@@ -30,13 +30,16 @@ class _MetadataService:
...
     def set_experiment(self, experiment: str):
         _MetadataStore.get_or_create()
         context = _Context.get_or_create(
             resource_id=experiment,
+            display_name=experiment,
             schema_title=constants.SYSTEM_EXPERIMENT,
             schema_version=constants.SCHEMA_VERSIONS[constants.SYSTEM_EXPERIMENT],
+            metadata={constants.UI_DETECTION_KEY: constants.UI_DETECTION_VALUE},
         )
         self._experiment = context.name
```
sasha-gitg, line 44: If we went with instance methods, we could reference the context directly and not have to indirectly track the name.

google/cloud/aiplatform/metadata/metadata.py
```diff
@@ -46,12 +49,30 @@ def set_run(self, run: str):
...
+            display_name=metrics_artifact_id,
+            schema_title=constants.SYSTEM_METRICS,
+            schema_version=constants.SCHEMA_VERSIONS[constants.SYSTEM_METRICS],
+        )
+        _Execution.add_artifact(
+            execution_id=run_execution_id, artifact_id=metrics_artifact_id, input=False
+        )
+
         self._run = execution.name
+        self._metric = artifact.name
```
sasha-gitg, line 75: Same comment about setting the _run and _metric as the underlying object.

tests/unit/aiplatform/test_metadata.py
```diff
@@ -0,0 +1,265 @@
+# -*- coding: utf-8 -*-
```
sasha-gitg, line 1: These tests are great!

sasha-gitg (changes requested):

google/cloud/aiplatform/metadata/constants.py
```diff
@@ -26,3 +26,6 @@
     SYSTEM_EXPERIMENT: _DEFAULT_SCHEMA_VERSION,
     SYSTEM_METRICS: _DEFAULT_SCHEMA_VERSION,
 }
+
+UI_DETECTION_KEY = "experiment_deleted"
```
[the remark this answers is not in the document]
jialuzh: Good pont. Done.

google/cloud/aiplatform/metadata/context.py
jialuzh: Thanks for the suggestion. I updated this method and the one in exeuction.py to instance methods. I did changed the input artifacts/executions to full resource str instead of the full object as the full resource name should be all we need to call the backend API and we would already have that if we have the resource object.

Also updated metadata.py and related tests.

google/cloud/aiplatform/metadata/constants.py
```diff
@@ -26,3 +26,8 @@
     SYSTEM_EXPERIMENT: _DEFAULT_SCHEMA_VERSION,
     SYSTEM_METRICS: _DEFAULT_SCHEMA_VERSION,
 }
+
+# GCP UI currently only displays experiment context which has a "experiment_deleted" key and False value.
+# This is a temporary solution to integrate tensorboard results into GCP UI.
```
nachocano, line 31: I would just mention something that this is needed until the backend supports deletion of contexts. Don't give much details about GCP UI or anything, as this is public stuff

google/cloud/aiplatform/metadata/metadata.py
jialuzh: I added an instance field to the class to track the experiment context. I think we would still want to keep the string/ID name here as those names are used to generate run/metric resource names.

google/cloud/aiplatform/metadata/constants.py
```diff
@@ -26,3 +26,8 @@
...
+UI_DETECTION_KEY = "experiment_deleted"
```
nachocano, line 32: nit: EXPERIMENT_DELETED_KEY maybe? Let's do not couple this to UI

google/cloud/aiplatform/metadata/context.py
```diff
@@ -57,3 +61,20 @@ def _update_resource(
         cls, client: utils.MetadataClientWithOverride, resource: proto.Message,
     ) -> proto.Message:
         return client.update_context(context=resource)
+
+    def add_artifacts_or_executions(
```
nachocano, line 65: add_artifacts_and_executions? instead of or?

google/cloud/aiplatform/metadata/execution.py
```diff
@@ -57,3 +60,23 @@ def _update_resource(
         cls, client: utils.MetadataClientWithOverride, resource: proto.Message,
     ) -> proto.Message:
         return client.update_execution(execution=resource)
+
+    def add_artifact(
+        self, artifact: str, input: bool,
```
nachocano, line 65: artifact_name maybe to be consistent with other methods?

google/cloud/aiplatform/metadata/metadata.py
jialuzh: I think for run and metrics, we might prefer to just record the resource ID here, the reason is that those resources would only be used in log_xxx, and get_xxx calls later. And we would want to query our BE to get the most up to date resources for logging data instead of using the one that was queried at set_run(), which might be out of date. 

I agree that the object would also include the resource ID/name, but I think it might cause some confusion about why we would not use them directly in the log_xxx calls and instead query our BE API for a new instances.

Let me know what you think.

google/cloud/aiplatform/metadata/context.py
```diff
@@ -1,5 +1,7 @@
 # -*- coding: utf-8 -*-
 
+from typing import Optional, Dict, Sequence
```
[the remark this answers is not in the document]
jialuzh: Oops. Updated this one and reformatted all the imports in our codes.

tests/unit/aiplatform/test_metadata.py
jialuzh: Thanks!

nachocano: +1

google/cloud/aiplatform/metadata/metadata.py
nachocano: I agree with Sasha, I think we should either track resources, or track names... no need to track both name and resources...

google/cloud/aiplatform/metadata/metadata.py
```diff
@@ -30,28 +30,49 @@ class _MetadataService:
     def __init__(self):
         self._experiment = None
         self._run = None
+        self._metric = None
+        self._experiment_resource = None
```
nachocano, line 34: probably just repurpose self._experiment to actually store the resource? same with others...

google/cloud/aiplatform/metadata/metadata.py
```diff
@@ -30,28 +30,49 @@ class _MetadataService:
     def __init__(self):
         self._experiment = None
         self._run = None
+        self._metric = None
```
nachocano, line 33: do you need to track _metric name here? You are generating it deterministically. Maybe just have a private method to retrieve the metric name as used it wherever needed

google/cloud/aiplatform/metadata/constants.py
nachocano: @jialuzh, we will also need to set another field: experiment_type = MB. Can you add it as part of this PR?

google/cloud/aiplatform/metadata/constants.py
jialuzh: Good point, comment updated. Actually do we have a bug for this which we can use as a TODO here?

google/cloud/aiplatform/metadata/constants.py
jialuzh: done

google/cloud/aiplatform/metadata/context.py
jialuzh: done

google/cloud/aiplatform/metadata/metadata.py
jialuzh: yes, we need this as we are creating the metric artifact in set_run now and in log_metrics, we need to get the artifact by its resource name.

google/cloud/aiplatform/metadata/constants.py
nachocano: b/182292666, but I don't know if we should be linking internal bugs either

google/cloud/aiplatform/metadata/metadata.py
jialuzh: Ok, I updated all 3 to store the whole resource instead of the name. And also added a comments in the log_xxx method to explain why we want to query a new resource instead of using the one we stored.

google/cloud/aiplatform/metadata/metadata.py
jialuzh: Changed my mind based you and Nacho's comments, now all 3 are storing resources.

google/cloud/aiplatform/metadata/execution.py
jialuzh: Updated artifact to artifact_name.

For type, I think it is better to keep it required as there is no real default behavior here. We would likely use this to assign artifact as an Input of a execution later, making it default as output could cause unexpected issues.

google/cloud/aiplatform/metadata/metadata.py
jialuzh: done.

google/cloud/aiplatform/metadata/constants.py
jialuzh: Hmm, I did saw there was some bug numbers in TODOs in other files. I just created a TODO w/o the bug number just in case.

google/cloud/aiplatform/metadata/metadata.py
```diff
@@ -58,8 +58,8 @@ def set_run(self, run: str):
             schema_title=constants.SYSTEM_RUN,
             schema_version=constants.SCHEMA_VERSIONS[constants.SYSTEM_RUN],
         )
-        self._experiment_resource.add_artifacts_or_executions(
-            executions=[run_execution.resource_name]
+        self._experiment_resource.add_artifacts_and_executions(
+            execution_resource_names=[run_execution.resource_name]
```
nachocano, line 60: nit: execution_names to maintain consistency with other methods

google/cloud/aiplatform/metadata/metadata.py
```diff
@@ -69,7 +69,7 @@ def set_run(self, run: str):
             schema_title=constants.SYSTEM_METRICS,
             schema_version=constants.SCHEMA_VERSIONS[constants.SYSTEM_METRICS],
         )
-        run_execution.add_artifact(artifact=metrics_artifact.resource_name, input=False)
+        run_execution.add_artifact(artifact_resource_name=metrics_artifact.resource_name, input=False)
```
nachocano: nit: artifact_name

tests/unit/aiplatform/test_metadata.py
```diff
@@ -92,7 +92,9 @@ def get_context_mock():
             display_name=_TEST_EXPERIMENT,
             schema_title=constants.SYSTEM_EXPERIMENT,
             schema_version=constants.SCHEMA_VERSIONS[constants.SYSTEM_EXPERIMENT],
-            metadata={constants.EXPERIMENT_DELETED_KEY: constants.EXPERIMENT_DELETED_VALUE},
+            metadata={
```
nachocano, line 95: this will probably break

nachocano:
LGTM

google/cloud/aiplatform/metadata/metadata.py
jialuzh: I'd prefer artifact_resource_name here as `resource.name` method would actually return the resource ID instead of the full resource name, so artifact_name could be misleading and confuse the user whether we are expecting a full resource name or just the resource ID here.

tests/unit/aiplatform/test_metadata.py
jialuzh: oops, thanks for catching this. Updated now.

google/cloud/aiplatform/metadata/metadata.py
nachocano: Example of other places:
https://github.com/googleapis/python-aiplatform/blob/3082f97837a4f2c033c9db4ad752f51419ef17cf/google/cloud/aiplatform/datasets/dataset.py#L49
https://github.com/googleapis/python-aiplatform/blob/3082f97837a4f2c033c9db4ad752f51419ef17cf/google/cloud/aiplatform/datasets/dataset.py#L58

no strong opinion, fine with either, up to you

jialuzh:
@nachocano @sasha-gitg I updated the PR to also include a UT change for metadata resource. Right now the test would run against the public get_or_create method instead of _create method. Please take another look and let me know if you have any input on this.

nachocano:
> @nachocano @sasha-gitg I updated the PR to also include a UT change for metadata resource. Right now the test would run against the public get_or_create method instead of _create method. Please take another look and let me know if you have any input on this.

This is much better! Thanks @jialuzh!

LGTM

sasha-gitg (approved):

PR #320, opened 2021-04-14
fix: fix list failing without order_by and local sorting

sasha-gitg:
b/185403782


vinnysenthil (approved):
LGTM - thanks for catching + fixing this!

PR #325, opened 2021-04-15
chore: prevent normalization of semver versioning

dandhlee:
When there is a patch version added to semver versioning, setuptools.setup(version) will normalize the versioning from -patch to .patch which is not correct SEMVER versioning. The added feature with setuptools.sic(version) will prevent this from happening.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -20,6 +20,21 @@ import os
 
 import setuptools  # type: ignore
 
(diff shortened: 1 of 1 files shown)


parthea (approved):

parthea (approved):

dandhlee:
closing this as we don't need to strictly follow semver guidelines

PR #326, opened 2021-04-16
(feat) implement get_experiment and get_pipeline methods for Metadata service

jialuzh:
This PR includes the following changes:
Add List methods and QueryExeuctionInputAndOutput method and associated UTs
Implements get_pipeline and get_experiment methods and associated UTs
Updated resouce init method to enable resource create directly with a proto.Message and skip the Get call.
Update package setup.py to include pandas as a new dependency.


jialuzh:
@nachocano @sasha-gitg Please help take a look at this PR

google/cloud/aiplatform/metadata/context.py
```diff
@@ -30,6 +31,25 @@ class _Context(_Resource):
     _resource_noun = "contexts"
     _getter_method = "get_context"
 
+    def add_artifacts_and_executions(
+        self,
+        artifact_resource_names: Optional[Sequence[str]] = None,
+        execution_resource_names: Optional[Sequence[str]] = None,
+    ):
+        """Creates a new Metadata resource.
```
nachocano, line 39: nit: does this creates a new metadata resource?

google/cloud/aiplatform/metadata/execution.py
```diff
@@ -31,6 +34,51 @@ class _Execution(_Resource):
     _resource_noun = "executions"
     _getter_method = "get_execution"
 
+    def add_artifact(
+        self, artifact_resource_name: str, input: bool,
+    ):
+        """Creates a new Metadata resource.
```
nachocano, line 40: same here

google/cloud/aiplatform/metadata/execution.py
```diff
@@ -31,6 +34,51 @@ class _Execution(_Resource):
     _resource_noun = "executions"
     _getter_method = "get_execution"
 
+    def add_artifact(
```
nachocano, line 37: shall we call this add_execution_events?

google/cloud/aiplatform/metadata/metadata.py
```diff
@@ -94,11 +119,78 @@ def log_metrics(self, metrics: Dict[str, Union[str, float, int]]):
...
+            aiplatform.log_params({'learning_rate': 0.1})
+            aiplatform.log_metrics({'accuracy': 0.9})
+
+            aiplatform.set_run(run='run-2')
+            aiplatform.log_params({'learning_rate': 0.2})
+            aiplatform.log_metrics({'accuracy': 0.95})
+
+            Will result in the following DataFrame
+            ___________________________________________________________________________
+            | experiment_name | run_name      | param.learning_rate | metric.accuracy |
```
nachocano, line 138: do we need the experiment_name column?

google/cloud/aiplatform/metadata/metadata.py
```diff
@@ -94,11 +119,78 @@ def log_metrics(self, metrics: Dict[str, Union[str, float, int]]):
...
+            Args:
+              experiment: Name of the Experiment to filter results.
+
+            Returns:
+              Pandas Dataframe of Experiment with metrics and parameters.
+            """
+
+        # TODO remove this extra API call once our BE can support project ID in filters.
+        try:
+            experiment_context = _Context(resource_name=experiment)
```
nachocano, line 153: IIRC I was doing the get to avoid catching exceptions in the caller, and just doing something like:

```
experiment_context  = _Context.get(resource_name=experiment)
if not experiment_context:
   ....
```

No strong opinion, this also works

google/cloud/aiplatform/metadata/metadata.py
```diff
@@ -94,11 +119,78 @@ def log_metrics(self, metrics: Dict[str, Union[str, float, int]]):
...
+        # TODO remove this extra API call once our BE can support project ID in filters.
+        try:
+            pipeline_context = _Context(resource_name=pipeline)
+        except exceptions.NotFound:
+            logging.error(f"Pipeline {pipeline} not found.")
+            return
+        if pipeline_context.schema_title != constants.SYSTEM_PIPELINE:
+            raise ValueError(
+                f"Please provide a valid pipeline name. {pipeline} is not a pipeline."
+            )
```
nachocano, line 187: nit: can probably dedupe all this if you create a private _get_context or something similar and pass the resource_name, "pipeline"|"experiment" and the schema_title?

google/cloud/aiplatform/metadata/metadata.py
```diff
@@ -111,5 +203,50 @@ def _validate_experiment_and_run(self, method_name: str):
                 f"No run set. Make sure to call aiplatform.set_run('my-run') before trying to {method_name}. "
             )
 
+    def _query_runs_to_data_frame(
+        self, context_name: str, context_resource_name: str, source: str
+    ) -> pd.DataFrame:
+        filter = f'schema_title="{constants.SYSTEM_RUN}" AND in_context("{context_resource_name}")'
```
nachocano, line 209: here should be SYSTEM_PIPELINE_RUN for the other one?

google/cloud/aiplatform/metadata/metadata.py
```diff
@@ -111,5 +203,50 @@ def _validate_experiment_and_run(self, method_name: str):
...
+        run_executions = _Execution.list(filter=filter)
+
+        context_dict = []
```
nachocano, line 212: dict or array?

google/cloud/aiplatform/metadata/metadata.py
```diff
@@ -47,7 +47,7 @@ def set_experiment(self, experiment: str):
         )
         self._experiment = context
 
-    def set_run(self, run: str):
+    def start_run(self, run: str):
```
nachocano, line 50: might need to check the ValueError descriptions, I'm not sure if we were mentioning to call set_run somewhere?

nachocano:
Thanks @jialuzh ! Just a few cosmetic comments. Please feel free to address if you want.

LGTM

google/cloud/aiplatform/metadata/execution.py
jialuzh: The BE API method was called add_execution_event as it takes event as the input instead of the artifact. Here we basically hide the event concept and directly input artifact as the input so I think add_artifact is more proper. Also, as this is an instance method in the exeuction class, I think we do not need to have execution in the name either to avoid duplication.

google/cloud/aiplatform/metadata/metadata.py
jialuzh: I think having the experiment_name in the df would help user for their analysis. Some use case I can imaging would be user might want to pull a couple different experiment, or pipeline data and then append them together for some comparison(ex, maybe compare recent experiment with previous ones). And in those cases, the experiment_name column would help them distinguish the data.

google/cloud/aiplatform/metadata/metadata.py
jialuzh: Good point, I think get can help remove some of the duplicated codes. But I believe _get was designed to be a private method right now, so probably should not directly be called here? Do you think we should make the get method public?

google/cloud/aiplatform/metadata/metadata.py
jialuzh: Actually no.  SYSTEM_PIPELINE_RUN is a context scheme in stead of execution schema. basically under each system.pipeline, there is a SYSTEM_PIPELINE_RUN and a system.Run, the pipelineRun would store the DAG graph, and the Run will more or less be a shortcut to store param and connect to metric artifacts. Dushyanth's doc would have some more info although I think some edges was omitted in his doc.

google/cloud/aiplatform/metadata/metadata.py
nachocano: SG

google/cloud/aiplatform/metadata/execution.py
nachocano: SG

google/cloud/aiplatform/metadata/metadata.py
nachocano: I made it private because I was just using get_or_create, but I think we can make it public, yes, in case you prefer this other style instead of catching exceptions in metadata.py...

google/cloud/aiplatform/metadata/metadata.py
nachocano: ack, SG

google/cloud/aiplatform/metadata/context.py
jialuzh: Oops. Thanks for catching this. Updated.

google/cloud/aiplatform/metadata/execution.py
jialuzh: updated.

google/cloud/aiplatform/metadata/metadata.py
jialuzh: Good point. refracted this part of code to a method.

google/cloud/aiplatform/metadata/metadata.py
jialuzh: renamed this to context_summary to avoid the confusion.

google/cloud/aiplatform/metadata/metadata.py
jialuzh: Thanks for catching this. I updated all the ValueErrors and also some comments in the file to start_run

google/cloud/aiplatform/metadata/resource.py
```diff
@@ -69,23 +74,32 @@ def __init__(
             project=project, location=location, credentials=credentials,
```
jialuzh, line 74: @sasha-gitg I still used the __init__ method here instead of the empty constructor as the implementation for both looks the same to me. The only field we would need to set after this call would be _gca_resource which we have special handle for it already, so I think we should be ok just use the normal init method. But please let me know if i missed anything.

google/cloud/aiplatform/metadata/metadata.py
```diff
@@ -94,11 +119,68 @@ def log_metrics(self, metrics: Dict[str, Union[str, float, int]]):
         )
         artifact.update(metadata=metrics)
 
-    def get_experiment(self, experiment: str):
-        raise NotImplementedError("get_experiment not implemented")
+    def get_experiment(self, experiment: str) -> pd.DataFrame:
```
sasha-gitg, line 122: I think experiment should be Optional so if I call `get_experiment()` then I get the experiment dataframe for the current set experiment.

google/cloud/aiplatform/metadata/metadata.py
```diff
@@ -94,11 +119,68 @@ def log_metrics(self, metrics: Dict[str, Union[str, float, int]]):
...
+            Returns:
+              Pandas Dataframe of Experiment with metrics and parameters.
+            """
+
+        source = "experiment"
+        experiment_resource_name = self._get_experiment_or_pipeline_resource_name(
+            name=experiment, source=source, expected_schema=constants.SYSTEM_EXPERIMENT,
+        )
+
+        if experiment_resource_name:
```
sasha-gitg, line 156: It looks like this Optionally returns an Experiment. I think we should throw if the experiment is not found. Same comment for Pipelines below.

google/cloud/aiplatform/metadata/metadata.py
```diff
@@ -108,8 +190,68 @@ def _validate_experiment_and_run(self, method_name: str):
...
+                f"No run set. Make sure to call aiplatform.start_run('my-run') before trying to {method_name}. "
+            )
+
+    @staticmethod
+    def _get_experiment_or_pipeline_resource_name(
+        name: str, source: str, expected_schema: str
+    ) -> str:
+        try:
+            context = _Context(resource_name=name)
+        except exceptions.NotFound:
```
sasha-gitg, line 202: I think this should be raised.

google/cloud/aiplatform/metadata/metadata.py
```diff
@@ -108,8 +190,68 @@ def _validate_experiment_and_run(self, method_name: str):
...
         if not self._run:
             raise ValueError(
-                f"No run set. Make sure to call aiplatform.set_run('my-run') before trying to {method_name}. "
+                f"No run set. Make sure to call aiplatform.start_run('my-run') before trying to {method_name}. "
+            )
+
+    @staticmethod
+    def _get_experiment_or_pipeline_resource_name(
+        name: str, source: str, expected_schema: str
+    ) -> str:
```
sasha-gitg, line 199: docstring

google/cloud/aiplatform/metadata/metadata.py
```diff
@@ -108,8 +190,68 @@ def _validate_experiment_and_run(self, method_name: str):
...
+            logging.error(f"{source} {name} not found.")
+            return
+        if context.schema_title != expected_schema:
+            raise ValueError(
+                f"Please provide a valid {source} name. {name} is not a {source}."
+            )
+        return context.resource_name
+
+    def _query_runs_to_data_frame(
+        self, context_name: str, context_resource_name: str, source: str
```
sasha-gitg, line 212: Is the context name the resource id on the context resource name?

google/cloud/aiplatform/metadata/resource.py
```diff
@@ -38,7 +38,8 @@ class _Resource(base.AiPlatformResourceNounWithFutureManager, abc.ABC):
 
     def __init__(
         self,
-        resource_name: str,
+        resource_name: str = None,
```
sasha-gitg, line 41:
```suggestion
        resource_name: str = Optional[None],
```

google/cloud/aiplatform/metadata/resource.py
```diff
@@ -38,7 +38,8 @@ class _Resource(base.AiPlatformResourceNounWithFutureManager, abc.ABC):
...
+        resource: proto.Message = None,
```
sasha-gitg, line 42: This type should probably be scoped down to a Union of the Artifact, Context, Execution protos. Also it's optional here.

google/cloud/aiplatform/metadata/resource.py
sasha-gitg: SGTM. I think if we ever expose this to users we will need remove it.

google/cloud/aiplatform/metadata/resource.py
```diff
@@ -69,23 +74,32 @@ def __init__(
...
+                location=self.location,
+            )
 
         self._gca_resource = getattr(self.api_client, self._getter_method)(
             name=full_resource_name
         )
 
+    @property
+    def metadata(self) -> Dict:
+        return self._gca_resource.metadata.items()
```
sasha-gitg, line 97: This returns a list of tuples. I think you want:
```python

json_format.to_dict(self._gca_resource.metadata)
```

google/cloud/aiplatform/metadata/resource.py
```diff
@@ -189,6 +203,70 @@ def update(
...
 
+    @classmethod
+    def list(
+        cls,
+        filter: Optional[str] = None,
+        metadata_store_id: Optional[str] = "default",
+        project: Optional[str] = None,
+        location: Optional[str] = None,
+        credentials: Optional[auth_credentials.Credentials] = None,
+    ):
```
sasha-gitg, line 214: Return annotation.

google/cloud/aiplatform/metadata/resource.py
```diff
@@ -189,6 +203,70 @@ def update(
         )
         self._gca_resource = update_gca_resource
 
+    @classmethod
+    def list(
+        cls,
+        filter: Optional[str] = None,
+        metadata_store_id: Optional[str] = "default",
```
sasha-gitg, line 210: Optional can be removed.

setup.py
```diff
@@ -45,6 +45,7 @@
     install_requires=(
         "google-api-core[grpc] >= 1.22.2, < 2.0.0dev",
         "proto-plus >= 1.10.1",
+        "pandas>=1.0.0",
```
sasha-gitg, line 48: This should be added to a extras_require field adjacent to install_requires ie:

```python
install_requires=(...),
extras_require={
    "full": ["pandas>=1.0.0"]
}
```

Then the install with pandas would be 

```
pip install google-cloud-aiplatform[full]
```

sasha-gitg (changes requested):

google/cloud/aiplatform/metadata/metadata.py
sasha-gitg: Also, I wonder if this should be `get_experiment_dataframe`.

google/cloud/aiplatform/metadata/execution.py
```diff
@@ -31,6 +34,51 @@ class _Execution(_Resource):
...
+
+        Returns:
+              A Sequence of _Artifacts
+        """
+
+        try:
+            artifacts = self.api_client.query_execution_inputs_and_outputs(
+                execution=self.resource_name
+            ).artifacts
+        except exceptions.NotFound:
```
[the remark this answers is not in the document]
jialuzh: If the execution does not have any artifacts associated with it then I think it would have a NotFound error. For our current use case if the run is setup correctly, then this should not happen, but this could be an issue if we use this API for other use cases in the future.

google/cloud/aiplatform/metadata/resource.py
jialuzh: I see. Seems like the tuple list also worked here. I wonder shall we still add another import and use dict instead of the tuple list?

google/cloud/aiplatform/metadata/artifact.py
```diff
@@ -58,3 +59,13 @@ def _update_resource(
...
     ) -> proto.Message:
         return client.update_artifact(artifact=resource)
+
+    @classmethod
+    def _list_resources(
+        cls,
+        client: utils.MetadataClientWithOverride,
+        parent: str,
+        filter: Optional[str] = None,
+    ):
```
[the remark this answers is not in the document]
jialuzh: done

google/cloud/aiplatform/metadata/metadata.py
jialuzh: Good point, made experiment as optional. 

For the function naming, let me loop in Nate as we had the get_experiment name in the DD. I can get another PR to update the name if we decide to change it.

google/cloud/aiplatform/metadata/metadata.py
jialuzh: done.

google/cloud/aiplatform/metadata/metadata.py
jialuzh: done

google/cloud/aiplatform/metadata/metadata.py
jialuzh: done

google/cloud/aiplatform/metadata/metadata.py
jialuzh: right, it would be the experiment or pipeline name, which is also the context_id. Updated the name to avoid confusion and added a docstring

google/cloud/aiplatform/metadata/resource.py
jialuzh: Should be "resource_name: Optional[str] = None," right?

google/cloud/aiplatform/metadata/resource.py
jialuzh: Done

google/cloud/aiplatform/metadata/resource.py
jialuzh: done. Actually does it make any difference between mark one as Optional and not optional but has a default value set?

google/cloud/aiplatform/metadata/resource.py
jialuzh: done

setup.py
jialuzh: done

google/cloud/aiplatform/metadata/metadata.py
jialuzh: Updated this part of code based on Sasha's feedback and we decided to raise the not found error instead of catch it. So no need to use the get method here.

google/cloud/aiplatform/metadata/resource.py
jialuzh: I see. Make sense. I update a couple more locations where metadata_store_id was set to "default" but also marked as Optional.

google/cloud/aiplatform/metadata/resource.py
jialuzh: discussed offline on this and we agreed to return a dict for consistency and future usage. 
I used dict(self._gca_resource.metadata.items()) to convert the list of tuple into a dict.

google/cloud/aiplatform/metadata/execution.py
```diff
@@ -55,28 +103,28 @@ def _create_resource(
...
+        list_request = ListExecutionsRequest(parent=parent, filter=filter,)
+        return client.list_executions(request=list_request)
 
-        self.api_client.add_execution_events(
-            execution=self.resource_name, events=[event],
-        )
+    @classmethod
+    def _update_resource(
+        cls, client: utils.MetadataClientWithOverride, resource: proto.Message,
+    ) -> proto.Message:
```
sasha-gitg, line 129: Docstring

google/cloud/aiplatform/metadata/resource.py
```diff
@@ -69,23 +77,32 @@ def __init__(
...
+                location=self.location,
+            )
 
         self._gca_resource = getattr(self.api_client, self._getter_method)(
             name=full_resource_name
         )
 
+    @property
+    def metadata(self) -> Dict:
+        return dict(self._gca_resource.metadata.items())
```
sasha-gitg, line 100: Please add TODO to update with json_format please.

sasha-gitg (approved):

google/cloud/aiplatform/metadata/execution.py
jialuzh: done. also added docstring for this method for other resources.

google/cloud/aiplatform/metadata/resource.py
jialuzh: done.

PR #337, opened 2021-04-17
feat: Added deploy_model_with_dedicated_resources_sample, deploy_model_with_automatic_resources_sample, upload_model and get_model samples

ivanmkc:
Added deploy_model_with_dedicated_resources_sample, deploy_model_with_automatic_resources_sample, upload_model and get_model samples and tests.


samples/model-builder/test_constants.py
```diff
@@ -16,6 +16,7 @@
 from uuid import uuid4
 
 from google.auth import credentials
+from google.cloud import aiplatform
```
ivanmkc, line 19: I needed to import this for `aiplatform.explain.ExplanationMetadata`

samples/model-builder/conftest.py
```diff
@@ -160,15 +160,32 @@ def mock_run_automl_image_training_job():
...
+    yield mock
+
+@pytest.fixture
+def mock_init_model(mock_model):
+    with patch.object(aiplatform, "Model") as mock:
+        mock.return_value = mock_model
+        yield mock
+
+@pytest.fixture
+def mock_batch_predict_model(mock_model):
```
ivanmkc, line 174: Updated these fixtures to use the new `mock_model`

samples/model-builder/deploy_model_custom_trained_model_sample.py
```diff
@@ -0,0 +1,67 @@
...
+def deploy_model_custom_trained_model_sample(
+    project,
+    location,
+    model_name: str,
+    machine_type: str,
+    endpoint: Optional["Endpoint"] = None,
+    deployed_model_display_name: Optional[str] = None,
+    traffic_percentage: Optional[int] = 0,
+    traffic_split: Optional[Dict[str, int]] = None,    
+    min_replica_count: Optional[int] = 1,
```
sasha-gitg, line 30: Please remove the Optional type hint (just leave `int`) on `min_replica_count` and `max_replica_count`. After reviewing this, I realized Endpoint.deploy and Model.deploy have inconsistent types on those arguments. I opened a PR to fix the function type hints in the SDK here: https://github.com/googleapis/python-aiplatform/pull/345

samples/model-builder/deploy_model_sample.py
```diff
@@ -0,0 +1,68 @@
...
+#  [START aiplatform_sdk_deploy_model_sample]
+def deploy_model_sample(
+    project,
+    location,
+    model_name: str,
+    endpoint: Optional["Endpoint"] = None,
+    deployed_model_display_name: Optional[str] = None,
+    traffic_percentage: Optional[int] = 0,
+    traffic_split: Optional[Dict[str, int]] = None,
+    min_replica_count: Optional[int] = 1,
```
sasha-gitg, line 29: Same comment on `min_replica_count` and `max_replica_count`

samples/model-builder/deploy_model_sample.py
```diff
@@ -0,0 +1,68 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+from google.cloud import aiplatform
+from typing import Optional, Sequence, Tuple, Dict
+from google.cloud.aiplatform import explain
+
+#  [START aiplatform_sdk_deploy_model_sample]
+def deploy_model_sample(
```
sasha-gitg, line 21: I'm assuming this is for models with automatic resources as machine type is not exposed. If that's the case, I think the following arguments can be removed but please confirm:

`accelerator_type`
`accelerator_count`
`explanation_metadata`
`explanation_parameters`

sasha-gitg (changes requested):

samples/model-builder/deploy_model_sample.py
ivanmkc: Yes it's for automatic resources.

I checked the internal deploy function and it looks like `accelerator_type` and `accelerator_count` if `machine-type` is not provided. However, even if it intuitively makes sense, I would argue that the docstrings should make their co-dependence clear. 

I would also recommend a refactor to bundle `machine_type`, `accelerator_type` and `accelerator_count` together, since they can't be used independently. If you want to keep the client-facing API flat, that's fine but in private functions it will be cleaner bundled.

From looking at `_deploy_call`, I don't see `explanation_metadata` and `explanation_parameters` relying on `machine_type` being set. Is there some restriction from the service the requires `machine_type?

samples/model-builder/deploy_model_custom_trained_model_sample.py
ivanmkc: Done.

samples/model-builder/deploy_model_sample.py
ivanmkc: Done.

samples/model-builder/deploy_model_sample.py
sasha-gitg: Agree on the refactor and updating the docstring. Only custom models take `explanation_metadata` and `explanation_parameters` as AutoMLs support it natively (if supported) and machine_type is required from custom models.

sasha-gitg: @vinnysenthil to confirm

ivanmkc: @sasha-gitg Logged a bug: https://b.corp.google.com/issues/186121905

samples/model-builder/deploy_model_sample.py
```diff
@@ -0,0 +1,59 @@
...
+    max_replica_count: int = 1,
+    metadata: Optional[Sequence[Tuple[str, str]]] = (),
+    sync: bool = True,
+):
+
+    aiplatform.init(project=project, location=location)
+
+    model = aiplatform.Model(model_name=model_name)
+
+    # AutoML models require a machine_type of None
```
sasha-gitg, line 39: @vinnysenthil do all AutoML Models require automatic resources?

ivanmkc: Actually, we talked about this and it's a little unclear what the truth of this is.

We're still trying to find out the answer from the predict team.

ivanmkc: @sasha-gitg I talked to prediction team and unfortunately, they do not have the answer. I will remove the comment for now and update this if I find out more.

sasha-gitg (approved):

vinnysenthil (approved):

PR #346, opened 2021-04-21
chore: remove TPU references from docstrings

sasha-gitg:


andrewferlitsch:
fyi> I saw a planning slide yesterday that showed TPU will be supported as preview in Q2.

vinnysenthil (approved):

PR #347, opened 2021-04-21
fix: rename get_experiment/pipeline to get_experiment_df and get_pipeline_df

jialuzh:
Rename 2 function name to make them more clear
Also fixed a bug that get_experiment_df without input would raise an exception


jialuzh:
@sasha-gitg I am renaming the get_experiment/pipeline methods per your and other's feedback. Please help take a look.

sasha-gitg (approved):

jialuzh:
@sasha-gitg I did some update to the test files and also fixed a small bug that I found while testing. Please take another look and merge the PR is it looks good to you.

PR #348, opened 2021-04-21
feat: Added explain tabular samples

ivanmkc:
Waiting for merge of https://github.com/googleapis/python-aiplatform/pull/337

### DIFF
--- a/samples/model-builder/conftest.py
+++ b/samples/model-builder/conftest.py
@@ -219,7 +219,7 @@ Model Fixtures
 
 @pytest.fixture
 def mock_model():
-    mock = MagicMock(aiplatform.Model)
+    mock = MagicMock(aiplatform.models.Model)
     yield mock
 
 
@@ -237,8 +237,9 @@ def mock_batch_predict_model(mock_model):
 
 
 @pytest.fixture
-def mock_upload_model():
+def mock_upload_model(mock_model):
     with patch.object(aiplatform.Model, "upload") as mock:
+        mock.return_value = mock_model
         yield mock
 
 
--- a/samples/model-builder/conftest.py
+++ b/samples/model-builder/conftest.py
@@ -219,7 +219,7 @@ Model Fixtures
 
 @pytest.fixture
 def mock_model():
-    mock = MagicMock(aiplatform.models.Model)
+    mock = MagicMock(aiplatform.Model)
     yield mock
 
 
@@ -235,10 +235,12 @@ def mock_batch_predict_model(mock_model):
     with patch.object(mock_model, "batch_predict") as mock:
         yield mock
 
+
 @pytest.fixture
 def mock_upload_model():
     with patch.object(aiplatform.Model, "upload") as mock:
         yield mock
+
 
 @pytest.fixture
 def mock_deploy_model(mock_model, mock_endpoint):
@@ -269,7 +271,7 @@ Endpoint Fixtures
 
 @pytest.fixture
 def mock_endpoint():
-    mock = MagicMock(aiplatform.models.Endpoint)
+    mock = MagicMock(aiplatform.Endpoint)
     yield mock
 
 
@@ -284,6 +286,7 @@ def mock_get_endpoint(mock_endpoint):
     with patch.object(aiplatform, "Endpoint") as mock_get_endpoint:
         mock_get_endpoint.return_value = mock_endpoint
         yield mock_get_endpoint
+
 
 @pytest.fixture
 def mock_endpoint_explain(mock_endpoint):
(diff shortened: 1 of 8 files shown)


samples/model-builder/conftest.py
```diff
@@ -219,7 +219,7 @@ def mock_run_custom_training_job():
 
 @pytest.fixture
 def mock_model():
-    mock = MagicMock(aiplatform.models.Model)
+    mock = MagicMock(aiplatform.Model)
```
vinnysenthil, line 222: This can be left as is (`aiplatform.models.Model`)

samples/model-builder/conftest.py
```diff
@@ -238,7 +238,7 @@ def mock_batch_predict_model(mock_model):
 
 @pytest.fixture
 def mock_upload_model():
-    with patch.object(aiplatform.models.Model, "upload") as mock:
+    with patch.object(aiplatform.Model, "upload") as mock:
         yield mock
```
vinnysenthil, line 242: We can should return a mock `Model` object here
```py
@pytest.fixture
def mock_upload_model(mock_model):
    with patch.object(aiplatform.Model, "upload") as mock:
        mock.return_value = mock_model
        yield mock
```

samples/model-builder/conftest.py
```diff
@@ -271,7 +271,7 @@ def mock_create_batch_prediction_job():
 
 @pytest.fixture
 def mock_endpoint():
-    mock = MagicMock(aiplatform.models.Endpoint)
+    mock = MagicMock(aiplatform.Endpoint)
```
vinnysenthil, line 274: This can also be left as is (`aiplatform.models.Endpoint`)

samples/model-builder/explain_tabular_sample.py
```diff
@@ -0,0 +1,36 @@
...
+):
+
+    aiplatform.init(project=project, location=location)
+
+    endpoint = aiplatform.Endpoint(endpoint_id)
+
+    response = endpoint.explain(instances=[instance_dict], parameters={})
+
+    for prediction_ in response.predictions:
+        print(prediction_)
```
vinnysenthil, line 33: Why the underscore at the end of the var name? Also `Endpoint.explain()` returns a high-level `Prediction` object that contains `predictions` **and** `explanations` ([see here](https://github.com/googleapis/python-aiplatform/blob/24d6920a530995a3a5c6ad6b25d0b867f7aebe27/google/cloud/aiplatform/models.py#L71-L73)). We should probably showcase both. @sasha-gitg thoughts?

vinnysenthil (changes requested):
Just one point about showcasing explanations returned on the `explain()` call + a few nits.

samples/model-builder/explain_tabular_sample.py
sasha-gitg: I agree that we should demonstrate that both `predictions` and `explanations` can get accessed from this object.

samples/model-builder/explain_tabular_sample.py
```diff
@@ -0,0 +1,36 @@
...
+# limitations under the License.
+
+
+from google.cloud import aiplatform
+
+from typing import Dict
+
+
+#  [START aiplatform_sdk_explain_tabular_sample]
+def explain_tabular_sample(
```
sasha-gitg, line 22: This could be a generic explain sample and doesn't necessarily need to be qualified as tabular.

sasha-gitg (approved):

samples/model-builder/explain_tabular_sample.py
ivanmkc: The sheet has me doing explain_tabular_sample, I assume it's used by the tech writers in such a sample.

samples/model-builder/explain_tabular_sample.py
ivanmkc: Sure, will add.

samples/model-builder/conftest.py
ivanmkc: Thanks, done.

samples/model-builder/explain_tabular_sample.py
```diff
@@ -0,0 +1,51 @@
...
+    project: str, location: str, endpoint_id: str, instance_dict: Dict
+):
+
+    aiplatform.init(project=project, location=location)
+
+    endpoint = aiplatform.Endpoint(endpoint_id)
+
+    response = endpoint.explain(instances=[instance_dict], parameters={})
+
+    for explanation in response.explanations:
```
ivanmkc, line 32: Same as GAPIC sample for consistency.

samples/model-builder/explain_tabular_sample.py
vinnysenthil: Perhaps if this is explicitly `explain_tabular_sample` we can also print feature attributions?

vinnysenthil (approved):
LGTM, thanks Ivan

PR #349, opened 2021-04-22
fix: Pass project to BQ and GCS clients

vinnysenthil:
Quick update to pass project to clients used in `iter_outputs()`

Manually tested `iter_outputs()` on batch prediction job that used BQ and GCS as outputs.

Fixes [b/182902518](http://b/182902518) 🦕

### DIFF
--- a/google/cloud/aiplatform/jobs.py
+++ b/google/cloud/aiplatform/jobs.py
@@ -724,7 +724,8 @@ class BatchPredictionJob(_Job):
 
             # Build a Storage Client using the same credentials as JobServiceClient
             storage_client = storage.Client(
-                credentials=self.api_client._transport._credentials
+                project=self.project,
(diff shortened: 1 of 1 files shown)


google/cloud/aiplatform/jobs.py
```diff
@@ -724,7 +724,8 @@ def iter_outputs(
 
             # Build a Storage Client using the same credentials as JobServiceClient
             storage_client = storage.Client(
-                credentials=self.api_client._transport._credentials
+                project=self.project,
```
sasha-gitg, line 727: Should the project be the same as the project for the resource as the config's project may be set to a different project? Is it a safe assumption that the jobs outputs will be written to the same project as the resource?

sasha-gitg (changes requested):

sasha-gitg (approved):

PR #350, opened 2021-04-22
fix: Add all supported uCAIP GA regions

vinnysenthil:
[Source of truth](http://shortn/_0dyRkrvrda)

Fixes [b/186123342](http://b/186123342) 🦕


sasha-gitg (approved):
Thanks Vinny!

PR #343, opened 2021-04-20
feat: Added create_training_pipeline_custom_job_sample and create_training_pipeline_custom_training_managed_dataset_sample and fixed create_training_pipeline_image_classification_sample

ivanmkc:
Added 2 samples with tests:
- create_training_pipeline_custom_job_sample
- create_training_pipeline_custom_training_managed_dataset_sample.py

Fixed:
- create_training_pipeline_image_classification_sample.py

PR #351, opened 2021-04-23
feat: Added create_training_pipeline_custom_package_job_sample and create_training_pipeline_custom_container_job_sample and reworked create_training_pipeline_custom_job_sample

ivanmkc:
Waiting for https://github.com/googleapis/python-aiplatform/pull/343 to merge.
Added 2 samples.

- Replaced samples/model-builder/create_training_pipeline_custom_training_managed_dataset_sample.py by incorporating managed dataset logic into other samples.


samples/model-builder/test_constants.py
```diff
@@ -149,3 +149,19 @@
...
+    [
+        f"--model_name={MODEL_NAME}",
+        "--model_base_path=$(AIP_STORAGE_URI)",
+        "--rest_api_port=8080",
+        "--port=8500",
+        "--file_system_poll_wait_seconds=31540000",
+    ],
+)
+MODEL_SERVING_CONTAINER_PREDICT_ROUTE = (f"/v1/models/{MODEL_NAME}:predict",)
+MODEL_SERVING_CONTAINER_HEALTH_ROUTE = f"/v1/models/{MODEL_NAME}"
```
vinnysenthil, line 163: Do we use these? They look like the partial path used to call the container via REST.

samples/model-builder/conftest.py
```diff
@@ -195,14 +195,46 @@ def mock_run_automl_image_training_job(mock_image_training_job):
...
+        mock.return_value = None
+        yield mock
+
+
+@pytest.fixture
+def mock_run_custom_package_training_job():
+    with patch.object(
+        aiplatform.training_jobs.CustomPythonPackageTrainingJob, "run"
+    ) as mock:
         yield mock
```
vinnysenthil, line 238: Please use the same pattern [as the AutoML TrainingJob fixtures](https://github.com/googleapis/python-aiplatform/blob/83b3268e7f4fdd1852759b0d96fba564facb302c/samples/model-builder/conftest.py#L140-L193). We should mock `CustomContainerTrainingJob` and `CustomPythonPackageTrainingJob` objects, like we[ did for `CustomTrainingJob`](https://github.com/googleapis/python-aiplatform/blob/83b3268e7f4fdd1852759b0d96fba564facb302c/samples/model-builder/conftest.py#L140-L143). Then patch "run" on that mock. We should not mock `__init__`.

vinnysenthil (changes requested):

samples/model-builder/test_constants.py
ivanmkc: Removed

samples/model-builder/conftest.py
ivanmkc: Fixed.

vinnysenthil (approved):
Looks good, thanks! Please hold merge for Sasha's review on MBSDK usage

sasha-gitg (approved):

PR #353, opened 2021-04-26
feat: Add export model

vinnysenthil:
### Summary of Changes
- Add `Model.export_model()`
- Transform and expose `Model.supported_export_formats`
- Override `__getattr__` in `AiPlatformResourceNoun` to fall back on the `_gca_resource` if a particular attribute has not been exposed by the SDK
- Add 5 unit tests to test `export_model()`

Fixes [b/186130284](http://b/186130284) 🦕

### DIFF
--- a/google/cloud/aiplatform/base.py
+++ b/google/cloud/aiplatform/base.py
@@ -541,14 +541,6 @@ class AiPlatformResourceNoun(metaclass=abc.ABCMeta):
     def __repr__(self) -> str:
         return f"{object.__repr__(self)} \nresource name: {self.resource_name}"
 
-    def __getattr__(self, attr) -> Any:
-        try:
-            return self._gca_resource.__getattr__(attr)
-        except AttributeError:
-            raise AttributeError(
-                f"'{self.__class__.__name__}' object has no attribute '{attr}'"
-            )
-
 
 def optional_sync(
     construct_object_on_arg: Optional[str] = None,
--- a/google/cloud/aiplatform/base.py
+++ b/google/cloud/aiplatform/base.py
@@ -541,6 +541,14 @@ class AiPlatformResourceNoun(metaclass=abc.ABCMeta):
     def __repr__(self) -> str:
         return f"{object.__repr__(self)} \nresource name: {self.resource_name}"
 
+    def __getattr__(self, attr) -> Any:
+        try:
+            return self._gca_resource.__getattr__(attr)
+        except AttributeError:
+            raise AttributeError(
+                f"'{self.__class__.__name__}' object has no attribute '{attr}'"
+            )
+
 
 def optional_sync(
     construct_object_on_arg: Optional[str] = None,
--- a/google/cloud/aiplatform/models.py
+++ b/google/cloud/aiplatform/models.py
@@ -1224,18 +1224,20 @@ class Model(base.AiPlatformResourceNounWithFutureManager):
         return self._gca_resource.description
 
     @property
-    def _supported_export_formats(self):
+    def _supported_export_formats(
+        self,
+    ) -> Dict[str, List[gca_model_compat.Model.ExportFormat.ExportableContent]]:
         """Store list of ExportFormat objects as a private dict for easy lookup"""
         return {
-            ef.id: [
+            export_format.id: [
                 gca_model_compat.Model.ExportFormat.ExportableContent(content)
-                for content in ef.exportable_contents
+                for content in export_format.exportable_contents
             ]
-            for ef in self._gca_resource.supported_export_formats
+            for export_format in self._gca_resource.supported_export_formats
         }
 
     @property
-    def supported_export_formats(self):
+    def supported_export_formats(self) -> List[str]:
         """The formats in which this Model may be exported.
         If empty, this Model is not available for export.
         """
@@ -2129,11 +2131,10 @@ class Model(base.AiPlatformResourceNounWithFutureManager):
         if not self.supported_export_formats:
             raise ValueError(f"The model `{self.resource_name}` is not exportable.")
 
-        # Both or neither provided
-        if bool(artifact_destination) == bool(image_destination):
+        # No destination provided
+        if not any((artifact_destination, image_destination)):
             raise ValueError(
-                "Please provide either `artifact_destination` or `image_destination` "
-                "but not both."
+                "Please provide an `artifact_destination` or `image_destination`."
             )
 
         export_format_id = export_format_id.lower()
--- a/google/cloud/aiplatform/models.py
+++ b/google/cloud/aiplatform/models.py
@@ -39,7 +39,7 @@ from google.cloud.aiplatform.compat.types import (
     machine_resources as gca_machine_resources_compat,
     machine_resources_v1beta1 as gca_machine_resources_v1beta1,
     model as gca_model_compat,
-    model_service_v1 as gca_model_service_v1,
+    model_service as gca_model_service_v1,
     model_v1beta1 as gca_model_v1beta1,
     env_var as gca_env_var_compat,
     env_var_v1beta1 as gca_env_var_v1beta1,
@@ -1224,10 +1224,17 @@ class Model(base.AiPlatformResourceNounWithFutureManager):
         return self._gca_resource.description
 
     @property
-    def _supported_export_formats(
+    def supported_export_formats(
         self,
     ) -> Dict[str, List[gca_model_compat.Model.ExportFormat.ExportableContent]]:
-        """Store list of ExportFormat objects as a private dict for easy lookup"""
+        """The formats and content types in which this Model may be exported.
+        If empty, this Model is not available for export.
+
+        For example, if this model can be exported as a Tensorflow SavedModel and
+        have the artifacts written to Cloud Storage, the expected value would be:
+
+            {'tf-saved-model': [<ExportableContent.ARTIFACT: 1>]}
+        """
         return {
             export_format.id: [
                 gca_model_compat.Model.ExportFormat.ExportableContent(content)
@@ -1235,13 +1242,6 @@ class Model(base.AiPlatformResourceNounWithFutureManager):
             ]
             for export_format in self._gca_resource.supported_export_formats
         }
-
-    @property
-    def supported_export_formats(self) -> List[str]:
-        """The formats in which this Model may be exported.
-        If empty, this Model is not available for export.
-        """
-        return list(self._supported_export_formats.keys())
 
     def __init__(
         self,
@@ -2060,6 +2060,9 @@ class Model(base.AiPlatformResourceNounWithFutureManager):
         )
 
     @base.optional_sync()
+    def _wait_on_export(self, operation_future, sync=True) -> None:
+        operation_future.result()
+
     def export_model(
         self,
         export_format_id: str,
@@ -2067,7 +2070,7 @@ class Model(base.AiPlatformResourceNounWithFutureManager):
         image_destination: Optional[str] = None,
         sync: bool = True,
     ) -> None:
-        """Exports a trained, exportable, Model to a location specified by the user.
+        """Exports a trained, exportable Model to a location specified by the user.
         A Model is considered to be exportable if it has at least one `supported_export_formats`.
         Either `artifact_destination` or `image_destination` must be provided.
 
@@ -2089,8 +2092,6 @@ class Model(base.AiPlatformResourceNounWithFutureManager):
                 Required. The ID of the format in which the Model must be exported.
                 Each Model lists the export formats it supports, which can be
                 found using `Model.supported_export_formats`.
-                If no value is provided here, then the first from the list
-                of the Model's supported formats is used by default.
             artifact_destination (str):
                 The Cloud Storage location where the Model artifact is to be
                 written to. Under the directory given as the destination a
@@ -2098,10 +2099,10 @@ class Model(base.AiPlatformResourceNounWithFutureManager):
                 "``model-export-<model-display-name>-<timestamp-of-export-call>``",
                 where timestamp is in YYYY-MM-DDThh:mm:ss.sssZ ISO-8601
                 format, will be created. Inside, the Model and any of its
-                supporting files will be written. This field should only be
-                set when the ``exportableContent`` field of the
-                [Model.supported_export_formats] object contains
-                ``ARTIFACT``.
+                supporting files will be written.
+
+                This field should only be set when, in [Model.supported_export_formats],
+                the value for the key given in `export_format_id` contains ``ARTIFACT``.
             image_destination (str):
                 The Google Container Registry or Artifact Registry URI where
                 the Model container image will be copied to. Accepted forms:
@@ -2112,12 +2113,10 @@ class Model(base.AiPlatformResourceNounWithFutureManager):
                 -  Artifact Registry path. For example:
                 ``us-central1-docker.pkg.dev/projectId/repoName/imageName:tag``.
 
-                This field
-                should only be set when the ``exportableContent`` field of
-                the [Model.supported_export_formats] object contains
-                ``IMAGE``.
+                This field should only be set when, in [Model.supported_export_formats],
+                the value for the key given in `export_format_id` contains ``IMAGE``.
             sync (bool):
-                Whether to execute this deletion synchronously. If False, this method
+                Whether to execute this export synchronously. If False, this method
                 will be executed in concurrent Future and any downstream object will
                 be immediately returned and synced when the Future has completed.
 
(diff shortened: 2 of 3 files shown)


google/cloud/aiplatform/models.py
```diff
@@ -1221,6 +1223,24 @@ def description(self):
         """Description of the model."""
         return self._gca_resource.description
 
+    @property
+    def _supported_export_formats(self):
```
sasha-gitg, line 1227: type hint

google/cloud/aiplatform/models.py
```diff
@@ -1221,6 +1223,24 @@ def description(self):
...
+        return {
+            ef.id: [
+                gca_model_compat.Model.ExportFormat.ExportableContent(content)
+                for content in ef.exportable_contents
+            ]
+            for ef in self._gca_resource.supported_export_formats
+        }
+
+    @property
+    def supported_export_formats(self):
```
sasha-gitg, line 1238: type hint

google/cloud/aiplatform/models.py
```diff
@@ -1221,6 +1223,24 @@ def description(self):
...
 
+    @property
+    def _supported_export_formats(self):
+        """Store list of ExportFormat objects as a private dict for easy lookup"""
+        return {
+            ef.id: [
+                gca_model_compat.Model.ExportFormat.ExportableContent(content)
+                for content in ef.exportable_contents
+            ]
+            for ef in self._gca_resource.supported_export_formats
```
sasha-gitg, line 1234: Prefer a more descriptive variable name.

google/cloud/aiplatform/models.py
```diff
@@ -2036,3 +2056,139 @@ def list(
...
+
+            ValueError if invalid arguments or export formats are provided.
+        """
+
+        # Model does not support exporting
+        if not self.supported_export_formats:
+            raise ValueError(f"The model `{self.resource_name}` is not exportable.")
+
+        # Both or neither provided
+        if bool(artifact_destination) == bool(image_destination):
```
sasha-gitg, line 2133: I wonder if it makes sense to attempt to export to both destinations if given both arguments.

google/cloud/aiplatform/base.py
```diff
@@ -541,6 +541,14 @@ def update_time(self) -> datetime.datetime:
     def __repr__(self) -> str:
         return f"{object.__repr__(self)} \nresource name: {self.resource_name}"
 
+    def __getattr__(self, attr) -> Any:
```
sasha-gitg, line 544: Since this is an exposed API, I would consider removing this and, instead, explicitly add attributes with documentation. A possible drop in replacement is a `gca_resource` property, perhaps with a more descriptive name.

tests/unit/aiplatform/test_models.py
```diff
@@ -1131,3 +1189,104 @@ def test_print_model_if_exception(self):
             repr(test_model)
             == f"{object.__repr__(test_model)} failed with {str(mock_exception)}"
         )
+
+    @pytest.mark.usefixtures("get_model_with_supported_export_formats_artifact")
```
sasha-gitg, line 1193: Please parameterize with both `sync=True` and  `sync=False` path on all these tests.

sasha-gitg (changes requested):

google/cloud/aiplatform/models.py
vinnysenthil: Added return type hint

google/cloud/aiplatform/models.py
vinnysenthil: Added return type hint

google/cloud/aiplatform/models.py
vinnysenthil: `ef` -> `export_format`

google/cloud/aiplatform/models.py
vinnysenthil: Confirmed that both exporting to both destinations is a valid CUJ, updating the validation + logic to reflect that.

tests/unit/aiplatform/test_models.py
vinnysenthil: Done

google/cloud/aiplatform/base.py
vinnysenthil: Dropping after further discussion, we'll stick to exposing class properties individually

google/cloud/aiplatform/models.py
```diff
@@ -2036,3 +2058,138 @@ def list(
...
+
+    @base.optional_sync()
+    def export_model(
+        self,
+        export_format_id: str,
+        artifact_destination: Optional[str] = None,
+        image_destination: Optional[str] = None,
+        sync: bool = True,
+    ) -> None:
+        """Exports a trained, exportable, Model to a location specified by the user.
```
sasha-gitg, line 2070:
```suggestion
        """Exports a trained, exportable Model to a location specified by the user.
```

google/cloud/aiplatform/models.py
```diff
@@ -2036,3 +2058,138 @@ def list(
...
+                export_format_id='custom-model'
+                image_destination='us-central1-docker.pkg.dev/projectId/repo/image'
+            )
+
+        Args:
+            export_format_id (str):
+                Required. The ID of the format in which the Model must be exported.
+                Each Model lists the export formats it supports, which can be
+                found using `Model.supported_export_formats`.
+                If no value is provided here, then the first from the list
```
sasha-gitg, line 2092: This line doesn't seem to make sense because the argument is required.

google/cloud/aiplatform/models.py
```diff
@@ -2036,3 +2058,138 @@ def list(
...
+                The Google Container Registry or Artifact Registry URI where
+                the Model container image will be copied to. Accepted forms:
+
+                -  Google Container Registry path. For example:
+                ``gcr.io/projectId/imageName:tag``.
+
+                -  Artifact Registry path. For example:
+                ``us-central1-docker.pkg.dev/projectId/repoName/imageName:tag``.
+
+                This field
```
sasha-gitg, line 2115: Same comment as above.

google/cloud/aiplatform/models.py
```diff
@@ -2036,3 +2058,138 @@ def list(
...
+                of the Model's supported formats is used by default.
+            artifact_destination (str):
+                The Cloud Storage location where the Model artifact is to be
+                written to. Under the directory given as the destination a
+                new one with name
+                "``model-export-<model-display-name>-<timestamp-of-export-call>``",
+                where timestamp is in YYYY-MM-DDThh:mm:ss.sssZ ISO-8601
+                format, will be created. Inside, the Model and any of its
+                supporting files will be written. This field should only be
+                set when the ``exportableContent`` field of the
```
sasha-gitg, line 2102: I'm not sure what this means because `supported_export_formats` is a list of strings and doesn't seem to have an `exportableContent` field.

google/cloud/aiplatform/models.py
```diff
@@ -2036,3 +2058,138 @@ def list(
...
+
+                -  Artifact Registry path. For example:
+                ``us-central1-docker.pkg.dev/projectId/repoName/imageName:tag``.
+
+                This field
+                should only be set when the ``exportableContent`` field of
+                the [Model.supported_export_formats] object contains
+                ``IMAGE``.
+            sync (bool):
+                Whether to execute this deletion synchronously. If False, this method
```
sasha-gitg, line 2120:
```suggestion
                Whether to execute this export synchronously. If False, this method
```

google/cloud/aiplatform/models.py
```diff
@@ -2036,3 +2058,138 @@ def list(
...
+
+        content_types = gca_model_compat.Model.ExportFormat.ExportableContent
+        supported_content_types = self._supported_export_formats[export_format_id]
+
+        if (
+            artifact_destination
+            and content_types.ARTIFACT not in supported_content_types
+        ):
+            raise ValueError(
+                "This model can not be exported as an artifact. Try exporting as "
```
sasha-gitg, line 2157: Should this be qualified further with export_format_id of this model not being exportable as an artifact?

google/cloud/aiplatform/models.py
```diff
@@ -2036,3 +2058,138 @@ def list(
...
+                "a container image by passing the `image_destination` argument."
+            )
+
+        if image_destination and content_types.IMAGE not in supported_content_types:
+            raise ValueError(
+                "This model can not be exported as a container image. Try exporting "
```
sasha-gitg, line 2163: Same comment as above.

google/cloud/aiplatform/models.py
```diff
@@ -2036,3 +2058,138 @@ def list(
...
+                "the model artifacts by passing a `artifact_destination` argument."
+            )
+
+        # Construct request payload
+        output_config = gca_model_service_v1.ExportModelRequest.OutputConfig(
```
sasha-gitg, line 2168: Preference to use compat instead of v1 directly for these protos.

google/cloud/aiplatform/models.py
```diff
@@ -2036,3 +2058,138 @@ def list(
...
+        )
+
+        _LOGGER.log_action_started_against_resource_with_lro(
+            "Export", "model", self.__class__, operation_future
+        )
+
+        # Block before returning
+        operation_future.result()
+
+        _LOGGER.log_action_completed_against_resource("model", "exported", self)
```
sasha-gitg, line 2195: We should reconsider not returning here. For the export artifact to gcs case, it creates a new directory in the given directory and the user will not have a reference to that directory.

tests/unit/aiplatform/test_models.py
```diff
@@ -1131,3 +1212,170 @@ def test_print_model_if_exception(self):
             repr(test_model)
             == f"{object.__repr__(test_model)} failed with {str(mock_exception)}"
         )
+
+    @pytest.mark.parametrize("sync", [True, False])
```
sasha-gitg, line 1216: Tests look good. Thanks!

sasha-gitg (changes requested):

google/cloud/aiplatform/models.py
vinnysenthil: Yes, good point!

google/cloud/aiplatform/models.py
vinnysenthil: Perhaps we should expose `_supported_export_formats` directly instead of `supported_export_formats` in order for the user to see the format + content type pairings for themselves. Alternatively I can delete that portion of the docstrings for now.

google/cloud/aiplatform/models.py
vinnysenthil: This is being imported from `aiplatform.compat.types`, [see here](https://github.com/googleapis/python-aiplatform/pull/353/files#diff-72d6e2818d2224db618771b26094747754d49a2acf33160eaf017ff1b6ddf840R42). Is this not the correct way to access compat types?

google/cloud/aiplatform/models.py
vinnysenthil: Should we return a storage `Blob` to the new dir in the event of a GCS export?

google/cloud/aiplatform/models.py
sasha-gitg: We should try to import from here to pin to the default version unless there is a reason to control the version in the implementation: https://github.com/googleapis/python-aiplatform/blob/master/google/cloud/aiplatform/compat/__init__.py

google/cloud/aiplatform/models.py
sasha-gitg: I think the usage pattern would be to load the model using the gcs uri after export instead of interacting with the Blob. For example, loading a TF model or passing the uri downstream. I think the main advantage of the blob is that it provides an API to download the file locally but would make the other use cases a bit more laborious.

There are also additional considerations for this API when converted to a Pipeline Component in that a Blob object is not a supported output. 

I'm leaning towards returning ExportModelOperationMetadata.OutputInfo where the key is only present if the output destination is present. 

The main issue with this approach is that our async support only supports returning either a resource noun or None. So we can either remove the sync support for now or you can split the call like this:


```Python
def export(..., sync=True) -> Dict[str, str]:
  ...
  operation_future = self.api_client.export_model(
            name=self.resource_name, output_config=output_config
        )
        
   self._wait_on_export(operation_future, sync=sync)
   
   return convert_to_dict(operation_future.operation.OutputInfo)

@base.optional_sync
def _wait_on_export(self, operation_future, sync=True) -> None:
   operation.result()
```  

Let me know your thoughts.

google/cloud/aiplatform/models.py
sasha-gitg: Is format + content required use this API effectively or is format enough?

vinnysenthil: I think both are required to use the API effectively since you wouldn't know whether a particular export format can be saved as an image, artifact, or both. I'll go ahead and expose this in `supported_export_formats` and get rid of `_supported_export_formats` if there's no objection

sasha-gitg: SGTM

google/cloud/aiplatform/models.py
vinnysenthil: Updating
```py
from google.cloud.aiplatform.compat.types import model_service_v1
```
to
```py
from google.cloud.aiplatform.compat.types import model_service
```

google/cloud/aiplatform/models.py
vinnysenthil: Clarifying to check `Model.supported_export_types`

google/cloud/aiplatform/models.py
vinnysenthil: Implemented suggested approach

google/cloud/aiplatform/models.py
```diff
@@ -2036,3 +2058,139 @@ def list(
...
+
+            or
+
+            my_model.export(
+                export_format_id='custom-model'
+                image_destination='us-central1-docker.pkg.dev/projectId/repo/image'
+            )
+
+        Args:
+            export_format_id (str):
```
sasha-gitg, line 2091: Please add the value ids currently supported.

google/cloud/aiplatform/models.py
```diff
@@ -2036,3 +2058,139 @@ def list(
...
+    def _wait_on_export(self, operation_future, sync=True) -> None:
+        operation_future.result()
+
+    def export_model(
+        self,
+        export_format_id: str,
+        artifact_destination: Optional[str] = None,
+        image_destination: Optional[str] = None,
+        sync: bool = True,
+    ) -> None:
```
sasha-gitg, line 2072:
```suggestion
    ) -> Dict[str, str]:
```

google/cloud/aiplatform/models.py
```diff
@@ -2036,3 +2058,139 @@ def list(
...
+                -  Artifact Registry path. For example:
+                ``us-central1-docker.pkg.dev/projectId/repoName/imageName:tag``.
+
+                This field should only be set when, in [Model.supported_export_formats],
+                the value for the key given in `export_format_id` contains ``IMAGE``.
+            sync (bool):
+                Whether to execute this export synchronously. If False, this method
+                will be executed in concurrent Future and any downstream object will
+                be immediately returned and synced when the Future has completed.
+
```
sasha-gitg, line 2122: Returns section. It should describe what the user should expect.

google/cloud/aiplatform/models.py
```diff
@@ -2036,3 +2058,139 @@ def list(
...
+        _LOGGER.log_action_started_against_resource_with_lro(
+            "Export", "model", self.__class__, operation_future
+        )
+
+        # Block before returning
+        self._wait_on_export(operation_future=operation_future, sync=sync)
+
+        _LOGGER.log_action_completed_against_resource("model", "exported", self)
+
+        return json_format.MessageToDict(operation_future.metadata.output_info._pb)
```
sasha-gitg, line 2196: If I only provide one output target do I still get both keys in the dict and if so, what is the value for the key representing the output that wasn't set?

sasha-gitg (changes requested):

google/cloud/aiplatform/models.py
vinnysenthil: The only keys in the dict are for the output targets that were set in the export model request.

```py

m = aiplatform.Model('8842430840248991744')

m.export_model(
    export_format_id='tf-saved-model',
    artifact_destination='gs://vinnys-mbsdk-staging-uscentral1/'
)
```
returns
```py
{'artifactOutputUri': 'gs://vinnys-mbsdk-staging-uscentral1/model-8842430840248991744/tf-saved-model/2021-04-30T17:22:41.298770Z'}
```

google/cloud/aiplatform/models.py
```diff
@@ -2036,3 +2058,139 @@ def list(
             location=location,
             credentials=credentials,
         )
+
+    @base.optional_sync()
+    def _wait_on_export(self, operation_future, sync=True) -> None:
```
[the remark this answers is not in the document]
vinnysenthil: Done

google/cloud/aiplatform/models.py
vinnysenthil: Done

google/cloud/aiplatform/models.py
vinnysenthil: Usually a particular Model only has one valid export format id, think it might be confusing to the user to list all possible export IDs. It would be best if they checked `Model.supported_export_formats` first. What do you think?

google/cloud/aiplatform/models.py
vinnysenthil: Done

google/cloud/aiplatform/models.py
sasha-gitg: SGTM

sasha-gitg (approved):

PR #357, opened 2021-04-29
feat: Added default AutoMLTabularTrainingJob column transformations

ivanmkc:
- [x] Functionality to default to all columns for GCS sources
- [x] Functionality to default to all columns for BQ sources
- [x] Units tests
- [x] Check if target column needs to be omitted
- [x] Test with live BQ datasources
- [x] Test with live GCS datasources
- [x] Test with end-to-end CUJ

### Testing code
```
# Train
from google.cloud.aiplatform import training_jobs

ds = aiplatform.TabularDataset.create(display_name="sample-gsod", bq_source="bq://bigquery-public-data.samples.gsod")

job = training_jobs.AutoMLTabularTrainingJob(
    display_name="tabular_regression_job_test",
    optimization_objective="minimize-rmse",
    optimization_prediction_type="regression",
    column_transformations=None,
    optimization_objective_recall_value=None,
    optimization_objective_precision_value=None,
)

model_from_job = job.run(
    dataset=ds,
    target_column="mean_temp",
    model_display_name="tabular_model",
    sync=True,
)
```

fixes: https://b.corp.google.com/issues/181042526

### DIFF
--- a/google/cloud/aiplatform/datasets/_datasources.py
+++ b/google/cloud/aiplatform/datasets/_datasources.py
@@ -86,9 +86,9 @@ class TabularDatasource(Datasource):
             raise ValueError("One of gcs_source or bq_source must be set.")
 
         if gcs_source:
-            dataset_metadata = {"input_config": {"gcs_source": {"uri": gcs_source}}}
+            dataset_metadata = {"inputConfig": {"gcsSource": {"uri": gcs_source}}}
         elif bq_source:
-            dataset_metadata = {"input_config": {"bigquery_source": {"uri": bq_source}}}
+            dataset_metadata = {"inputConfig": {"bigquerySource": {"uri": bq_source}}}
 
         self._dataset_metadata = dataset_metadata
 
--- a/google/cloud/aiplatform/datasets/tabular_dataset.py
+++ b/google/cloud/aiplatform/datasets/tabular_dataset.py
@@ -41,13 +41,13 @@ class TabularDataset(datasets._Dataset):
 
     @property
     def column_names(self) -> List[str]:
-        input_config = self._gca_resource.metadata.get("input_config")
+        input_config = self._gca_resource.metadata.get("inputConfig")
 
         if input_config is None:
-            raise RuntimeError("No input_config found for dataset")
+            raise RuntimeError("No inputConfig found for dataset")
 
-        gcs_source = input_config.get("gcs_source")
-        bq_source = input_config.get("bigquery_source")
+        gcs_source = input_config.get("gcsSource")
+        bq_source = input_config.get("bigquerySource")
 
         if gcs_source:
             gcs_source_uris = gcs_source.get("uri")
--- a/google/cloud/aiplatform/datasets/tabular_dataset.py
+++ b/google/cloud/aiplatform/datasets/tabular_dataset.py
@@ -42,13 +42,13 @@ class TabularDataset(datasets._Dataset):
 
     @property
     def column_names(self) -> List[str]:
-        """Retrieve the columns for the dataset by extracting it from the Google Cloud Storage or 
+        """Retrieve the columns for the dataset by extracting it from the Google Cloud Storage or
         Google BigQuery source.
 
         Returns:
             List[str]
                 A list of columns names
-                
+
         Raises:
             RuntimeError: When no valid source is found.
         """
@@ -109,7 +109,7 @@ class TabularDataset(datasets._Dataset):
         Returns:
             List[str]
                 A list of columns names in the CSV file.
-                
+
         Raises:
             RuntimeError: When the retrieved CSV file is invalid.
         """
@@ -142,9 +142,11 @@ class TabularDataset(datasets._Dataset):
             header_line = header_line.split("\n")[:1]
 
             csv_reader = csv.reader(header_line, delimiter=",")
-        except:
+        except (ValueError, RuntimeError) as err:
             raise RuntimeError(
-                f"There was a problem extracting the headers from the CSV file at: { gcs_csv_file_path }"
+                "There was a problem extracting the headers from the CSV file at '{}': {}".format(
+                    gcs_csv_file_path, err
+                )
             )
         finally:
             logging.disable(logging.NOTSET)
@@ -168,7 +170,7 @@ class TabularDataset(datasets._Dataset):
             project (str):
                 Required. Project to initiate the BigQuery client with.
             bq_table_uri (str):
-                Required. A URI to a BigQuery table. 
+                Required. A URI to a BigQuery table.
                 Can include "bq://" prefix but not required.
 
         Returns:
--- a/google/cloud/aiplatform/datasets/tabular_dataset.py
+++ b/google/cloud/aiplatform/datasets/tabular_dataset.py
@@ -41,7 +41,23 @@ class TabularDataset(datasets._Dataset):
 
     @property
     def column_names(self) -> List[str]:
-        input_config = self._gca_resource.metadata.get("inputConfig")
+        """Retrieve the columns for the dataset by extracting it from the Google Cloud Storage or 
+        Google BigQuery source.
+
+        Returns:
+            List[str]
+                A list of columns names
+                
+        Raises:
+            RuntimeError: When no valid source is found.
+        """
+
+        metadata = self._gca_resource.metadata
+
+        if metadata is None:
+            raise RuntimeError("No metadata found for dataset")
+
+        input_config = metadata.get("inputConfig")
 
         if input_config is None:
             raise RuntimeError("No inputConfig found for dataset")
@@ -92,8 +108,6 @@ class TabularDataset(datasets._Dataset):
                 Must include "gs://" prefix.
 
         Returns:
-            str
-
             List[str]
                 A list of columns names in the CSV file.
                 
--- a/google/cloud/aiplatform/datasets/tabular_dataset.py
+++ b/google/cloud/aiplatform/datasets/tabular_dataset.py
@@ -62,30 +62,38 @@ class TabularDataset(datasets._Dataset):
                 )
         elif bq_source:
             bq_table_uri = bq_source.get("uri")
-            return TabularDataset._retrieve_bq_source_columns(bq_table_uri)
+            if bq_table_uri:
+                return TabularDataset._retrieve_bq_source_columns(
+                    self.project, bq_table_uri
+                )
 
         raise RuntimeError("No valid CSV or BigQuery datasource found.")
 
     @classmethod
     def _retrieve_gcs_source_columns(
-        cls, project_id: str, gcs_csv_file_path: str
+        cls, project: str, gcs_csv_file_path: str
     ) -> List[str]:
         """Retrieve the columns from a comma-delimited CSV file stored on Google Cloud Storage
 
         Example Usage:
 
             column_names = _retrieve_gcs_source_columns(
+                "project_id",
                 "gs://example-bucket/path/to/csv_file"
             )
 
             # column_names = ["column_1", "column_2"]
 
         Args:
+            project (str):
+                Required. Project to initiate the Google Cloud Storage client with.
             gcs_csv_file_path (str):
                 Required. A full path to a CSV files stored on Google Cloud Storage.
                 Must include "gs://" prefix.
 
         Returns:
+            str
+
             List[str]
                 A list of columns names in the CSV file.
                 
@@ -128,18 +136,21 @@ class TabularDataset(datasets._Dataset):
         return next(csv_reader)
 
     @classmethod
-    def _retrieve_bq_source_columns(cls, bq_table_uri: str) -> List[str]:
+    def _retrieve_bq_source_columns(cls, project: str, bq_table_uri: str) -> List[str]:
         """Retrieve the columns from a table on Google BigQuery
 
         Example Usage:
 
             column_names = _retrieve_bq_source_columns(
+                "project_id",
                 "bq://project_id.dataset.table"
             )
 
             # column_names = ["column_1", "column_2"]
 
         Args:
+            project (str):
+                Required. Project to initiate the BigQuery client with.
             bq_table_uri (str):
                 Required. A URI to a BigQuery table. 
                 Can include "bq://" prefix but not required.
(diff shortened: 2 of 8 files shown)


google/cloud/aiplatform/datasets/_datasources.py
```diff
@@ -86,9 +86,9 @@ def __init__(
             raise ValueError("One of gcs_source or bq_source must be set.")
 
         if gcs_source:
-            dataset_metadata = {"input_config": {"gcs_source": {"uri": gcs_source}}}
```
ivanmkc, line 89: These were wrong before

tests/unit/aiplatform/test_datasets.py
```diff
@@ -176,7 +178,22 @@ def get_dataset_image_mock():
...
 @pytest.fixture
-def get_dataset_tabular_mock():
+def get_dataset_tabular_gcs_mock():
+    with patch.object(
+        dataset_service_client.DatasetServiceClient, "get_dataset"
+    ) as get_dataset_mock:
+        get_dataset_mock.return_value = gca_dataset.Dataset(
+            display_name=_TEST_DISPLAY_NAME,
+            metadata_schema_uri=_TEST_METADATA_SCHEMA_URI_TABULAR,
+            metadata=_TEST_METADATA_TABULAR_GCS,
```
ivanmkc, line 188: Mock with GCS source

tests/unit/aiplatform/test_datasets.py
```diff
@@ -190,6 +207,21 @@ def get_dataset_tabular_mock():
...
 
+@pytest.fixture
+def get_dataset_tabular_missing_metadata_mock():
+    with patch.object(
+        dataset_service_client.DatasetServiceClient, "get_dataset"
+    ) as get_dataset_mock:
+        get_dataset_mock.return_value = gca_dataset.Dataset(
+            display_name=_TEST_DISPLAY_NAME,
+            metadata_schema_uri=_TEST_METADATA_SCHEMA_URI_TABULAR,
+            metadata=None,
```
ivanmkc, line 218: Mock with no metadata

google/cloud/aiplatform/training_jobs.py
```diff
@@ -2918,10 +2917,18 @@ def _run(
 
         training_task_definition = schema.training_job.definition.automl_tabular
 
+        if self._column_transformations is None:
+            column_transformations = [
+                {"auto": {"column_name": column_name}}
+                for column_name in dataset.column_names()
```
ivanmkc, line 2923: I wanted to use just `column_names` instead of `column_names()` because it's a property but I couldn't get the unit-test mock working for the life of me.

google/cloud/aiplatform/datasets/tabular_dataset.py
```diff
@@ -33,6 +39,151 @@ class TabularDataset(datasets._Dataset):
...
+
+        metadata = self._gca_resource.metadata
+
+        if metadata is None:
+            raise RuntimeError("No metadata found for dataset")
+
+        input_config = metadata.get("inputConfig")
+
+        if input_config is None:
+            raise RuntimeError("No inputConfig found for dataset")
```
ivanmkc, line 63: This line has no code coverage. Let me know if that matters.

google/cloud/aiplatform/datasets/tabular_dataset.py
```diff
@@ -33,6 +39,151 @@ class TabularDataset(datasets._Dataset):
...
+                    self.project, gcs_source_uris[0]
+                )
+        elif bq_source:
+            bq_table_uri = bq_source.get("uri")
+            if bq_table_uri:
+                return TabularDataset._retrieve_bq_source_columns(
+                    self.project, bq_table_uri
+                )
+
+        raise RuntimeError("No valid CSV or BigQuery datasource found.")
```
ivanmkc, line 86: This line has no testing coverage. Let me know if that matters.

google/cloud/aiplatform/datasets/tabular_dataset.py
```diff
@@ -33,6 +39,151 @@ class TabularDataset(datasets._Dataset):
...
+                ).decode("utf-8")
+                first_new_line_index = line.find("\n")
+                start_index += increment
+
+            header_line = line[:first_new_line_index]
+
+            # Split to make it an iterable
+            header_line = header_line.split("\n")
+
+            csv_reader = csv.reader(header_line, delimiter=",")
```
ivanmkc, line 144: Leverage  CSV to read the csv format

google/cloud/aiplatform/datasets/tabular_dataset.py
```diff
@@ -33,6 +39,151 @@ class TabularDataset(datasets._Dataset):
...
+            bq_table_uri = bq_source.get("uri")
+            if bq_table_uri:
+                return TabularDataset._retrieve_bq_source_columns(
+                    self.project, bq_table_uri
+                )
+
+        raise RuntimeError("No valid CSV or BigQuery datasource found.")
+
+    @classmethod
+    def _retrieve_gcs_source_columns(
```
sasha-gitg, line 89: These next two methods seem like they can be static methods.

google/cloud/aiplatform/datasets/tabular_dataset.py
```diff
@@ -33,6 +39,151 @@ class TabularDataset(datasets._Dataset):
         schema.dataset.metadata.tabular,
     )
 
+    @property
+    def column_names(self) -> List[str]:
```
sasha-gitg, line 43: I get this log:
```INFO:google.resumable_media._helpers:No MD5 checksum was returned from the service while downloading```

source; https://github.com/googleapis/google-resumable-media-python/blob/master/google/resumable_media/_helpers.py#L39

when using this property. Can we suppress or resolve this log?

File: google/cloud/aiplatform/training_jobs.py [hunk 10]
```diff
@@ -2918,10 +2917,19 @@ def _run(
 
         training_task_definition = schema.training_job.definition.automl_tabular
 
+        if self._column_transformations is None:
```
sasha-gitg, line 2920: Please log here we are defaulting to auto for all columns as column_transformations was not provided.

sasha-gitg (changes requested):
Looks great. Tested column_names and it works nicely.  

Left a few requested changes. Thanks!

google/cloud/aiplatform/datasets/tabular_dataset.py
```diff
@@ -33,6 +39,151 @@ class TabularDataset(datasets._Dataset):
...
+                line += blob.download_as_bytes(
+                    start=start_index, end=start_index + increment
+                ).decode("utf-8")
+                first_new_line_index = line.find("\n")
+                start_index += increment
+
+            header_line = line[:first_new_line_index]
+
+            # Split to make it an iterable
+            header_line = header_line.split("\n")
```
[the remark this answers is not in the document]
ivanmkc: Sure, will do.

google/cloud/aiplatform/datasets/tabular_dataset.py
ivanmkc: It is generally preferred in Python to use static where possible, then class and then method?

google/cloud/aiplatform/datasets/tabular_dataset.py
ivanmkc: Will look into it.

google/cloud/aiplatform/training_jobs.py
ivanmkc: Makes sense, will add.

google/cloud/aiplatform/datasets/tabular_dataset.py
sasha-gitg: In general, though there's probably exceptions, static methods are preferred over class methods when the method doesn't interact with the class. In this case it looks like `cls` is unused. The main reason to choose a static method over a standalone method is to indicate that the method is logically coupled to the class. That seems to make sense in this case because retrieving columns names would is unique to tabular datasets. If we find we need this functionality elsewhere in the SDK then we should implement these methods as standalone method.

ivanmkc: Sounds good.

ivanmkc: Done

google/cloud/aiplatform/training_jobs.py
ivanmkc: Fixed it.

google/cloud/aiplatform/datasets/tabular_dataset.py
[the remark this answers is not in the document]
ivanmkc: Done

google/cloud/aiplatform/datasets/tabular_dataset.py
ivanmkc: I don't get this log, do you have repro steps?

google/cloud/aiplatform/training_jobs.py
ivanmkc: Done

ivanmkc:
```
INFO:google.cloud.aiplatform.training_jobs:No column transformations provided, so now retrieving columns from dataset in order to set default column transformations.
INFO:google.cloud.aiplatform.training_jobs:The column transformation of type 'auto' was set for the following columns': ['station_number', 'wban_number', 'year', 'month', 'day', 'num_mean_temp_samples', 'mean_dew_point', 'num_mean_dew_point_samples', 'mean_sealevel_pressure', 'num_mean_sealevel_pressure_samples', 'mean_station_pressure', 'num_mean_station_pressure_samples', 'mean_visibility', 'num_mean_visibility_samples', 'mean_wind_speed', 'num_mean_wind_speed_samples', 'max_sustained_wind_speed', 'max_gust_wind_speed', 'max_temperature', 'max_temperature_explicit', 'min_temperature', 'min_temperature_explicit', 'total_precipitation', 'snow_depth', 'fog', 'rain', 'snow', 'hail', 'thunder', 'tornado'].
```

google/cloud/aiplatform/datasets/tabular_dataset.py
sasha-gitg:
```Python
for ds in aiplatform.TabularDataset.list():
  print(ds.column_names)
```

Might be specific to the version of the library. I see that log with:
```
google-cloud-storage==1.35.1
google-resumable-media==1.2.0
```

What versions are you testing with?

google/cloud/aiplatform/datasets/tabular_dataset.py
ivanmkc: Fixed

google/cloud/aiplatform/datasets/tabular_dataset.py
ivanmkc: Fixed

google/cloud/aiplatform/training_jobs.py
ivanmkc: @sasha-gitg Does this look okay or is it too verbose?

ivanmkc: I thought it would be nice to show the names so the user can verify the columns.

google/cloud/aiplatform/datasets/tabular_dataset.py
ivanmkc: I could reproduce it with on GCS datasets.

I added code to suppress logging during the GCS download.

google/cloud/aiplatform/datasets/tabular_dataset.py
ivanmkc: Hm it's a little hard to define "compatible".

One way to define compatible is to make sure it matches the exact behaviour as the service code. That would be hard to keep in sync as it can change over time.

I think it's best to add a note in the docstring that only CSV's formatted in the default ("excel") dialect are supported.

@sasha-gitg what do you think?

google/cloud/aiplatform/training_jobs.py
sasha-gitg: LGTM

google/cloud/aiplatform/datasets/tabular_dataset.py
ivanmkc: Here are the docs in any case: https://cloud.google.com/ai-platform-unified/docs/datasets/prepare-tabular?hl=en#csv

My thinking is that, the user wouldn't be able to make a dataset in the first place if their data was in the wrong format (it should error out at dataset creation, but not sure).

So we can assume that, when they call `column_names` their data is already comma-delimited.
That's why I used `csv.reader(header_line, delimiter=",")`

sasha-gitg: SGTM. No need to add the dialect to the docstring.

google/cloud/aiplatform/datasets/tabular_dataset.py
```diff
@@ -15,16 +15,23 @@
...
+from google.cloud import bigquery
+from google.cloud import storage
+
 from google.cloud.aiplatform import datasets
 from google.cloud.aiplatform.datasets import _datasources
 from google.cloud.aiplatform import initializer
 from google.cloud.aiplatform import schema
 from google.cloud.aiplatform import utils
 
+from typing import List
```
sasha-gitg, line 32: This should be imported above with the rest of the types.

google/cloud/aiplatform/datasets/tabular_dataset.py
```diff
@@ -33,6 +40,154 @@ class TabularDataset(datasets._Dataset):
...
+        blob = bucket.blob(gcs_blob)
+
+        # Incrementally download the CSV file until the header is retrieved
+        first_new_line_index = -1
+        start_index = 0
+        increment = 1000
+        line = ""
+
+        try:
+            logging.disable(logging.CRITICAL)
```
sasha-gitg, line 131: Preference to be more precise by filtering the module logs we're trying to suppress. Like so:

https://github.com/googleapis/python-aiplatform/blob/857f63d475463189ebb89b25d0ca08d9544c3bf3/google/cloud/aiplatform/initializer.py#L172-L176

google/cloud/aiplatform/datasets/tabular_dataset.py
```diff
@@ -15,16 +15,23 @@
...
+from google.cloud import storage
+
 from google.cloud.aiplatform import datasets
 from google.cloud.aiplatform.datasets import _datasources
 from google.cloud.aiplatform import initializer
 from google.cloud.aiplatform import schema
 from google.cloud.aiplatform import utils
 
+from typing import List
+import logging
```
sasha-gitg, line 33: Import should be up top with stdlib imports.

sasha-gitg (approved):
Looks great. Added a few comments. Thanks Ivan!

google/cloud/aiplatform/datasets/tabular_dataset.py
ivanmkc: Ah no lint check?

google/cloud/aiplatform/datasets/tabular_dataset.py
ivanmkc: Fixed

google/cloud/aiplatform/datasets/tabular_dataset.py
ivanmkc: Fixed

PR #359, opened 2021-04-30
fix(aiplatform): Fix doc formatting

### LINKED ISSUE #358: Docstring for a few of the methods are not parsable
Docstring for a few of the methods under aiplatform are not parsable, which breaks API doc generation and parsing tools. For example in training_jobs.py the docstring 

```python
    def _raise_failure(self):
        """Helper method to raise failure if TrainingPipeline fails.

        Raises:
            RuntimeError If training failed.
        """
```
should be changed to 
```python 
    def _raise_failure(self):
        """Helper method to raise failure if TrainingPipeline fails.

        Raises:
            RuntimeError: If training failed.
        """
```

SinaChavoshi:
Fixing docstring formatting for aiplatform.(Fixes #358)

### DIFF
--- a/google/cloud/aiplatform/helpers/value_converter.py
+++ b/google/cloud/aiplatform/helpers/value_converter.py
@@ -21,7 +21,8 @@ from proto.message import MessageMeta
 
 
 def to_value(self: Message) -> Value:
-    """Converts a message type to a :class:`~google.protobuf.struct_pb2.Value` object.
+    """Converts a message type to a :class:`~google.protobuf.struct_pb2.Value`
+    object.
 
     Args:
       message: the message to convert
@@ -34,7 +35,8 @@ def to_value(self: Message) -> Value:
 
 
 def from_value(cls: MessageMeta, value: Value) -> Message:
-    """Creates instance of class from a :class:`~google.protobuf.struct_pb2.Value` object.
+    """Creates instance of class from a
+    :class:`~google.protobuf.struct_pb2.Value` object.
 
     Args:
       value: a :class:`~google.protobuf.struct_pb2.Value` object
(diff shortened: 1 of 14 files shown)


google/cloud/aiplatform/helpers/value_converter.py
```diff
@@ -21,7 +21,8 @@
```
sasha-gitg, line 21: Please revert this file.

google/cloud/aiplatform/utils.py
```diff
@@ -318,14 +320,14 @@ def __init__(
...
 
-            client_class (AiPlatformServiceClient):
-                Required. Class of the client to use.
-            client_options (client_options.ClientOptions):
-                Required. Client options to pass to client.
-            client_info (gapic_v1.client_info.ClientInfo):
-                Required. Client info to pass to client.
-            credentials (auth_credentials.credentials):
-                Optional. Client credentials to pass to client.
+            client_class (AiPlatformServiceClient):     Required. Class
```
sasha-gitg, line 323: It looks like this one needs to be indented with an "Args:" section as it was missing it initially.

google/cloud/aiplatform/utils.py
```diff
@@ -367,12 +369,11 @@ def __init__(
     ):
         """Stores parameters needed to instantiate client.
 
-        client_options (client_options.ClientOptions):
-            Required. Client options to pass to client.
-        client_info (gapic_v1.client_info.ClientInfo):
-            Required. Client info to pass to client.
-        credentials (auth_credentials.credentials):
-            Optional. Client credentials to pass to client.
+        client_options (client_options.ClientOptions):     Required.
```
sasha-gitg, line 372: Same comment as above.

sasha-gitg (changes requested):

google/cloud/aiplatform/helpers/value_converter.py
SinaChavoshi: done.

sasha-gitg (approved):
Thanks Sina!

PR #361, opened 2021-05-04
Remove MB type, enforce metrics value type and add experiment description support

jialuzh:
This PR include the following 3 changes:
1. Remove experiment type from experiment context metadata.
2. Enforce metrics value type checking, only int and float is allowed.
3. Add support for adding experiment description through init call. Description can only be added or updated when experiment name and description are provided at the same time.


jialuzh:
@nachocano @sasha-gitg Please help review this PR with the changes that were discussed on Mon.

google/cloud/aiplatform/initializer.py
```diff
@@ -65,8 +66,9 @@ def init(
         Args:
             project (str): The default project to use when making API calls.
             location (str): The default location to use when making API calls. If not
-                set defaults to us-central-1
-            experiment (str): The experiment name
+                set defaults to us-central-1.
+            experiment (str): The experiment name.
+            experiment_description (str): description of an experiment.
```
sasha-gitg, line 71:
```suggestion
            experiment_description (str): The description of the experiment.
```

google/cloud/aiplatform/metadata/metadata.py
```diff
@@ -52,20 +52,29 @@ def run_name(self) -> Optional[str]:
...
+            description (str):
+                Optional. Description of an experiment.
         Raises:
             ValueError if a context with the same name as the experiment is create but with a different schema.
+            Or if exp_description is set but experiment is not.
         """
 
+        if experiment is None:
+            raise ValueError(
+                "Experiment name needs to be set in order to add experiment descriptions."
```
sasha-gitg, line 70: I think this exception may be more informative if it was moved up the call stack to init with remediation included ie. `Experiment needs to be set using aiplatform.init`

google/cloud/aiplatform/metadata/metadata.py
```diff
@@ -248,6 +265,24 @@ def _validate_experiment_and_run(self, method_name: str):
...
+                Required. Metrics key/value pairs. Only flot and int are supported format for value.
+        Raises:
+            TypeError if value contains unsupported types.
+        """
+
+        for key, value in metrics.items():
+            if isinstance(value, int) or isinstance(value, float):
+                continue
+            raise TypeError(
+                f"metrics contains unsupported value types. key: {key}; value: {value}"
```
sasha-gitg, line 283:
```suggestion
                f"metrics contain unsupported value types. key: {key}; value: {value}; type: {type(value)}"
```

sasha-gitg (approved):

google/cloud/aiplatform/metadata/metadata.py
jialuzh: I was originally thinking to support update description if experiment was previous set through init so that check logic would need to be in metadata.py, but later changed my mind on that..... Agree it is better to handle the exception in init with current logic. I also added a couple tests to cover the exception.

PR #364, opened 2021-05-04
feat(aiplatform): Add support for setting User agent header

SinaChavoshi:
Add class attributes to help with specifying user_agent header product and comments fields. This is used in metrics reporting.

### DIFF
--- a/google/cloud/aiplatform/initializer.py
+++ b/google/cloud/aiplatform/initializer.py
@@ -259,8 +259,8 @@ class _Config:
             "google-cloud-aiplatform",
         ).version
         client_info = gapic_v1.client_info.ClientInfo(
-            gapic_version=gapic_version, 
-            user_agent=f"{constants.USER_AGENT_PRODUCT}/{gapic_version}"
+            gapic_version=gapic_version,
+            user_agent=f"{constants.USER_AGENT_PRODUCT}/{gapic_version}",
         )
 
         kwargs = {
--- a/google/cloud/aiplatform/initializer.py
+++ b/google/cloud/aiplatform/initializer.py
@@ -42,6 +42,10 @@ from google.cloud.aiplatform.compat.types import (
 class _Config:
     """Stores common parameters and options for API calls."""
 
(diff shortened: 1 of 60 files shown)


google/cloud/aiplatform/initializer.py
```diff
@@ -259,15 +259,18 @@ def create_client(
             "google-cloud-aiplatform",
         ).version
         client_info = gapic_v1.client_info.ClientInfo(
-            gapic_version=gapic_version, user_agent=f"model-builder/{gapic_version}"
+            gapic_version=gapic_version,
+            user_agent=
+            f"{self.user_agent_product}/{gapic_version} ({self.user_agent_comments})"
```
sasha-gitg, line 264: Would like @dizcology input here on adding the user_agent_comment suffix. 

Will there be any issue here with converting the user_agent header for the default case from `"model_builder/v1"` to `"model_builder/v1 ()"`?

google/cloud/aiplatform/initializer.py
```diff
@@ -42,6 +41,10 @@
 class _Config:
     """Stores common parameters and options for API calls."""
 
+    # Class attribute to set user_agent header used for metrics reporting.
+    user_agent_product = "model-builder"
```
sasha-gitg, line 45: I have a preference for storing this under `constants.py`. So setting is a bit more direct and aligned with setting the `API_BASE_PATH`. ie:

```aiplatform.constants.USER_AGENT_PRODUCT = "other-product"```

google/cloud/aiplatform/initializer.py
SinaChavoshi: Since this is not immediately needed I will remove it for now, once we have a need to add additional metrics in comment field we can add this back.

google/cloud/aiplatform/initializer.py
SinaChavoshi: Done, thats a good call, thank you!

dizcology (approved):

sasha-gitg (approved):

PR #366, opened 2021-05-05
feat: expose env var in cust training class run func args

morgandu:
Fixes #<[b/186513106](https://b.corp.google.com/issues/186513106)> 🦕


sasha-gitg:
I think this is failing because, unlike `model_to_upload`, `training_task_inputs` is a struct.Value:

https://github.com/googleapis/python-aiplatform/blob/master/google/cloud/aiplatform_v1/types/training_pipeline.py#L161

 Try passing in a dictionary with the same schema. Something like::

```Python
spec["pythonPackageSpec"]["env"] = [
  {'name':key, 'value': value} for key, value in environment_variables.items()]
```

sasha-gitg (approved):
Thanks Morgan!

PR #369, opened 2021-05-05
chore: add extras dependencies and install for testing

sasha-gitg:


jialuzh (approved):

vinnysenthil (approved):

setup.py
```diff
@@ -29,6 +29,12 @@
 with io.open(readme_filename, encoding="utf-8") as readme_file:
     readme = readme_file.read()
 
+tensorboard_extra_require = ["tensorflow-cpu >= 2.0.0"]
```
yfang1, line 32: Actually this needs to be "tensorflow-cpu >= 2.3.0, < 2.5.0", since the TB uploader is only available after TB=2.3.0.

The TB team asks for the upper bound so we can possible make backwards incompatible changes in the future.

sasha-gitg: It looks like we need to support `<=2.5.0rc` for  Python 3.9. Does that work?

yfang1: Just tried with that bound. Looks like we  would need:
```
"tensorflow-cpu >= 2.3.0, <=2.5.0rc",
  "grpcio-testing~=1.34.0",
  "grpcio~=1.34.0",
  "six~=1.15.0",
```
Otherwise pip will choose 1.37 grpcio and 1.16 six for some reason.

sasha-gitg: That should be OK, unless those versions cause incompatibility issues.

yfang1: alright, could you put those versions in? Worst case I can fix the dependencies myself later on.

Or if you want me to fold the change in #368 it's fine too.

yfang1 (approved):

setup.py
sasha-gitg: Just pushed the changes. I'll merge this PR when the tests pass.

PR #379, opened 2021-05-10
fix: env formatting

sasha-gitg:


vinnysenthil (approved):
LGTM — glad we caught this.

PR #396, opened 2021-05-17
samples: Add Batch Prediction Job with dedicated resources sample

vinnysenthil:
Manually tested usage with:
```py
import os
import create_batch_prediction_job_dedicated_resources_sample

bpj = create_batch_prediction_job_dedicated_resources_sample.create_batch_prediction_job_dedicated_resources_sample(
    project=os.getenv("BUILD_SPECIFIC_GCLOUD_PROJECT"),
    location='us-central1',
    model_resource_name='3512561418744365056',
    job_display_name='temp_create_batch_prediction_job_test_manual_vinnys',
    gcs_source='gs://ucaip-samples-test-output/inputs/icn_batch_prediction_input.jsonl',
    gcs_destination='gs://ucaip-samples-test-output/',
)
```


samples/model-builder/create_batch_prediction_job_dedicated_resources_sample.py
```diff
@@ -0,0 +1,59 @@
...
+def create_batch_prediction_job_dedicated_resources_sample(
+    project: str,
+    location: str,
+    model_resource_name: str,
+    job_display_name: str,
+    gcs_source: Union[str, Sequence[str]],
+    gcs_destination: str,
+    machine_type: str = "n1-standard-2",
+    accelerator_count: int = 1,
+    accelerator_type: aiplatform_v1.AcceleratorType = aiplatform.gapic.AcceleratorType.NVIDIA_TESLA_K80,
```
sasha-gitg, line 30: We've used string elsewhere:

https://github.com/googleapis/python-aiplatform/blob/master/samples/model-builder/create_training_pipeline_custom_container_job_sample.py#L33

 I think we can do the same here.

sasha-gitg (approved):

samples/model-builder/create_batch_prediction_job_dedicated_resources_sample.py
vinnysenthil: Sg, I'll update to use string.

PR #404, opened 2021-05-18
feat: add tensorboard support to custom job and hyperparameter tuning job

sasha-gitg:

### DIFF
--- a/google/cloud/aiplatform/jobs.py
+++ b/google/cloud/aiplatform/jobs.py
@@ -1168,7 +1168,8 @@ class CustomJob(_RunnableJob):
                 AIP_TENSORBOARD_LOG_DIR
 
                 `service_account` is required with provided `tensorboard`.
-                (TODO: add documentation when released)
+                For more information on configuring your service account please visit:
+                https://cloud.google.com/vertex-ai/docs/experiments/tensorboard-training
             sync (bool):
                 Whether to execute this method synchronously. If False, this method
                 will unblock and it will be executed in a concurrent Future.
@@ -1475,7 +1476,8 @@ class HyperparameterTuningJob(_RunnableJob):
                 AIP_TENSORBOARD_LOG_DIR
 
                 `service_account` is required with provided `tensorboard`.
-                (TODO: add documentation when released)
+                For more information on configuring your service account please visit:
+                https://cloud.google.com/vertex-ai/docs/experiments/tensorboard-training
             sync (bool):
                 Whether to execute this method synchronously. If False, this method
                 will unblock and it will be executed in a concurrent Future.
--- a/google/cloud/aiplatform/jobs.py
+++ b/google/cloud/aiplatform/jobs.py
@@ -45,11 +45,13 @@ from google.cloud.aiplatform.compat.types import (
     batch_prediction_job_v1 as gca_bp_job_v1,
     batch_prediction_job_v1beta1 as gca_bp_job_v1beta1,
     custom_job as gca_custom_job_compat,
+    custom_job_v1beta1 as gca_custom_job_v1beta1,
     explanation_v1beta1 as gca_explanation_v1beta1,
     io as gca_io_compat,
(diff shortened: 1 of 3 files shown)


google/cloud/aiplatform/jobs.py
```diff
@@ -1152,6 +1155,20 @@ def run(
...
+                logs. Format:
+                ``projects/{project}/locations/{location}/tensorboards/{tensorboard}``
+
+                The training script should write Tensorboard to following AI Platform environment
+                variable:
+
+                AIP_TENSORBOARD_LOG_DIR
+
+                `service_account` is required with provided `tensorboard`.
+                (TODO: add documentation when released)
```
ivanmkc, line 1171: Just flagging a TODO

google/cloud/aiplatform/jobs.py
```diff
@@ -1453,9 +1494,22 @@ def run(
                 restart_job_on_worker_restart=restart_job_on_worker_restart,
             )
 
+        if tensorboard:
+            v1beta1_gca_resource = (
```
ivanmkc, line 1498: Is there some hidden coupling between Tensorboard and HyperparameterTuningJob?

Weird that this isn't made explicit in the protos.

google/cloud/aiplatform/jobs.py
sasha-gitg: Added.

google/cloud/aiplatform/jobs.py
```diff
@@ -1170,9 +1188,18 @@ def run(
                 restart_job_on_worker_restart=restart_job_on_worker_restart,
             )
 
+        if tensorboard:
+            v1beta1_gca_resource = gca_custom_job_v1beta1.CustomJob()
+            v1beta1_gca_resource._pb.MergeFromString(
```
ivanmkc, line 1193: I assume that `gca_custom_job_v1beta1` is needed because a certain api_client version is needed and not because Tensorboard is needed?

This code `if tensorboard:` makes it seem like they are explicitly coupled.

google/cloud/aiplatform/jobs.py
```diff
@@ -1453,9 +1496,22 @@ def run(
                 restart_job_on_worker_restart=restart_job_on_worker_restart,
             )
 
+        if tensorboard:
```
ivanmkc, line 1499: Ditto.

google/cloud/aiplatform/jobs.py
sasha-gitg: It's in the protos nested in the CustomJobSpec in HP Tuning Job:
https://github.com/googleapis/python-aiplatform/blob/master/google/cloud/aiplatform_v1beta1/types/hyperparameter_tuning_job.py#L108

https://github.com/googleapis/python-aiplatform/blob/aab9e589426331bfe7ac3f6efa97109e0bd0db0d/google/cloud/aiplatform_v1beta1/types/custom_job.py#L181

vinnysenthil (approved):

google/cloud/aiplatform/jobs.py
ivanmkc: Yes I understand. Just saying that ideally the `if statement` is split in two parts to decouple logic in case other features besides tensorboard also need `v1beta1`.

if is_v1beta1_needed:            
    v1beta1_gca_resource = gca_custom_job_v1beta1.CustomJob()            
    v1beta1_gca_resource._pb.MergeFromString(self._gca_resource._pb.SerializeToString())            
    self._gca_resource = v1beta1_gca_resource            

if tensorboard:
    self._gca_resource.job_spec.tensorboard = tensorboard

ivanmkc: Not blocking, but just something to think about.

sasha-gitg: Ack. I'll leave as is and when we introduce a new v1beta1 feature we'll refactor. Thanks!

PR #464, opened 2021-06-08
fix: response not being printed

Lawiss:
The code was printing the string "response" instead of the variable `response`


Lawiss:
@googlebot I signed it!

samples/snippets/predict_tabular_classification_sample.py
```diff
@@ -44,7 +44,7 @@ def predict_tabular_classification_sample(
     response = client.predict(
         endpoint=endpoint, instances=instances, parameters=parameters
     )
-    print("response")
+    print(response)
```
leahecole, line 47: Non blocking nit - I'm not sure what's all contained in the response but would it make more sense to do something like? cc @googleapis/cdpe-cloudai for product opinion

`print("response: ", response)`

vinnysenthil:
Hi @Lawiss, thanks for raising this. The following structure is intentional:
```py
print("response")
print(" deployed_model_id:", response.deployed_model_id)
# See gs://google-cloud-aiplatform/schema/predict/prediction/tables_classification.yaml for the format of the predictions.
predictions = response.predictions
for prediction in predictions:
    print(" prediction:", dict(prediction))
```

We print out every field in response in a tree-like format with a label instead of the entire object. In this case, the only two fields in this response are `deployed_model_id` and `predictions`. Please let me know if you have any questions or concerns.

Lawiss:
Hello @vinnysenthil, thanks for the explanation 👌

PR #466, opened 2021-06-08
feat: column specs for tabular transformation

sirtorry:
- adds column_specs as an alternative to column_transformation
- adds training_jobs.AutoMLTabularTrainingJob.get_auto_column_specs
- adds aiplatform.column


google/cloud/aiplatform/datasets/tabular_dataset.py
```diff
@@ -39,6 +39,29 @@ class TabularDataset(datasets._Dataset):
         schema.dataset.metadata.tabular,
     )
 
+    def auto_column_specs(self, target_column: str) -> Dict[str, str]:
```
ivanmkc, line 42: Start function with a verb?

sirtorry, line 42: move auto_column_specs to training_jobs class

ivanmkc: We decided offline to move to the training job and just use the target_column property.

google/cloud/aiplatform/datasets/time_series_dataset.py
```diff
@@ -26,6 +26,7 @@
 from google.cloud.aiplatform import utils
 
 
+#TODO: extend tabular dataset
```
sirtorry, line 29: create a bug, resolve in a different PR.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -2706,7 +2743,34 @@ def _add_additional_experiments(self, additional_experiments: List[str]):
         """
         self._additional_experiments.extend(additional_experiments)
 
+    @classmethod
+    def get_auto_column_specs(
+        self, dataset: datasets.TabularDataset, target_column: str,
```
ivanmkc, line 2748: Class isn’t used so this could be a static method.

tests/unit/aiplatform/test_automl_tabular_training_jobs.py
```diff
@@ -227,11 +234,21 @@ def test_run_call_pipeline_service_create(
...
 
+        column_specs = training_jobs.AutoMLTabularTrainingJob.get_auto_column_specs(
+            dataset=mock_dataset_tabular, target_column=_TEST_TRAINING_TARGET_COLUMN,
+        )
+
+        assert column_specs == _TEST_TRAINING_COLUMN_SPECS
+        column_specs[
+            _TEST_TRAINING_COLUMN_NAMES[0]
+        ] = aiplatform.column.data_types.NUMERIC
+        column_specs[_TEST_TRAINING_COLUMN_NAMES[0]] = aiplatform.column.data_types.AUTO
```
ivanmkc, line 245: Why do you set it to numeric and then change it to auto next line?

google/cloud/aiplatform/training_jobs.py
```diff
@@ -2627,11 +2640,37 @@ def _run(
...
 
-        if self._column_transformations is None:
+        # user populated transformations
+        if self._column_transformations is not None and self._column_specs is not None:
+            _LOGGER.info(
+                "column_transformations and column_specs were both passed. column_transformations was used."
+            )
+        if self._column_transformations is not None:
+            column_transformations = self._column_transformations
+        if self._column_specs is not None and column_transformations is None:
```
ivanmkc, line 2658: Could simplify to:
`elif self._column_specs is not None`

google/cloud/aiplatform/column.py
```diff
@@ -0,0 +1,29 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+"""Vertex AI tabular data types"""
+
+
+class data_types:
```
ivanmkc, line 21: I actually do wonder why we don't use an enum instead for this and the schema.

@sasha-gitg did you have a reason why?

ivanmkc (changes requested):
Left comments

google/cloud/aiplatform/training_jobs.py
```diff
@@ -2360,6 +2361,17 @@ def __init__(
                 If an input column has no transformations on it, such a column is
                 ignored by the training, except for the targetColumn, which should have
                 no transformations defined on.
+                Only one of column_transformations or column_specs should be passed.
```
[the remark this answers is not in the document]
sirtorry: @sasha-gitg thoughts

tests/unit/aiplatform/test_automl_tabular_training_jobs.py
sirtorry: see new assert statement

google/cloud/aiplatform/column.py
sasha-gitg: This class is used to generate Pipeline Components. AFAIK Pipeline Components do not support Python enums or Python enums nested in a dictionary.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -2854,12 +2871,37 @@ def _run(
...
+                for data_type in transformation:
+                    column = transformation[data_type]
+                    if column["column_name"] not in column_names:
+                        raise ValueError(f"'{column}' is not in the dataset.")
+                    if column["column_name"] is target_column:
+                        raise ValueError("Target column is in transformations.")
+        elif self._column_specs is not None:
+            column_transformations = [
+                {self._column_specs[column]: {"column_name": column}}
+                for column in self._column_specs
```
sasha-gitg, line 2901: It's more idiomatic to use `self._column_specs.items()`

tests/unit/aiplatform/test_automl_tabular_training_jobs.py
```diff
@@ -309,11 +337,18 @@ def test_run_call_pipeline_if_no_model_display_name(
...
+        )
+
+        assert column_specs == _TEST_TRAINING_COLUMN_SPECS
+
         job = training_jobs.AutoMLTabularTrainingJob(
             display_name=_TEST_DISPLAY_NAME,
             optimization_objective=_TEST_TRAINING_OPTIMIZATION_OBJECTIVE_NAME,
             optimization_prediction_type=_TEST_TRAINING_OPTIMIZATION_PREDICTION_TYPE,
-            column_transformations=_TEST_TRAINING_COLUMN_TRANSFORMATIONS,
+            column_specs=column_specs,
```
sasha-gitg, line 351: We should keep the old tests with column_transformations and create new tests with column specs.

Additionally there should be tests to check the argument validation logic.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -2575,6 +2580,15 @@ def __init__(
                 "minimize-rmse" (default) - Minimize root-mean-squared error (RMSE).
                 "minimize-mae" - Minimize mean-absolute error (MAE).
                 "minimize-rmsle" - Minimize root-mean-squared log error (RMSLE).
+            column_specs (Optional[Dict[str, str]]):
```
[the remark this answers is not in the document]
sasha-gitg: Please also add an example payload of this argument and verbiage on it's format. It's also a bit unintuitive that column_specs and column_transformations have identical docstrings.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -2584,6 +2598,8 @@ def __init__(
                 If an input column has no transformations on it, such a column is
                 ignored by the training, except for the targetColumn, which should have
                 no transformations defined on.
+                Only one of column_transformations or column_specs should be passed.
```
sasha-gitg, line 2601: This line should also be included with `column_specs`.

google/cloud/aiplatform/__init__.py
```diff
@@ -91,4 +92,5 @@
     "TextDataset",
     "TimeSeriesDataset",
     "VideoDataset",
+    "data_types",
```
sasha-gitg, line 95: This name needs to be qualified a bit more. `aiplatform.data_types` seems too generic. Perhaps, `automl_tabular_column_types`.

google/cloud/aiplatform/training_jobs.py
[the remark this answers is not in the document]
sasha-gitg: Please provide a full usage example above in the constructor docstring.

sasha-gitg (changes requested):

google/cloud/aiplatform/training_jobs.py
```diff
@@ -2934,7 +2974,34 @@ def _add_additional_experiments(self, additional_experiments: List[str]):
         """
         self._additional_experiments.extend(additional_experiments)
 
+    def get_auto_column_specs(
```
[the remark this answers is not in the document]
sirtorry: this method is not used during dataset creation, only training job creation. therefore, @ivanmkc believes it should be here.

google/cloud/aiplatform/training_jobs.py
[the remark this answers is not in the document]
sirtorry: we'll support both for now. we can discuss deprecation when there's a major version bump

google/cloud/aiplatform/column.py
sirtorry: todo: track with a bug

google/cloud/aiplatform/training_jobs.py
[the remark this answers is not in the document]
ivanmkc: Yes, we talked about having it on TabularDataset, however, the `column_specs` can only be generated once the `target_column` is known, which is at the training step.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -2854,12 +2871,37 @@ def _run(
...
+            _LOGGER.info(
+                "column_transformations and column_specs were both passed. column_transformations was used."
+            )
+        if self._column_transformations is not None:
+            column_transformations = self._column_transformations
+            column_names = dataset.column_names
+            for transformation in column_transformations:
+                for data_type in transformation:
+                    column = transformation[data_type]
+                    if column["column_name"] not in column_names:
```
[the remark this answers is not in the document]
ivanmkc: I think that's a fair point, since the service might change in the future.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -2575,7 +2585,7 @@ def __init__(
                 "minimize-rmse" (default) - Minimize root-mean-squared error (RMSE).
                 "minimize-mae" - Minimize mean-absolute error (MAE).
                 "minimize-rmsle" - Minimize root-mean-squared log error (RMSLE).
-            column_transformations (Optional[Union[Dict, List[Dict]]]):
+            column_specs (Dict[str, str]):
```
sasha-gitg, line 2588: It seems confusing to have the same docstring for for both `column_specs` and `column_transformations`. The docstring for `column_specs` should be updated to describe the expected format for `column_spec` along with a reference to accepted transformation values.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -2535,6 +2536,15 @@ def __init__(
     ):
         """Constructs a AutoML Tabular Training Job.
 
+        Example usage:
+
+        job = training_jobs.AutoMLTabularTrainingJob(
+            display_name="my_display_name",
+            optimization_prediction_type="classification",
+            optimization_objective="minimize-log-loss",
+            column_specs=my_column_specs,
```
sasha-gitg, line 2545: Extending this argument like so, can give users a quick idea of how to format column_specs:
 ```suggestion
            column_specs={ "column_1": "auto", "column_2": "auto"},
```

File: google/cloud/aiplatform/training_jobs.py [hunk 17]
```diff
@@ -2854,11 +2888,18 @@ def _run(
 
         Returns:
             model: The trained Vertex AI Model resource or None if training did not
-                produce a Vertex AI Model.
+                produce an Vertex AI Model.
```
sasha-gitg, line 2891: It seems like "a" is correct indefinite article unless the guidance from docs is different.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -2934,6 +2973,50 @@ def _add_additional_experiments(self, additional_experiments: List[str]):
...
+            dataset (datasets.TabularDataset):
+                Required. Intended dataset.
+            target_column(str):
+                Required. Intended target column.
+        Returns:
+            Dict[str, str]
+                Column names as keys and 'auto' as values
+
+        Raises:
+            RuntimeError: When no valid source is found.
```
sasha-gitg, line 2999: Where is this being raised?

sasha-gitg (changes requested):

google/cloud/aiplatform/training_jobs.py
```diff
@@ -2584,6 +2594,17 @@ def __init__(
                 If an input column has no transformations on it, such a column is
                 ignored by the training, except for the targetColumn, which should have
                 no transformations defined on.
+                Only one of column_transformations or column_specs should be passed.
+            column_transformations (Union[Dict, List[Dict]]):
+                Optional. Transformations to apply to the input columns (i.e. columns other
+                than the targetColumn). Each transformation may produce multiple
+                result values from the column's value, and all are used for training.
+                When creating transformation for BigQuery Struct column, the column
```
ivanmkc, line 2602: You also might want to note that only columns with no child columns of their own should have a transformation applied.

For example, if you have a column "field_a.field_b", then pass in the transformation for "field_a.field_b" but do not pass in the transformation for "field_a".

google/cloud/aiplatform/training_jobs.py
ivanmkc: Really? Sounds strange to me.

tests/unit/aiplatform/test_automl_tabular_training_jobs.py
```diff
@@ -515,12 +570,250 @@ def test_run_call_pipeline_service_create_if_set_additional_experiments(
...
+            target_column=_TEST_TRAINING_TARGET_COLUMN,
+        )
+
+        assert column_specs == _TEST_TRAINING_COLUMN_SPECS
+
+        job = training_jobs.AutoMLTabularTrainingJob(
+            display_name=_TEST_DISPLAY_NAME,
+            optimization_objective=_TEST_TRAINING_OPTIMIZATION_OBJECTIVE_NAME,
+            optimization_prediction_type=_TEST_TRAINING_OPTIMIZATION_PREDICTION_TYPE,
+            column_specs=column_specs,
```
ivanmkc, line 595: Consider adding tests for error cases, such as when both `column_specs` and `column_transformations` are passed. And if `target_column` is not included.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -2934,6 +2973,50 @@ def _add_additional_experiments(self, additional_experiments: List[str]):
...
+                Required. Intended target column.
+        Returns:
+            Dict[str, str]
+                Column names as keys and 'auto' as values
+
+        Raises:
+            RuntimeError: When no valid source is found.
+            ValueError: When target_column is not in dataset
+        """
+        if target_column not in dataset.column_names:
```
ivanmkc, line 3002: I believe the general thinking is that we defer to the backend to throw these errors instead of checking on the client. You can remove this check.

tests/unit/aiplatform/test_automl_tabular_training_jobs.py
```diff
@@ -515,12 +570,250 @@ def test_run_call_pipeline_service_create_if_set_additional_experiments(
...
+        ] = training_jobs.AutoMLTabularTrainingJob.column_data_types.CATEGORICAL
+        column_specs[
+            _TEST_TRAINING_COLUMN_NAMES_ALTERNATIVE[2]
+        ] = training_jobs.AutoMLTabularTrainingJob.column_data_types.TEXT
+
+        job = training_jobs.AutoMLTabularTrainingJob(
+            display_name=_TEST_DISPLAY_NAME,
+            optimization_objective=_TEST_TRAINING_OPTIMIZATION_OBJECTIVE_NAME,
+            optimization_prediction_type=_TEST_TRAINING_OPTIMIZATION_PREDICTION_TYPE,
+            column_specs=column_specs,
```
ivanmkc, line 759: Consider adding a test to provide coverage for the error cases, such as when both `column_specs` and `column_transformations` are provided.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -2636,7 +2657,20 @@ def __init__(
             training_encryption_spec_key_name=training_encryption_spec_key_name,
             model_encryption_spec_key_name=model_encryption_spec_key_name,
         )
-        self._column_transformations = column_transformations
+        # user populated transformations
+        if column_transformations is not None and column_specs is not None:
+            _LOGGER.info(
+                "column_transformations and column_specs were both passed. column_transformations was used."
+            )
```
ivanmkc, line 2664: Why not raise Error instead?

ivanmkc: I think that's more common, see what happens when gcs_source and bq_source are simultaneously passed to a Dataset.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -2859,6 +2898,13 @@ def _run(
 
         training_task_definition = schema.training_job.definition.automl_tabular
 
+        # convert column specs to column transformations
+        if self._column_specs is not None:
```
sasha-gitg, line 2902: To simplify, I think we can avoid the _column_specs instance attribute and capture this in the constructor above like so. Also it's more readable to name the tuple values in the list comprehension:

 ```Python
if column_transformations is not None:
    self._column_transformations = column_transformations
elif column_specs is not None:
    self._column_transformations = [
        {transformation: {"column_name": column_name}}
        for column_name, transformation in column_specs.items()
    ]
else:
    self._column_transformations = None
```

sasha-gitg (approved):
Mainly LGTM. One more comment it seems I didn't include in the last review. Thanks!

ivanmkc (approved):
LGTM after Sasha's suggestion

ivanmkc:
What about the deprecation warning for column_transformations?

PR #469, opened 2021-06-08
fix: pass credentials to BQ and GCS clients

sasha-gitg:


google/cloud/aiplatform/datasets/tabular_dataset.py
```diff
@@ -156,7 +165,11 @@ def _retrieve_gcs_source_columns(project: str, gcs_csv_file_path: str) -> List[s
         return next(csv_reader)
 
     @staticmethod
-    def _retrieve_bq_source_columns(project: str, bq_table_uri: str) -> List[str]:
+    def _retrieve_bq_source_columns(
+        project: str,
+        bq_table_uri:str,
```
ivanmkc, line 170: bq_table_uri: str

google/cloud/aiplatform/datasets/tabular_dataset.py
```diff
@@ -73,20 +73,28 @@ def column_names(self) -> List[str]:
                 gcs_source_uris.sort()
 
                 # Get the first file in sorted list
-                return TabularDataset._retrieve_gcs_source_columns(
-                    self.project, gcs_source_uris[0]
+                return self._retrieve_gcs_source_columns(
```
ivanmkc, line 76: Is there any use case where the user would want to pass in different credentials when calling `dataset.column_names`?

If not, this looks fine.

ivanmkc (approved):
LGTM

Once tests pass.

PR #488, opened 2021-06-18
feat: add cancel method to pipeline client

ji-yaqi:


google/cloud/aiplatform/pipeline_jobs.py
```diff
@@ -264,7 +264,7 @@ def state(self) -> Optional[gca_pipeline_state_v1beta1.PipelineState]:
     @property
     def _has_run(self) -> bool:
         """Helper property to check if this pipeline job has been run."""
-        return bool(self._gca_resource.name)
+        return bool(self._gca_resource.create_time)
```
sasha-gitg, line 267: What is the motivation for relying on `create_time`? I wonder if we should use it elsewhere.

sasha-gitg (approved):
LGTM! Thanks!

google/cloud/aiplatform/pipeline_jobs.py
ji-yaqi: For Pipeline Job, name is created by us from the SDK side, so it will exist even if we don't start the run. (Compared to other jobs, their names are created from the BE)

ji-yaqi:
The Kokoro doc presubmit error is tracked in: https://github.com/googleapis/sphinx-docfx-yaml/issues/47

PR #494, opened 2021-06-21
fix: check if training_task_metadata is populated before logging backingCustomJob

sasha-gitg:


vinnysenthil (approved):
Manually tested, LGTM!

diegodulval:
I don't know what's happening, I still get the following error, even putting version 1.1.1

self._gca_resource.training_task_metadata.get("backingCustomJob")
"AttributeError: 'NoneType' object has no attribute 'get'

I'm using this project here: https://github.com/aliz-aip/stefadam-vertex-test2/tree/master/vertex_ai

PR #507, opened 2021-06-24
feat: add basic explain get_metadata function for tf2. 

ji-yaqi:
Basic explain get_metadata function, for tf2. 

Usage:

```
builder = google.cloud.aiplatform.explain.metadata.tf.v2.saved_model_metadata_builder.SavedModelMetadataBuilder(
    "MODEL_PATH"
)
metadata = builder.get_metadata()
```

The main difference from Vertex SDK and XAI SDK's `get_metadata` function:
1. In Vertex SDK, users the camelCase while in XAI, variables users snake_case
e.g. Vertex: `inputTensorName`, XAI: `input_tensor_name`

2. If modality and encoding in inputs are not set, XAI SDK populates the default: 
```
      'modality': 'categorical',
      'encoding': 'identity'
```
Vertex SDK doesn't populate the default values. It only populates the users's settings.
The two methods will not make any difference in the actual settings.


setup.py
```diff
@@ -31,7 +31,10 @@
 
 tensorboard_extra_require = ["tensorflow >=2.3.0, <=2.5.0"]
 metadata_extra_require = ["pandas >= 1.0.0"]
-full_extra_require = tensorboard_extra_require + metadata_extra_require
+xai_extra_require = ["tensorflow-cpu>=2.3.0, <=2.5.0"]
```
sasha-gitg, line 34: "tensorflow" is the CPU-only release: https://www.tensorflow.org/install/pip

This will also make it easier to manage with TB's TF dependency.

setup.py
```diff
@@ -31,7 +31,10 @@
...
+full_extra_require = (
```
sasha-gitg, line 35: tensorboard and xai have the same dependencies so we may need to deduplicate this with `list(set(...))`.

tests/unit/aiplatform/test_explain_saved_model_metadata_builder_test.py
```diff
@@ -0,0 +1,167 @@
...
+#
+
+import tensorflow as tf
+import numpy as np
+
+from google.cloud.aiplatform.explain.metadata.tf.v2 import saved_model_metadata_builder
+
+
+class SavedModelMetadataBuilderTest(tf.test.TestCase):
+    def _set_up(self):
```
sasha-gitg, line 25: Since this is only used by one test, it should perhaps not be a shared method.

sasha-gitg (changes requested):

sasha-gitg (approved):

PR #512, opened 2021-06-25
fix: Set prediction client when listing Endpoints

vinnysenthil:
Overrides `VertexAiResourceNounWithFutureManager._construct_sdk_resource_from_gapic` with `Endpoint._construct_sdk_resource_from_gapic`.

Subclass method initializes the Prediction service client, allowing users to call `predict()` / `explain()` on an `Endpoint` returned by a call to `list()`

Fixes [b/192002630](http://b/192002630) 🦕

### DIFF
--- a/google/cloud/aiplatform/models.py
+++ b/google/cloud/aiplatform/models.py
@@ -117,13 +117,8 @@ class Endpoint(base.VertexAiResourceNounWithFutureManager):
         )
         self._gca_resource = self._get_gca_resource(resource_name=endpoint_name)
 
-        project, location = self._get_and_validate_project_location(
-            resource_name=self._gca_resource.name, project=project, location=location
-        )
-
         self._prediction_client = self._instantiate_prediction_client(
-            location=location or initializer.global_config.location,
-            credentials=credentials,
+            location=self.location, credentials=credentials,
         )
 
     @property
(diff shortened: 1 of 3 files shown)


google/cloud/aiplatform/models.py
```diff
@@ -324,6 +324,45 @@ def _create(
...
+            Endpoint:
+                An initialized Endpoint resource.
+        """
+        endpoint = self._empty_constructor(
+            project=project, location=location, credentials=credentials
+        )
+
+        endpoint._gca_resource = gapic_resource
+        endpoint._prediction_client = self._instantiate_prediction_client(
+            location=location or initializer.global_config.location,
```
sasha-gitg, line 360: Just realized there is a bug here and in the implementation it's based on:

https://github.com/googleapis/python-aiplatform/blob/master/google/cloud/aiplatform/models.py#L119-L122

If passed a full resource name that doesn't match `location` or `init.location` then will instantiate the client in the incorrect location.

Please change this here and in the reference above to the use the constructed endpoint resource's location.

google/cloud/aiplatform/models.py
```diff
@@ -324,6 +324,45 @@ def _create(
             credentials=credentials,
         )
 
+    def _construct_sdk_resource_from_gapic(
+        self,
```
sasha-gitg, line 328: It seem like this should be converted to a classmethod along with the implementation in the base class.

sasha-gitg (changes requested):
This PR should also include a test to confirm the fix. Thanks!

google/cloud/aiplatform/models.py
vinnysenthil: Added calls to `_get_and_validate_project_location()` in both `Endpoint.__init__()` and `Endpoint._construct_sdk_resource_from_gapic()` to allow the full resource name to override location.

vinnysenthil: Actually I realized that downstream calls in `VertexAiResourceNounWithFutureManager._empty_constructor()` and `VertexAiResourceNounWithFutureManager.__init__()` already call `_get_and_validate_project_location()`.

I've updated the call to `_instantiate_prediction_client()` to use the location set by those methods instead of the `location` argument.

sasha-gitg (approved):
This is great. Thanks, Vinny!

PR #549, opened 2021-07-16
feat: add filter and timestamp splits

sirtorry:
Changes to training splits:

- Modifies default behaviour of fraction split
- Raises error when multiple split types passed
- Adds filter splits (b/172365904) to image, text and video
- Adds timestamp split (b/172368070) to tabular

Future:

- next PR will change forecasting training jobs to extend tabular; therefore, adding timestamp split to forecasting

succeeds #210

### DIFF
--- a/google/cloud/aiplatform/training_jobs.py
+++ b/google/cloud/aiplatform/training_jobs.py
@@ -64,6 +64,18 @@ _PIPELINE_COMPLETE_STATES = set(
     ]
 )
 
+_DATASETS_WITH_DATA_ITEMS = set(
+    [
+        schema.dataset.metadata.image,
+        schema.dataset.metadata.text,
+        schema.dataset.metadata.video,
+    ]
+)
+
+_DATASETS_WITH_STRUCTURED_DATA = set(
+    [schema.dataset.metadata.tabular, schema.dataset.metadata.time_series]
+)
+
 
 class _TrainingJob(base.VertexAiResourceNounWithFutureManager):
 
@@ -294,9 +306,9 @@ class _TrainingJob(base.VertexAiResourceNounWithFutureManager):
     def _create_input_data_config(
         dataset: Optional[datasets._Dataset] = None,
         annotation_schema_uri: Optional[str] = None,
-        training_fraction_split: Optional[float] = 0.8,
-        validation_fraction_split: Optional[float] = 0.1,
-        test_fraction_split: Optional[float] = 0.1,
+        training_fraction_split: Optional[float] = None,
+        validation_fraction_split: Optional[float] = None,
+        test_fraction_split: Optional[float] = None,
         training_filter_split: Optional[str] = None,
         validation_filter_split: Optional[str] = None,
         test_filter_split: Optional[str] = None,
@@ -413,59 +425,46 @@ class _TrainingJob(base.VertexAiResourceNounWithFutureManager):
                 -  AIP_TRAINING_DATA_URI ="bigquery_destination.dataset_*.training"
                 -  AIP_VALIDATION_DATA_URI = "bigquery_destination.dataset_*.validation"
                 -  AIP_TEST_DATA_URI = "bigquery_destination.dataset_*.test"
+        Raises:
+            ValueError: When more than 1 type of split configuration is passed or when
+                the split configuartion passed is incompatible with the dataset schema.
         """
 
         input_data_config = None
         if dataset:
+            # all possible splits
             filter_split = None
-            if any(
-                [
-                    training_filter_split is not None,
-                    validation_filter_split is not None,
-                    test_filter_split is not None,
-                ]
-            ):
-                if all(
-                    [
-                        training_filter_split is not None,
-                        validation_filter_split is not None,
-                        test_filter_split is not None,
-                        dataset._gca_resource.metadata_schema_uri
-                        in (
-                            schema.dataset.metadata.image,
-                            schema.dataset.metadata.text,
-                            schema.dataset.metadata.video,
-                        ),
-                    ]
-                ):
-                    # Create filter split spec
-                    filter_split = gca_training_pipeline.FilterSplit(
-                        training_filter=training_filter_split,
-                        validation_filter=validation_filter_split,
-                        test_filter=test_filter_split,
+            predefined_split = None
+            timestamp_split = None
+            fraction_split = None
+
+            # filter split config
+            if any([training_filter_split, validation_filter_split, test_filter_split]):
+                if dataset.metadata_schema_uri not in _DATASETS_WITH_DATA_ITEMS:
+                    raise ValueError(
+                        "A filter split may only be used with datesets with DataItems (image, text, video)"
                     )
-                elif any(
+                if any(
                     [
                         training_filter_split is None,
                         validation_filter_split is None,
                         test_filter_split is None,
                     ]
                 ):
-                    raise TypeError(
-                        "To use filter split, all of the following have to be provided: training_filter_split, validation_filter_split, test_filter_split"
-                    )
-                else:
                     raise ValueError(
-                        "A filter split may only be used with datesets with DataItems"
+                        """To use filter split, all of the following have to be provided:
+                        training_filter_split, validation_filter_split, test_filter_split"""
                     )
 
-            # Create predefined split spec
-            predefined_split = None
+                filter_split = gca_training_pipeline.FilterSplit(
+                    training_filter=training_filter_split,
+                    validation_filter=validation_filter_split,
+                    test_filter=test_filter_split,
+                )
+
+            # predefined split config
             if predefined_split_column_name:
-                if dataset._gca_resource.metadata_schema_uri not in (
-                    schema.dataset.metadata.tabular,
-                    schema.dataset.metadata.time_series,
-                ):
+                if dataset.metadata_schema_uri not in _DATASETS_WITH_STRUCTURED_DATA:
                     raise ValueError(
                         "A pre-defined split may only be used with a tabular or time series Dataset"
                     )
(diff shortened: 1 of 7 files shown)


google/cloud/aiplatform/training_jobs.py
```diff
@@ -386,12 +417,47 @@ def _create_input_data_config(
...
+                        test_filter_split is None,
+                    ]
+                ):
+                    raise TypeError(
+                        "To use filter split, all of the following have to be provided: training_filter_split, validation_filter_split, test_filter_split"
+                    )
+                else:
+                    raise ValueError(
+                        "A filter split may only be used with datesets with DataItems"
+                    )
```
sirtorry, line 460: I plan on writing a couple more tests to check the behaviour when multiple types of splits are passed. But before I do that, I'd like y'all's thoughts on whether this is what we'd want.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -408,6 +474,37 @@ def _create_input_data_config(
...
+                    predefined_split is None,
+                    timestamp_split is None,
+                ]
+            ):
+                fraction_split = gca_training_pipeline.FractionSplit(
+                    training_fraction=training_fraction_split,
+                    validation_fraction=validation_fraction_split,
+                    test_fraction=test_fraction_split,
+                )
+
```
sirtorry, line 507: I plan on writing a couple more tests to check the behaviour when multiple types of splits are passed. But before I do that, I'd like y'all's thoughts on whether this is what we'd want.

ivanmkc:
From the docstring, it's unclear to me what happens when None is passed for `training_fraction_split`. 
Followup question, can we default to None?

Also, can we remove the defaults from private methods and only include on public methods? Seems cleaner.

sirtorry:
> From the docstring, it's unclear to me what happens when None is passed for `training_fraction_split`.
> Followup question, can we default to None?
> 
> Also, can we remove the defaults from private methods and only include on public methods? Seems cleaner.

I have no objections to either. @sasha-gitg what are your thoughts?

google/cloud/aiplatform/training_jobs.py
```diff
@@ -386,12 +417,47 @@ def _create_input_data_config(
...
+                ]
+            ):
+                if all(
+                    [
+                        training_filter_split is not None,
+                        validation_filter_split is not None,
+                        test_filter_split is not None,
+                        dataset._gca_resource.metadata_schema_uri
+                        in (
+                            schema.dataset.metadata.image,
```
sasha-gitg, line 435: This should be a constant at the top of this file. Something like `_DATASETS_WITH_DATA_ITEMS`.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -386,12 +417,47 @@ def _create_input_data_config(
...
+                        test_filter=test_filter_split,
+                    )
+                elif any(
+                    [
+                        training_filter_split is None,
+                        validation_filter_split is None,
+                        test_filter_split is None,
+                    ]
+                ):
+                    raise TypeError(
```
sasha-gitg, line 454: This should be  ValueError.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -386,12 +417,47 @@ def _create_input_data_config(
...
+                        "To use filter split, all of the following have to be provided: training_filter_split, validation_filter_split, test_filter_split"
+                    )
+                else:
+                    raise ValueError(
+                        "A filter split may only be used with datesets with DataItems"
```
sasha-gitg, line 459: This error statement should include the dataset types that support DataItems.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -408,6 +474,38 @@ def _create_input_data_config(
                     key=predefined_split_column_name
                 )
 
+            # Create timestamp split spec
+            timestamp_split = None
+            if timestamp_split_column_name:
+                if dataset._gca_resource.metadata_schema_uri not in (
+                    schema.dataset.metadata.tabular,
```
sasha-gitg, line 481: Should also be a constant.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -408,6 +474,38 @@ def _create_input_data_config(
...
+
+                timestamp_split = gca_training_pipeline.TimestampSplit(
+                    training_fraction=training_fraction_split,
+                    validation_fraction=validation_fraction_split,
+                    test_fraction=test_fraction_split,
+                    key=timestamp_split_column_name,
+                )
+
+            fraction_split = None
+            if all(
```
sasha-gitg, line 496: I think here we should be checking if we received any of the fraction split arguments, and not the timestamp split, and, if so, create the FractionSplit proto and default any missing values to there defaults in previous versions.

Then below this there should be a block like this:

```python
n_split_configs = len([split for split in [fraction_split, filter_split, timestamp_split, predefined_split] if split is not None])
if  n_split_configs > 1:
  raise ValueError("Can only specify one of:"
     "1. training_fraction_split, validation_fraction_split, testing_fraction_split"
     "2. training_fraction_split, validation_fraction_split, testing_fraction_split, timestamp_split,"
     "3. ....")
     
if n_split_configs = 0:
  # create default 0.8, 0.1, 0.1
```

google/cloud/aiplatform/training_jobs.py
```diff
@@ -408,6 +474,38 @@ def _create_input_data_config(
...
+                    )
+
+                timestamp_split = gca_training_pipeline.TimestampSplit(
+                    training_fraction=training_fraction_split,
+                    validation_fraction=validation_fraction_split,
+                    test_fraction=test_fraction_split,
+                    key=timestamp_split_column_name,
+                )
+
+            fraction_split = None
```
sasha-gitg, line 495: I agree with the comment that *_fraction_splits need to default to None throughout because we will not be able to differentiate when the user explicitly set those values and when they have been defaulted which is important as we need to determine if we received an invalid set of arguments.

If we agree on that then we need to also update the documentation that if no fraction_spit, timestamp_split, filter_split, or predfined_split is set then we default to a 0.8, 0.1, 0.1 fraction_split.

There needs to be additional logic at the end here when if all splits are None then fallback to a 0.8, 0.1, 0.1 fraction split.

I don't think this is a breaking change as we moved to setting the argument as Optional.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -425,7 +523,9 @@ def _create_input_data_config(
             # create input data config
             input_data_config = gca_training_pipeline.InputDataConfig(
                 fraction_split=fraction_split,
+                filter_split=filter_split,
```
sasha-gitg, line 526: I think there is a bug here if we end up creating multiple split configs because of the passed in arguments then only one of these splits gets set depending on however the precedence is handled by the underlying GAPIC lib.

If we use the logic described above then we should be OK because we'll only be passing one config and the rest as None.

sasha-gitg (changes requested):

google/cloud/aiplatform/training_jobs.py
sirtorry: everything is now none and falls back to 0.8, 0.1, 0.1. with the exception of video, where 0.8, none, 0.2 has been used.

sirtorry: i'm also hesitant to have these schema specific values in the training jobs parent class.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -1653,14 +1829,35 @@ def run(
             accelerator_count (int):
                 The number of accelerators to attach to a worker replica.
             training_fraction_split (float):
-                The fraction of the input data that is to be
-                used to train the Model. This is ignored if Dataset is not provided.
+                Optional. The fraction of the input data that is to be used to train
```
sasha-gitg, line 1832: There should be additional documentation at the top of all public `run` method docstrings describing the different splits. Right now, It' just  the fraction split that has that treatment.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -1653,14 +1829,35 @@ def run(
...
-                used to validate the Model. This is ignored if Dataset is not provided.
+                Optional. The fraction of the input data that is to be used to validate
+                the Model. This is ignored if Dataset is not provided.
             test_fraction_split (float):
-                The fraction of the input data that is to be
-                used to evaluate the Model. This is ignored if Dataset is not provided.
+                Optional. The fraction of the input data that is to be used to evaluate
+                the Model. This is ignored if Dataset is not provided.
+            training_filter_split (str):
+                Optional. A filter on DataItems of the Dataset. DataItems that match
```
sasha-gitg, line 1841: It would be worth going back through the protos and ensuring any additional information that is not argument specific but split specific is surfaced like the minus sign usage here. This can only be done on public run methods.: https://github.com/googleapis/python-aiplatform/blob/master/google/cloud/aiplatform_v1/types/training_pipeline.py#L348

sasha-gitg (changes requested):

tests/unit/aiplatform/test_automl_video_training_jobs.py
```diff
@@ -299,17 +306,18 @@ def test_run_call_pipeline_service_create(
         model_from_job = job.run(
             dataset=mock_dataset_video,
             model_display_name=_TEST_MODEL_DISPLAY_NAME,
-            training_fraction_split=_TEST_FRACTION_SPLIT_TRAINING,
-            test_fraction_split=_TEST_FRACTION_SPLIT_TEST,
+            training_filter_split=_TEST_FILTER_SPLIT_TRAINING,
```
ivanmkc, line 309: Not sure if this is intentional missing a `validation_filter_split` or not.

tests/unit/aiplatform/test_automl_video_training_jobs.py
```diff
@@ -362,16 +370,17 @@ def test_run_call_pipeline_if_no_model_display_name(
 
         model_from_job = job.run(
             dataset=mock_dataset_video,
-            training_fraction_split=_TEST_FRACTION_SPLIT_TRAINING,
-            test_fraction_split=_TEST_FRACTION_SPLIT_TEST,
+            training_fraction_split=_TEST_ALTERNATE_FRACTION_SPLIT_TRAINING,
+            test_fraction_split=_TEST_ALTERNATE_FRACTION_SPLIT_TEST,
```
ivanmkc, line 374: Not sure if this is intentional missing a validation_filter_split or not.

google/cloud/aiplatform/training_jobs.py
ivanmkc: Good catch. You can put it under where the `Data fraction splits` info is in the `run` method docstring.

sirtorry:
I'll make the final changes sometime over the next few days.

sasha-gitg:
Closing in favor of #627

PR #567, opened 2021-07-22
fix: create pipeline job with user-specified job id

ji-yaqi:


google/cloud/aiplatform/pipeline_jobs.py
```diff
@@ -249,8 +249,11 @@ def run(
 
         _LOGGER.log_create_with_lro(self.__class__)
 
+        pipeline_job_id = self._gca_resource.name.split("/")[-1]
```
sasha-gitg, line 252: AFAIK `PipelneJob.name` is not used by the service in the request as it's output only. Possibly add a comment that `PipelineJob.name` is not used by the service.

sasha-gitg (approved):
LGTM! Thanks!

google/cloud/aiplatform/pipeline_jobs.py
ji-yaqi: Added, thanks!

PR #576, opened 2021-07-27
feat: Accommodate multiple parameters in children of VertexModel

anooshrees:

### DIFF
--- a/google/cloud/aiplatform/experimental/vertex_model/base.py
+++ b/google/cloud/aiplatform/experimental/vertex_model/base.py
@@ -167,10 +167,11 @@ class VertexModel(metaclass=abc.ABCMeta):
 
     """ Parent class that users can extend to use the Vertex AI SDK """
 
-    def __init__(self):
+    def __init__(self, *args, **kwargs):
         # Default to local training on creation, at least for this prototype.
         self.training_mode = "local"
         self.fit = vertex_fit_function_wrapper(self.fit)
+        self._constructor_arguments = (args, kwargs)
 
     @abc.abstractmethod
     def fit(self):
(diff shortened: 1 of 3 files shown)


anooshrees:
@sasha-gitg this PR is ready for review. Thanks!

google/cloud/aiplatform/experimental/vertex_model/base.py
```diff
@@ -168,10 +168,11 @@ class VertexModel(metaclass=abc.ABCMeta):
 
     """ Parent class that users can extend to use the Vertex AI SDK """
 
-    def __init__(self):
+    def __init__(self, *args, **kwargs):
         # Default to local training on creation, at least for this prototype.
```
sasha-gitg, line 172: This should include a docstring stating the expectation that subclasses should pass their constructor arguments into this base class.

google/cloud/aiplatform/experimental/vertex_model/utils/source_utils.py
```diff
@@ -87,28 +87,36 @@ def _make_source(
...
+    # First, grab args and kwargs using the _constructor_arguments variable in VertexModel
+    class_args = inspect.signature(obj.__class__.__init__).bind(
+        obj, *obj._constructor_arguments[0], **obj._constructor_arguments[1]
+    )
+
+    # need to index pass the first arg to avoid the call to self.
+    src = src + f"\tmodel = {cls_name}({class_args.args[1:]}, {class_args.kwargs})\n"
 
-    src = src + f"\tmodel.{instance_method}("
+    if instance_method is not None:
```
sasha-gitg, line 98: When would instance_method be `None`? It seems like it's required based on the signature of `_make_source`.

sasha-gitg (approved):
LGTM! Added a few comments. Thanks!

google/cloud/aiplatform/experimental/vertex_model/utils/source_utils.py
anooshrees: I used `None` during testing to test that the multiple constructor arguments passed.

PR #577, opened 2021-07-28
feat: Serialization and deserialization of PyTorch DataLoader objects

anooshrees:
- Added serialization/deserialization for DataLoaders with both remote and local sources
- Added serialization/deserialization for model objects
- Added tests for both forms of serialization

### DIFF
--- a/google/cloud/aiplatform/experimental/vertex_model/serializers/dataloaders.py
+++ b/google/cloud/aiplatform/experimental/vertex_model/serializers/dataloaders.py
@@ -76,38 +76,6 @@ def _serialize_remote_dataloader(
         destination_blob_prefix,
     ) = utils.extract_bucket_and_prefix_from_gcs_path(artifact_uri)
 
-    # Retrieve the blob and bucket name of the original GCS object
-    data_gcs_path = None
-
-    if dataloader_path is not None:
-        dataloader_path = pathlib.Path(dataloader_path)
-        local_file_name = dataset_type + "_" + dataloader_path.name
-        source_blob_name = local_file_name
-
-        if source_blob_prefix:
-            source_blob_name = "/".join([source_blob_prefix, source_blob_name])
-
-        source_bucket = client.bucket(source_bucket_name)
-        source_blob = source_bucket.blob(source_blob_name)
-
-        # Create a bucket and blob using the artifact URI
-        destination_bucket = client.bucket(destination_bucket_name)
-
-        destination_blob_name = source_blob_name
-        if destination_blob_prefix:
-            destination_blob_name = "/".join(
-                [destination_blob_prefix, destination_blob_name]
-            )
-
-        # Copy over the object from the source bucket to the new bucket
-        blob_copy = source_bucket.copy_blob(
-            source_blob, destination_bucket, destination_blob_name
-        )
-
-        data_gcs_path = "".join(
-            ["gs://", "/".join([destination_bucket_name, blob_copy.name])]
-        )
-
     path = serializer_utils.serialize_to_tmp_and_copy_to_gcs(
         "my_" + dataset_type + "_dataloader.pth",
         destination_bucket,
@@ -115,7 +83,7 @@ def _serialize_remote_dataloader(
         functools.partial(torch.save, obj),
     )
 
-    return path, data_gcs_path
+    return path, dataloader_path
 
 
 def _serialize_local_dataloader(
--- a/google/cloud/aiplatform/experimental/vertex_model/serializers/dataloaders.py
+++ b/google/cloud/aiplatform/experimental/vertex_model/serializers/dataloaders.py
@@ -15,10 +15,62 @@
 # limitations under the License.
 #
 
+import pathlib
+import torch
 
-def _serialize_remote_dataloader():
-    # writes the referenced data to the run-time bucket
-    raise NotImplementedError
+from google.cloud import storage
+from google.cloud.aiplatform import initializer
+from google.cloud.aiplatform import utils
+
+
+def _serialize_remote_dataloader(
+    artifact_uri: str, dataloader_path: str, dataset_type: str
+) -> str:
+    """writes the referenced data to the run-time bucket"""
+    # Create a client object
+    client = storage.Client(
+        project=initializer.global_config.project,
+        credentials=initializer.global_config.credentials,
+    )
+
+    # Retrieve the source and blob names from the dataloader path
+    (
+        source_bucket_name,
+        source_blob_prefix,
+    ) = utils.extract_bucket_and_prefix_from_gcs_path(dataloader_path)
+
+    # Retrieve the source and blob names from the artifact URI
+    (
+        destination_bucket_name,
+        destination_blob_prefix,
+    ) = utils.extract_bucket_and_prefix_from_gcs_path(artifact_uri)
+
+    # Retrieve the blob and bucket name of the original GCS object
+    dataloader_path = pathlib.Path(dataloader_path)
+    local_file_name = dataloader_path.name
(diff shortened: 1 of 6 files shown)


anooshrees:
Cover failed -- not entirely sure what that check does?

sasha-gitg (changes requested):

google/cloud/aiplatform/experimental/vertex_model/serializers/dataloaders.py
```diff
@@ -15,35 +15,253 @@
 # limitations under the License.
```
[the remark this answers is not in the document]
anooshrees: will do this in the next PR.

google/cloud/aiplatform/experimental/vertex_model/serializers/dataloaders.py
```diff
@@ -15,35 +15,256 @@
...
+        destination_blob_prefix,
+        torch.save,
+    )
+
+    return path, data_gcs_path
+
+
+def _serialize_local_dataloader(
+    artifact_uri: str,
+    dataloader_path: str,
```
sasha-gitg, line 118: This should be `Optional` at `root` can be `None` based on the `_serialize_dataloader`. Also it looks like the `None` case isn't handled as `dataloader_path` is always passed to `pathlib.Path`

google/cloud/aiplatform/experimental/vertex_model/serializers/dataloaders.py
```diff
@@ -15,35 +15,256 @@
...
+        dataset_type (str): the intended use of the dataset (ie. training, testing)
+
+    Returns:
+        The GCS path pointing to the serialized DataLoader, the GCS path pointing to the
+        serialized origin data.
+    """
+    my_dataset = obj.dataset
+    root = getattr(my_dataset, "root", None)
+
+    if root is None:
```
sasha-gitg, line 195: This block can be simplified to:
```Python
    # Decide whether to pass to remote or local serialization
    if root and root.startswith("gs://"):
        return _serialize_remote_dataloader(artifact_uri, root, obj, dataset_type)
    else:
        return _serialize_local_dataloader(artifact_uri, root, obj, dataset_type)
```

google/cloud/aiplatform/experimental/vertex_model/serializers/dataloaders.py
```diff
@@ -15,35 +15,256 @@
...
+    # Return a pandas DataFrame read from the csv in the cloud
+    return dataloader
+
+
+def serialize_to_tmp_and_copy_to_gcs(
+    file_name: str,
+    destination_bucket: storage.Bucket,
+    blob_prefix: str,
+    serialize_fn: Callable[[str], None],
+):
```
sasha-gitg, line 254: docstring

google/cloud/aiplatform/experimental/vertex_model/serializers/dataloaders.py
```diff
@@ -15,35 +15,256 @@
...
+    destination_bucket = client.bucket(destination_bucket_name)
+
+    destination_blob_name = source_blob_name
+    if destination_blob_prefix:
+        destination_blob_name = "/".join(
+            [destination_blob_prefix, destination_blob_name]
+        )
+
+    # Copy over the object from the source bucket to the new bucket
+    blob_copy = source_bucket.copy_blob(
```
sasha-gitg, line 98: This doesn't seem to be necessary. I think we can assume we can retrieve the data from the original remote source.

google/cloud/aiplatform/experimental/vertex_model/serializers/dataloaders.py
```diff
@@ -15,35 +15,256 @@
...
+
+    data_gcs_path = "".join(
+        ["gs://", "/".join([destination_bucket_name, blob_copy.name])]
+    )
+
+    path = serialize_to_tmp_and_copy_to_gcs(
+        "my_" + dataset_type + "_dataloader.pth",
+        destination_bucket,
+        destination_blob_prefix,
+        torch.save,
```
sasha-gitg, line 110: This needs to encapsulate the object as well: 

`functools.partial(torch.save, obj)`
or
`lambda f: torch.save(obj, f)`

google/cloud/aiplatform/experimental/vertex_model/serializers/dataloaders.py
```diff
@@ -15,35 +15,256 @@
...
+        return _serialize_local_dataloader(artifact_uri, root, obj, dataset_type)
+
+    # Decide whether to pass to remote or local serialization
+    if root.startswith("gs://"):
+        return _serialize_remote_dataloader(artifact_uri, root, obj, dataset_type)
+    else:
+        return _serialize_local_dataloader(artifact_uri, root, obj, dataset_type)
+
+
+def _deserialize_dataloader(artifact_uri: str) -> DataLoader:
```
sasha-gitg, line 205: How is the case where the local dataset is also serialized to GCS handled (`data_gcs_path` in the serializer)?

It seems like it needs to be copied from GCS and possibly have the local reference in the object replaced.

Right now it seems like we are only deserializing `path`, the torch object, from the serializer.

sasha-gitg (changes requested):

google/cloud/aiplatform/experimental/vertex_model/serializers/model.py
```diff
@@ -0,0 +1,118 @@
...
+        obj (torch.nn.Module): the model to serialize.
+        dataset_type (str): the model name and usage
+
+    Returns:
+        The GCS path pointing to the serialized objet.
+    """
+
+    compiled_custom_model = torch.jit.script(obj)
+
+    if artifact_uri[0:6] != "gs://":
```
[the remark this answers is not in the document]
sasha-gitg: This line seems like it needs to be updated.

google/cloud/aiplatform/experimental/vertex_model/serializers/model.py
```diff
@@ -0,0 +1,118 @@
...
+        path_to_model = artifact_uri + "/my_" + model_type + "_model.pth"
+        torch.jit.save(compiled_custom_model, path_to_model)
+        return path_to_model
+
+    with tempfile.TemporaryDirectory() as tmpdirname:
```
sasha-gitg, line 46: It seems like this block could also call `serialize_to_tmp_and_copy_to_gcs`. It might be useful to move that method into a generic utilities module and call it here as well.

google/cloud/aiplatform/experimental/vertex_model/serializers/dataloaders.py
```diff
@@ -15,35 +15,256 @@
...
+    destination_bucket: storage.Bucket,
+    blob_prefix: str,
+    serialize_fn: Callable[[str], None],
+):
+    with tempfile.TemporaryDirectory() as tmpdirname:
+        temp_dir = pathlib.Path(tmpdirname) / file_name
+        tmp_dir_path = pathlib.Path(temp_dir)
+        serialize_fn(temp_dir)
+
+        local_file_name = tmp_dir_path.name
```
sasha-gitg, line 260: This can be just be `blob_path = tmp_dir_path.name`

google/cloud/aiplatform/experimental/vertex_model/serializers/dataloaders.py
sasha-gitg: Yes, we can provide guidance that credentials provided to `aiplatform.init` should have permissions to access remote data sources referenced by objects that are passed to `fit`.

In most cases this should work using Application Default Credentials.

PR #578, opened 2021-07-29
chore: Update testing configs and structure for speed

vinnysenthil:
Implementing planned changes described in [go/vertex-testing](http://go/vertex-testing)


samples/snippets/requirements.txt
```diff
@@ -1,3 +0,0 @@
-pytest==6.2.4
```
busunkim96, line 1: Is this file being deleted intentionally?

busunkim96 (approved):

samples/snippets/requirements.txt
vinnysenthil: Yup - moved it to `requirements-test.txt` ([see here](https://github.com/googleapis/python-aiplatform/pull/578/files#diff-bc8fe5bfdfb4b4373668f263f7ddeccc0a02466f3f1aa9ffc29bfad5c3f3cb74R1))

busunkim96: Ah I was more curious about `google-cloud-storage` and `google-cloud-aiplatform`. The noxfile is looking for these file names specifically (in the same directory as the samples) so you might see issues with dependencies not getting installed.

https://github.com/googleapis/python-aiplatform/blob/b478075efb05553760514256fee9a63126a9916f/samples/snippets/noxfile.py#L152-L158

vinnysenthil: The reason for removing those was because the following lines in the noxfile install the `google-cloud-aiplatform` library from source which also includes `google-cloud-storage` as [a required dependency](https://github.com/googleapis/python-aiplatform/blob/b478075efb05553760514256fee9a63126a9916f/setup.py#L70). We have `INSTALL_LIBRARY_FROM_SOURCE` set to `True` in all the sample Kokoro configs.
https://github.com/googleapis/python-aiplatform/blob/b478075efb05553760514256fee9a63126a9916f/samples/snippets/noxfile.py#L160-L161

busunkim96: ^ Yep including the sample dependencies in `requirements.txt` is useful for local runs and any external developers.

vinnysenthil: Sounds good, updated!

dizcology:
Looks like coverage is failing with “Coverage failure: total of 96 is less than fail-under=99”.

vinnysenthil:
@dizcology yup, I'm trying to get away with testing just one or two Python versions in the presubmit to reduce execution time to 10 mins instead of 30 mins. Trying higher versions now, if code coverage still fails I'll return to testing more versions. The plan would still be to test all versions nightly and before release.

vinnysenthil:
Setting `unit-3.8` as the session to test during presubmit which gives 98.7% coverage (98.8% for all 4 Py versions). In the event coverage fails in a future PR - adding `unit-3.7` and/or `unit-3.6` in the config below should resolve things:
https://github.com/googleapis/python-aiplatform/blob/2e84d0b0d76cbd672997159ff00766e7f9038722/.kokoro/presubmit/presubmit.cfg#L4-L7

PR #579, opened 2021-07-29
fix: change default replica count to 1 for custom training job classes

morgandu:
Fixes b/193437435


sasha-gitg (approved):
LGTM! Thanks Morgan!

PR #583, opened 2021-08-02
feat: add support for export_evaluated_data_items_config in AutoMLTab…

geraint0923:
add support for export_evaluated_data_items_config in AutoMLTabularTrainingJob

### DIFF
--- a/google/cloud/aiplatform/training_jobs.py
+++ b/google/cloud/aiplatform/training_jobs.py
@@ -2705,6 +2705,9 @@ class AutoMLTabularTrainingJob(_TrainingJob):
         budget_milli_node_hours: int = 1000,
         model_display_name: Optional[str] = None,
         disable_early_stopping: bool = False,
+        export_evaluated_data_items: bool = False,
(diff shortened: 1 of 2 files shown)


sasha-gitg (approved):
LGTM!

geraint0923:
Thanks! @sasha-gitg  Do you mind adding an `auto-merge` to this PR since I don't have the permission to merge in this repo?

dparkar:
@sasha-gitg (Going by the the doc) Will this will only export it to a big query table ? Would it be possible to write a csv to cloud storage ? Or is there another work item planned for that ?

geraint0923:
> @sasha-gitg (Going by the the doc) Will this will only export it to a big query table ? Would it be possible to write the predictions as a csv to cloud storage ? Or is there another work item planned for that ?

@dparkar Thanks for reaching out. We currently don't have a plan to support GCS output for this export feature since CSV doesn't have a straightforward support for the complex data types like array and struct.

PR #586, opened 2021-08-02
feat: expose base_output_dir for custom job

morgandu:
b/193822185


google/cloud/aiplatform/jobs.py
```diff
@@ -59,6 +59,8 @@
     study as gca_study_compat,
 )
 
+from google.cloud.aiplatform.utils import _timestamped_gcs_dir
```
sasha-gitg, line 62: This module is imported in the block above.

sasha-gitg (approved):
LGTM! Thanks Morgan!

PR #588, opened 2021-08-03
feat: Added the VertexAiResourceNoun.to_dict() method

Ark-kun:
This method allows getting the resource metadata as a standard Python dictionary.
Without this method, any user who would want to do this would have to access private fields.

### DIFF
--- a/google/cloud/aiplatform/base.py
+++ b/google/cloud/aiplatform/base.py
@@ -44,6 +44,7 @@ from google.auth import credentials as auth_credentials
 from google.cloud.aiplatform import initializer
 from google.cloud.aiplatform import utils
 from google.cloud.aiplatform.compat.types import encryption_spec as gca_encryption_spec
+from google.protobuf import json_format
 
 logging.basicConfig(level=logging.INFO, stream=sys.stdout)
 
@@ -606,6 +607,10 @@ class VertexAiResourceNoun(metaclass=abc.ABCMeta):
 
     def __repr__(self) -> str:
         return f"{object.__repr__(self)} \nresource name: {self.resource_name}"
+
+    def to_dict(self) -> dict:
+        """Returns the resource proto as a dictionary."""
+        return json_format.MessageToDict(self.gca_resource._pb)
 
 
 def optional_sync(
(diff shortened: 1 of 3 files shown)


google/cloud/aiplatform/base.py
```diff
@@ -607,6 +608,10 @@ def _assert_gca_resource_is_available(self) -> None:
     def __repr__(self) -> str:
         return f"{object.__repr__(self)} \nresource name: {self.resource_name}"
 
+    def to_dict(self) -> dict:
```
sasha-gitg, line 611: Please add a test case invoking this method.

sasha-gitg: I think the return type can be further qualified `Dict[str, Any]`

sasha-gitg (approved):

google/cloud/aiplatform/base.py
Ark-kun: Do you think it's a good idea to replace some of the MessageToDict usage cases with this new method? Then it will be auto-tested. https://github.com/googleapis/python-aiplatform/search?q=MessageToDict

Ark-kun: Done.
I've switched .metadata property to use `.to_dict()` and also switched one of the dataset tests.

ivanmkc (approved):
LGTM

Just wondering, what are the use cases are for exposing the internal resource metadata like this. Would it be better to create getters for each field we want to expose?

Ark-kun:
> Just wondering, what are the use cases are for exposing the internal resource metadata like this. Would it be better to create getters for each field we want to expose?

The aiplatform module is higher-level compared to the underlying API and gapic. Not everything is exposed.
I'm not sure manually adding all properties is fully maintainable.
In my case I wanted to return the full resource description in an industry standard JSON format that can easily be parsed and queried by any downstream tools.
When you want to "pass" a small resource description between different containers or components, using a format like JSON has many portability benefits. Adding `.to_dict()` makes it easy to "export" and serialize the resource description.

vinnysenthil (approved):

Ark-kun:
It looks like one of the tests has flaked. Is it possible to retry it?

PR #590, opened 2021-08-03
fix: Fixed the column name corruption bug in TabularDataset.column_names

### LINKED ISSUE #589: Weird bug in TabularDataset.column_names
I got very weird issue.
I've imported well-known Retail Stockout prediction dataset in CSV format. I've imported the dataset to the Vertex AI Datasets using google.cloud.aiplatform.TabularDataset python library code.

Most columns have the "Wk_" prefix. The screenshot shows that there is only one column with "2016_43_Quantity" in it - "Wk_2016_43_Quantity" column. Just like in the source CSV.
Everything is fine.

But here is the problem:
When I call the API to get the dataset metadata including the column names, all column names are fine except one which is stated as "WWk_2016_43_Quantity". (Notice the double "W" in the "WWk_" prefix).
In context:
...
 'Wk_2016_42_Quantity',
 'WWk_2016_43_Quantity',
 'Wk_2016_44_Quantity',
...

This discrepancy causes the subsequent model training to fail due to the dataset not having the `WWk_2016_43_Quantity` column (it has `Wk_2016_43_Quantity` instead).

I do not understand how this could have happened, but you can easily examine the imported dataset and see that the UX and and what returned by the google-cloud-aiplatform library differs.

#### Environment details

  - OS type and version: Linux
  - Python version: 3.7
  - `google-cloud-aiplatform` version: 1.1.1

#### Steps to reproduce

  1. Create dataset from the "gs://kubeflow-pipelines-regional-us-central1/mirror/cloud-ml-data/automl-tables/notebooks/stockout.csv" file
  2. Try getting its columns

#### Code example

```python
print(aiplatform.TabularDataset('projects/140626129697/locations/us-central1/datasets/2405036550225133568').column_names)
```

Ark-kun:
The `end` parameter of the [blob.download_as_bytes](https://googleapis.dev/python/storage/latest/blobs.html#google.cloud.storage.blob.Blob.download_as_bytes) function is inclusive, not exclusive. 

Fixes https://github.com/googleapis/python-aiplatform/issues/589

> There are 2 hard problems in computer science: cache invalidation, naming things, and off-by-1 errors.

Fixes #589

### DIFF
--- a/google/cloud/aiplatform/datasets/tabular_dataset.py
+++ b/google/cloud/aiplatform/datasets/tabular_dataset.py
@@ -150,7 +150,7 @@ class TabularDataset(datasets._Dataset):
 
             while first_new_line_index == -1:
                 line += blob.download_as_bytes(
-                    start=start_index, end=start_index + increment
+                    start=start_index, end=start_index + increment - 1
                 ).decode("utf-8")
 
                 first_new_line_index = line.find("\n")


sasha-gitg (approved):
LGTM! Thanks Alexey!

ivanmkc (approved):
LGTM

PR #594, opened 2021-08-04
feat: cloud training to local prediction

anooshrees:


google/cloud/aiplatform/experimental/vertex_model/base.py
```diff
@@ -147,23 +155,54 @@ def f(*args, **kwargs):
...
     return f
 
 
+def vertex_predict_function_wrapper(method: Callable[..., Any]):
+    @functools.wraps(method)
+    def p(*args, **kwargs):
+        obj = method.__self__
+
+        # Local predictions can be made
+        if method.__self__.training_mode == "local" or obj._model is None:
```
sasha-gitg, line 183: I would perhaps log a warning ("training_mode is 'cloud' but fit has not been called, running predict directly")  or throw if `training_mode` is `cloud` and obj._model is `None`.

google/cloud/aiplatform/experimental/vertex_model/base.py
```diff
@@ -147,23 +155,54 @@ def f(*args, **kwargs):
...
+            return method(*args, **kwargs)
+
+        # Assuming only local prediction for now
+        output_dir = obj._model._gca_resource.artifact_uri
+        model_uri = pathlib.Path(output_dir) / (
+            "my_" + obj.training_mode + "_model.pth"
+        )
+
+        my_model = model._deserialize_remote_model(str(model_uri))
```
sasha-gitg, line 192: To avoid deserializing at every predict call, you can assign my_model as an instance attribute.

google/cloud/aiplatform/experimental/vertex_model/utils/source_utils.py
```diff
@@ -119,4 +122,10 @@ def _make_source(
 
         src = src + ")\n"
 
+    if obj.training_mode == "cloud":
```
sasha-gitg, line 125: I think this conditional can be skipped as this will always be in the cloud.

sasha-gitg (approved):

PR #603, opened 2021-08-11
feat: Automatically detect dependencies from user code

anooshrees:
- Added class attribute for user to designate libraries
- Currently adding detection in source script


google/cloud/aiplatform/experimental/vertex_model/base.py
```diff
@@ -151,15 +151,19 @@ def f(*args, **kwargs):
             with open(script_path, "w") as f:
                 f.write(source)
 
+            # Handle dependencies by first checking for user requirements
+            if not obj._dependencies:
+                obj._dependencies = [
```
sasha-gitg, line 156: I think we can keep our default dependencies as a class attribute on the base class.

```
class VertexModel:
    _dependencies = (...,)
```

And allow the user to override. That way we don't need to perform this check.

It also makes the the default dependencies more transparent to the user.

google/cloud/aiplatform/experimental/vertex_model/utils/source_utils.py
```diff
@@ -20,6 +20,27 @@
 from typing import Dict
 from typing import Tuple
 
+import ast
+from collections import namedtuple
+
+Import = namedtuple("Import", ["module", "name", "alias"])
+
+
+def get_imports(path):
```
sasha-gitg, line 29: It seems like we are parsing the imports and then composing them together again. It may be simpler to use the nodes directly to create the import lines:

Something like:

```Python
import ast

def get_import_lines(path):
    with open(path) as f:
        root = ast.parse(f.read(), path)

    for node in ast.iter_child_nodes(root):
        line = ''
        if isinstance(node, ast.ImportFrom):
            line+=f'from {node.module} '
            
        if isinstance(node, (ast.Import, ast.ImportFrom)):
            line+='import '
            for i, name in enumerate(node.names):
                line+=f'{name.name}'
                if name.asname:
                    line+= f' as {name.asname}'
                if len(node.names) > 0 and i < len(node.names) - 1:
                    line+=', '
        yield line
```

google/cloud/aiplatform/experimental/vertex_model/utils/source_utils.py
```diff
@@ -75,14 +96,43 @@ def _make_source(
         the user has the option to specify a method to call from __main__.
     """
 
+    module = inspect.getmodule(obj.__class__)
```
sasha-gitg, line 99: Double checking that this is supported in Jupyter and Colab environments where the class is defined in the Notebook?

sasha-gitg (changes requested):
Mainly LGTM. A few comments.

google/cloud/aiplatform/experimental/vertex_model/utils/source_utils.py
anooshrees: Let me double check in the Jupyter notebook before pushing this code.

google/cloud/aiplatform/experimental/vertex_model/utils/source_utils.py
anooshrees: got it, thank you!

PR #631, opened 2021-08-17
fix: Update BatchPredictionJob.iter_outputs() and BQ docstrings

vinnysenthil:
Fixes [b/191995461](http://b/191995461) and [b/186764629](http://b/186764629)  🦕


sasha-gitg (approved):
Thanks Vinny!

PR #649, opened 2021-08-20
fix: pipeline none values

ji-yaqi:


google/cloud/aiplatform/utils/pipeline_utils.py
```diff
@@ -129,7 +129,7 @@ def _get_vertex_value(
           ValueError: if the parameter name is not found in pipeline root
           inputs, or value is none.
         """
-        if not value:
+        if value == None:
```
chensun, line 132: > Always use `if foo is None`: (or `is not None`) to check for a None value.

https://google.github.io/styleguide/pyguide.html#2144-decision

Also see: https://stackoverflow.com/a/26611

sasha-gitg (approved):

vinnysenthil (approved):

PR #655, opened 2021-08-25
feat: Lazy load Endpoint class

vinnysenthil:
_In Progress_

Fixes [b/194435202](http://b/194435202) 🦕

### DIFF
--- a/google/cloud/aiplatform/models.py
+++ b/google/cloud/aiplatform/models.py
@@ -79,6 +79,7 @@ class Endpoint(base.VertexAiResourceNounWithFutureManager):
 
     client_class = utils.EndpointClientWithOverride
     _is_client_prediction_client = False
+    _skipped_getter_call = True  # get_endpoint() has not been called
     _resource_noun = "endpoints"
     _getter_method = "get_endpoint"
     _list_method = "list_endpoints"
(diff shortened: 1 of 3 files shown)


google/cloud/aiplatform/models.py
```diff
@@ -115,12 +115,44 @@ def __init__(
...
+        """Sync GAPIC service representation of Endpoint resource once."""
+        if self._skipped_getter_call:
+            self._sync_gca_resource_if_skipped()
+        else:
+            super()._sync_gca_resource()
+
+    def _assert_gca_resource_is_available(self) -> None:
+        """Ensures Endpoint getter was called at least once before
+        asserting on gca_resource's availability."""
+        self._sync_gca_resource_if_skipped()
```
sasha-gitg, line 153: I wonder if it's worth the tradeoff to introduce the `_skipped_getter_call_flag`. Would it be possible to base that state off the `_gca_resource`? For example, wouldn't a property like `create_time` only be populated if we only executed the getter?

Something like:
```self._sync_gca_resource() if not self._gca_resource.create_time```

If this is possible then we can avoid the overhead of maintaining the state flag.

Also does the order of the executed statements in this method need to be reversed. Will this still throw the informative `_assert_gca_resource_available` error:

```Python
ds = aiplatform.Dataset(..., sync=False)
job = aiplatform.CustomTrainingJob(...)
model = job.run(..., sync=False)
ep = model.deploy(..., sync=False)

print(ep.network)
```

It seems like it may throw a None attribute error from `_sync_gca_resource_if_skipped`

google/cloud/aiplatform/models.py
```diff
@@ -110,12 +110,50 @@ def __init__(
...
+        if self._skipped_getter_call():
+            self._sync_gca_resource_if_skipped()
+        else:
+            super()._sync_gca_resource()
+
+    def _assert_gca_resource_is_available(self) -> None:
+        """Ensures Endpoint getter was called at least once before
+        asserting on gca_resource's availability."""
+        self._sync_gca_resource_if_skipped()
+        super()._assert_gca_resource_is_available()
```
sasha-gitg, line 155: It seems like the implementation can be reduced with the same control flow if

1.  The above lines (154, 155) are switched. ie: Assert the resource is available and then sync if getter was skiiped.

2. Remove the overridden `_sync_gca_resource` method.

sasha-gitg (approved):
Thanks Vinny! I added one comment that may simplify this change.

PR #658, opened 2021-08-26
fix: Populate service_account and network in PipelineJob instead of pipeline_spec

sasha-gitg:


browningjp-datatonic:
Fantastic, thank you!

PR #672, opened 2021-09-01
feat: add explanation metadata `get_metadata_protobuf` for reuse

ji-yaqi:
Add a explanationMetadata Object for re-use.

### DIFF
--- a/google/cloud/aiplatform/explain/metadata/metadata_builder.py
+++ b/google/cloud/aiplatform/explain/metadata/metadata_builder.py
@@ -30,5 +30,5 @@ class MetadataBuilder(_ABC):
         """Returns the current metadata as a dictionary."""
 
     @abc.abstractmethod
-    def get_metadata_object(self):
-        """Returns the current metadata as ExplanationMetadata object"""
+    def get_metadata_protobuf(self):
+        """Returns the current metadata as ExplanationMetadata protobuf"""
--- a/google/cloud/aiplatform/explain/metadata/metadata_builder.py
+++ b/google/cloud/aiplatform/explain/metadata/metadata_builder.py
@@ -28,3 +28,7 @@ class MetadataBuilder(_ABC):
     @abc.abstractmethod
     def get_metadata(self):
         """Returns the current metadata as a dictionary."""
+
+    @abc.abstractmethod
+    def get_metadata_object(self):
+        """Returns the current metadata as ExplanationMetadata object"""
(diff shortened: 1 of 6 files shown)


sasha-gitg (changes requested):
Changing the return type on `get_metadata` is a breaking change. Preference to avoid it if possible.

Another potential solution is expanding the explain parameters to accept a Union of Dict and Explain protos and handling the conversion underneath.

Proto plus seems to be able to accept dictionaries of representation at construction time: https://github.com/googleapis/proto-plus-python/blob/cfd5b6caca3fa9add89d8c69ea620505dd90dd7c/docs/messages.rst#serialization

What is the exception that is thrown if the dictionary representation is passed in?

ji-yaqi:
> Changing the return type on `get_metadata` is a breaking change. Preference to avoid it if possible.
> 
> Another potential solution is expanding the explain parameters to accept a Union of Dict and Explain protos and handling the conversion underneath.
> 
> Proto plus seems to be able to accept dictionaries of representation at construction time: https://github.com/googleapis/proto-plus-python/blob/cfd5b6caca3fa9add89d8c69ea620505dd90dd7c/docs/messages.rst#serialization
> 
> What is the exception that is thrown if the dictionary representation is passed in?

I think we are not changing the return type on `get_metadata`, but adding an additional `get_metadata_object` to return the real object. It's mainly type changes.

sasha-gitg:
Got it. Disregard the breaking change statement.

Would like to follow up on the exception thrown when passing the dictionary instead of the proto.

ji-yaqi:
> Got it. Disregard the breaking change statement.
> 
> Would like to follow up on the exception thrown when passing the dictionary instead of the proto.

I got the following: `Parameter to MergeFrom() must be instance of same class: expected google.cloud.aiplatform.v1beta1.ExplanationMetadata.InputMetadata got dict.`

google/cloud/aiplatform/explain/metadata/metadata_builder.py
```diff
@@ -28,3 +28,7 @@ class MetadataBuilder(_ABC):
     @abc.abstractmethod
     def get_metadata(self):
         """Returns the current metadata as a dictionary."""
+
+    @abc.abstractmethod
+    def get_metadata_object(self):
```
sasha-gitg, line 33: It seems like this can be informative if further qualified as ```get_metadata_protobuf``` or ```get_metadata_pb```.

ji-yaqi: Thanks, changed to `get_metadata_protobuf`

sasha-gitg (approved):

andrewferlitsch (approved):

vinnysenthil (approved):

PR #685, opened 2021-09-03
feat: Update container specification based on dependencies and GPU usage.

anooshrees:


google/cloud/aiplatform/experimental/vertex_model/base.py
```diff
@@ -237,20 +238,33 @@ def f(*args, **kwargs):
             if GITHUB_DEPENDENCY not in obj.dependencies:
                 obj.dependencies.append(GITHUB_DEPENDENCY)
 
+            # Right now, only the TensorFlow containers accomodate GPU usage.
+            if any("tensorflow" in dependency for dependency in obj.dependencies):
```
sasha-gitg, line 242: The selection of the container requires further specification. 

Pytorch and Tensorflow both provide GPU enabled containers: https://cloud.google.com/vertex-ai/docs/training/pre-built-containers#pytorch

So we should also check for Pytorch in the dependencies. 

Also we should map to the latest available version since we are not checking version. 2.6 for TF and 1.9 for Pytorch.

We should also check for the region set in `aiplatform.initializer.global_config.location`:

Something like:

```Python
location_prefix = location.split('-')[0]
container_location =  location_prefix if location_prefix in ('us', 'europe', 'asia') else 'us'
container_uri = f'{container_location}-docker.pkg.dev/...'
```

sasha-gitg (changes requested):

google/cloud/aiplatform/experimental/vertex_model/README.rst
```diff
@@ -136,7 +136,12 @@ When performing remote training and/or prediction with your code, you have the o
 .. code-block:: Python
 
    my_model = MyModelClass()
+   
    my_model.machine_type = "n1-standard-4"
+   replica_count = 1
```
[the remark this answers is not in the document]
anooshrees: correct, apologies for the typo.

google/cloud/aiplatform/experimental/vertex_model/base.py
```diff
@@ -234,23 +243,68 @@ def f(*args, **kwargs):
...
 
+            # Account for user-designated dependencies when
+            # setting up remote prediction
             if GITHUB_DEPENDENCY not in obj.dependencies:
                 obj.dependencies.append(GITHUB_DEPENDENCY)
 
+            dependency_installs = []
+            for dependency in obj.dependencies:
+                if dependency != GITHUB_DEPENDENCY:
+                    dependency_name = re.split(">|=|<", dependency)[0]
```
sasha-gitg, line 254: It seems like we should keep the version of the dependency.

google/cloud/aiplatform/experimental/vertex_model/base.py
```diff
@@ -234,23 +243,68 @@ def f(*args, **kwargs):
...
+            location_prefix = aiplatform.initializer.global_config.location.split("-")[
+                0
+            ]
+            container_location = (
+                location_prefix if location_prefix in ("us", "europe", "asia") else "us"
+            )
+
+            if any("tensorflow" in dependency for dependency in obj.dependencies):
+                if obj.accelerator_count > 0:
+                    obj.container_uri = f"{container_location}-docker.pkg.dev/vertex-ai/prediction/tf2-gpu.2-6:latest"
```
sasha-gitg, line 280: Is `container_uri` a field we intend the user to set? If it's just part of the implementation then I wouldn't assign to the object.

sasha-gitg: It seems like this should be a training container.

google/cloud/aiplatform/experimental/vertex_model/base.py
```diff
@@ -234,23 +243,68 @@ def f(*args, **kwargs):
...
+            )
+
+            # Container specification
+            location_prefix = aiplatform.initializer.global_config.location.split("-")[
+                0
+            ]
+            container_location = (
+                location_prefix if location_prefix in ("us", "europe", "asia") else "us"
+            )
+
```
sasha-gitg, line 277: Please open a ticket and add a TODO here for matching dependency version to container version.

google/cloud/aiplatform/experimental/vertex_model/base.py
```diff
@@ -379,10 +452,15 @@ class VertexModel(metaclass=abc.ABCMeta):
     dependencies = [
         "pandas>=1.3",
         "torch>=1.7",
-        "google-cloud-aiplatform @ git+https://github.com/googleapis/python-aiplatform@refs/pull/659/head#egg=google-cloud-aiplatform",
+        GITHUB_DEPENDENCY,
     ]
 
+    container_uri = "us-docker.pkg.dev/vertex-ai/training/scikit-learn-cpu.0-23:latest"
```
sasha-gitg, line 458: Do we intend the user to use/set this? If not, we shouldn't expose it publicly.

google/cloud/aiplatform/experimental/vertex_model/base.py
```diff
@@ -379,10 +452,15 @@ class VertexModel(metaclass=abc.ABCMeta):
     dependencies = [
         "pandas>=1.3",
```
sasha-gitg, line 453: Can we remove the pandas and torch dependencies?

google/cloud/aiplatform/experimental/vertex_model/base.py
```diff
@@ -379,10 +452,15 @@ class VertexModel(metaclass=abc.ABCMeta):
...
         "torch>=1.7",
-        "google-cloud-aiplatform @ git+https://github.com/googleapis/python-aiplatform@refs/pull/659/head#egg=google-cloud-aiplatform",
+        GITHUB_DEPENDENCY,
     ]
 
+    container_uri = "us-docker.pkg.dev/vertex-ai/training/scikit-learn-cpu.0-23:latest"
+
     machine_type = "n1-standard-4"
+    replica_count = 1
```
sasha-gitg, line 461: These values are class properties but they are being used as instance properties in the README. I would recommend setting these in the constructor below.

sasha-gitg (changes requested):

google/cloud/aiplatform/experimental/vertex_model/base.py
anooshrees: I split around the version here to make the pip install easier. How would I parse the string to install a specific version of each library? I'll look into it right now as well.

google/cloud/aiplatform/experimental/vertex_model/base.py
anooshrees: Okay, I'll rename this training_container and remove it as an object variable.

google/cloud/aiplatform/experimental/vertex_model/base.py
anooshrees: I wanted to keep these as the default, as we were keeping all other pytorch/pandas inclusion as a default as well. Removing it here would mean parsing all of the imports in the notebook to recreate these dependencies.

google/cloud/aiplatform/experimental/vertex_model/base.py
anooshrees: sounds good!

google/cloud/aiplatform/experimental/vertex_model/base.py
sasha-gitg: You should be able to pass the dependency definition name as-is without splitting.

anooshrees: okay, will make this change right now and test it out.

google/cloud/aiplatform/experimental/vertex_model/base.py
```diff
@@ -234,23 +243,70 @@ def f(*args, **kwargs):
...
+            ]
+            container_location = (
+                location_prefix if location_prefix in ("us", "europe", "asia") else "us"
+            )
+
+            training_container = f"{container_location}-docker.pkg.dev/vertex-ai/training/scikit-learn-cpu.0-23:latest"
+
+            if any("tensorflow" in dependency for dependency in obj.dependencies):
+                if obj.accelerator_count > 0:
+                    training_container = f"{container_location}-docker.pkg.dev/vertex-ai/prediction/tf2-gpu.2-6:latest"
```
sasha-gitg, line 284: Shouldn't this be the training instead of prediction?

sasha-gitg (changes requested):

sasha-gitg (approved):
LGTM! Thank you!

PR #690, opened 2021-09-07
fix: library should be released as 1.4.1

dizcology:
Release-As: 1.4.1


sasha-gitg (approved):
May want to change `fix` to `chore` so this commit doesn't show up in the release notes.

PR #695, opened 2021-09-08
fix: Remove sync parameter from create_endpoint_sample

aribray:
Remove unused `sync` parameter from `create_endpoint_sample`

Googlers see 195511018


ivanmkc:
You can pass the `sync` param into the `Endpoint.create` call. I see other samples are written like that, so might want to follow that to be consistent: https://github.com/googleapis/python-aiplatform/blob/main/samples/model-builder/create_training_pipeline_tabular_classification_sample.py

PR #699, opened 2021-09-10
Feat: Add debugging terminal support for CustomJob, HyperparameterTun…

morgandu:

Fixes #<b/195449603> 🦕


google/cloud/aiplatform/jobs.py
```diff
@@ -1219,13 +1219,60 @@ def from_local_script(
...
+        if self.state in _JOB_RUNNING_STATE:
+            _LOGGER.info(
+                "%s %s access the interactive shell terminals for this job at the following links:\n%s"
+                % (
+                    self.__class__.__name__,
+                    self._gca_resource.name,
+                    "\n".join(
+                        [
+                            "%s:\n%s" % (worker, uri)
+                            for worker, uri in self._gca_resource.web_access_uris.items()
```
sasha-gitg, line 1262: Can we rely on the presence of these uris in the resource proto as the condition to log them so we don't have to maintain more than one polling block?

Something like this in the main polling block:
```Python
  has_logged_web_access_uri = False
  while self.state not in _JOB_COMPLETE_STATES:
      current_time = time.time()
      if current_time - previous_time >= log_wait:
          self._log_job_state()
          log_wait = min(log_wait * multiplier, max_wait)
          previous_time = current_time
      if self._gca_resource.web_access_uris and not has_logged_web_access_uris:
        self._log_web_access_uris()
      time.sleep(wait)
```

google/cloud/aiplatform/jobs.py
```diff
@@ -1219,13 +1219,60 @@ def from_local_script(
             staging_bucket=staging_bucket,
         )
 
+    def web_access_uris(self):
```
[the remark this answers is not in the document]
sasha-gitg: Recommend either:
1. renaming to `log_web_access_uris`
or 
2. `get_web_access_uris` and return `web_access_uris` dictionary directly.

Preference for (2) so users can programmatically interact with the uris if needed.

sasha-gitg (changes requested):

google/cloud/aiplatform/jobs.py
morgandu: Do you mean the `def _block_until_complete(self)` from `_Job` class as the main polling block? I was going with modifying the `def _block_until_complete(self)` utilizing the `web_access_uris` present in resource proto initially. Here are the reasons I decided not to:
1) It works well for CustomJob class when modifying the `def _block_until_complete(self)`. However, for `HyperparameterTuningJob`, the pulling is needed at the trial level instead of the job level, so we will need a customized `def _block_until_complete(self)` for HP class. 
2) For `Custom*Training` classes, their `def _block_until_complete(self)` is at training pipeline level, where a separate pulling to its `backingCustomJob` will be needed, and it will be outside the Job's `def _block_until_complete(self)`. 
So all these will end up having more duplicated logic around `def _block_until_complete(self)`. I figure it will be cleaner to have 2 pulling block but with different function, and they can be reused whenever is needed.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -1429,9 +1437,36 @@ def _prepare_training_task_inputs_and_output_dir(
             training_task_inputs["network"] = network
         if tensorboard:
             training_task_inputs["tensorboard"] = tensorboard
+        if enable_web_access:
+            training_task_inputs["enable_web_access"] = enable_web_access
 
         return training_task_inputs, base_output_dir
 
+    def _get_custom_job(self) -> "jobs.CustomJob":
```
sasha-gitg, line 1445: Why does the  return type need to be a forward reference.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -1429,9 +1437,36 @@ def _prepare_training_task_inputs_and_output_dir(
...
+        """Helper method to get backing custom job
+
+        Returns:
+            A CustomJob.
+        """
+        custom_job_resource_name = self._gca_resource.training_task_metadata.get(
```
sasha-gitg, line 1451: This could return `None` which would  cause the  line below to throw.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -1445,6 +1480,18 @@ def _wait_callback(self):
 
             self._has_logged_custom_job = True
 
+        if (
+            self._gca_resource.training_task_metadata
+            and self._gca_resource.training_task_metadata.get("backingCustomJob")
+            and self._gca_resource.training_task_inputs.get("enable_web_access")
+            and not self._has_logged_web_access_uris
```
sasha-gitg, line 1487: Is it possible the `web_access_uris` have changed throughout the run? If, for example, one of the workers failed.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -1744,6 +1791,7 @@ def run(
         test_filter_split: Optional[str] = None,
         predefined_split_column_name: Optional[str] = None,
         timestamp_split_column_name: Optional[str] = None,
+        enable_web_access: Optional[bool] = False,
```
sasha-gitg, line 1794: Since `enable_web_access` defaults to `False` it seems that this doesn't need to be `Optional`.

sasha-gitg (changes requested):

google/cloud/aiplatform/training_jobs.py
morgandu: The current service setup is that if workers failed and restarted, the `web_access_uris` will redirect to the new workers, but won't change itself.

google/cloud/aiplatform/jobs.py
```diff
@@ -1219,13 +1267,39 @@ def from_local_script(
...
         )
 
+    def _get_web_access_uris(self) -> Dict[str, str]:
+        """Helper method to get web access uris of the custom job
+
+        Returns:
+            (Dict[str, str]) - web access uris of the custom job
+        """
+        self._sync_gca_resource()
+        return self._gca_resource.web_access_uris
```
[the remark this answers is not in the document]
morgandu: It works the same as `labels`, I see we have `labels` as it is and it is a `dict`

google/cloud/aiplatform/jobs.py
```diff
@@ -1046,6 +1104,29 @@ def network(self) -> Optional[str]:
         self._assert_gca_resource_is_available()
         return self._gca_resource.job_spec.network
 
+    def _get_web_access_uris(self):
+        """Helper method to get the web access uris of the custom job"""
+        self._web_access_uris = self._gca_resource.web_access_uris
```
sasha-gitg, line 1109: Based on the usage of this method, it seems like we can remove the `_web_access_uris` property and return `web_access_uris` in this method.

So usage is:
```Python
web_access_uri = self._get_web_access_uris()

for worker, uri in web_access_uris.items():
  ...
```

google/cloud/aiplatform/jobs.py
```diff
@@ -845,6 +843,56 @@ def __init__(
             project=project, location=location
         )
 
+    @abc.abstractmethod
+    def _get_web_access_uris(self) -> Dict[str, str]:
+        """Helper method to get web access uris for the runnable job"""
+        pass
+
+    @property
+    def web_access_uris(self) -> Dict[str, str]:
```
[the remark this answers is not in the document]
sasha-gitg: This return type hint is inconsistent the the hp tuning's `web_access_uris`.

google/cloud/aiplatform/jobs.py
```diff
@@ -1564,13 +1653,45 @@ def network(self) -> Optional[str]:
...
+                if trial.web_access_uris
+            ]
+
+    def _log_web_access_uris(self):
+        """Helper method to log the web access uris of the hyperparameter job"""
+
+        self._get_web_access_uris()
+
+        if self._web_access_uris:
+            for (trial_id, tria_web_access_uris) in self._web_access_uris:
```
sasha-gitg, line 1671:
```suggestion
            for (trial_id, trial_web_access_uris) in self._web_access_uris:
```

google/cloud/aiplatform/jobs.py
```diff
@@ -1564,13 +1653,45 @@ def network(self) -> Optional[str]:
         self._assert_gca_resource_is_available()
         return getattr(self._gca_resource.trial_job_spec, "network")
 
+    def _get_web_access_uris(self):
+        """Helper method to get the web access uris of the hyperparameter job"""
+        if self.trials:
+            self._web_access_uris = [
```
sasha-gitg, line 1659: Perhaps this should be a nested dictionary of trial id -> worker id -> web_access_uris based on the usage below as it captures that trial ids and workers ids are unique.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -1251,6 +1252,9 @@ def __init__(
         # once Custom Job is known we log the console uri and the tensorboard uri
         # this flags keeps that state so we don't log it multiple times
         self._has_logged_custom_job = False
+        self._custom_job = None
+        self._web_access_uris = None
+        self._logged_web_access_uris = []
```
sasha-gitg, line 1257: set here as well

google/cloud/aiplatform/training_jobs.py
```diff
@@ -1429,9 +1438,40 @@ def _prepare_training_task_inputs_and_output_dir(
...
+    @property
+    def web_access_uris(self) -> Optional[Dict[str, str]]:
+        """Get the web access uris of the backing custom job.
+
+        Returns:
+            (Dict[str, str]):
+                Web access uris of the backing custom job.
+        """
+        if self._custom_job:
+            self._web_access_uris = self._custom_job.web_access_uris
```
sasha-gitg, line 1455: I don't think we need to assign `_web_access_uris` and maintain property since we are returning the value.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -1429,9 +1438,40 @@ def _prepare_training_task_inputs_and_output_dir(
...
 
+    @property
+    def web_access_uris(self) -> Optional[Dict[str, str]]:
+        """Get the web access uris of the backing custom job.
+
+        Returns:
+            (Dict[str, str]):
+                Web access uris of the backing custom job.
+        """
+        if self._custom_job:
```
sasha-gitg, line 1454: Since this property isn't used by the logging logic, it most likely needs to instantiate the custom job instead of reference `self._custom_job` as it might be used after instantiating an already running training job.

ie
```Python
my_job = aiplatform.CustomTrainingJob.get(...)
my_job.state # RUNNING
my_job.web_access_uris
```

google/cloud/aiplatform/training_jobs.py
```diff
@@ -1445,6 +1485,19 @@ def _wait_callback(self):
...
+        ):
+            if not self._custom_job:
+                custom_job_resource_name = self._gca_resource.training_task_metadata.get(
+                    "backingCustomJob"
+                )
+                self._custom_job = jobs.CustomJob.get(
+                    resource_name=custom_job_resource_name
+                )
+            self._web_access_uris = self._custom_job.web_access_uris
+            self._log_web_access_uris()
```
sasha-gitg, line 1499: Perhaps we can pass in `web_access_uris` so we don't need to maintain `self._web_access_uris` property.

google/cloud/aiplatform/jobs.py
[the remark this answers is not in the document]
morgandu: ahhh, noted

google/cloud/aiplatform/jobs.py
```diff
@@ -1046,6 +1101,26 @@ def network(self) -> Optional[str]:
...
+        Returns:
+            (Dict[str, str]):
+                Web access uris of the custom job.
+        """
+        return dict(self._gca_resource.web_access_uris)
+
+    def _log_web_access_uris(self):
+        """Helper method to log the web access uris of the custom job"""
+
+        for worker, uri in self.web_access_uris.items():
```
sasha-gitg, line 1116:
```suggestion
        for worker, uri in self._get_web_access_uris.items():
```
To avoid the additional sync call.

google/cloud/aiplatform/jobs.py
```diff
@@ -1588,13 +1671,46 @@ def network(self) -> Optional[str]:
...
+        for trial in self.trials:
+            web_access_uris[trial.id] = web_access_uris.get(trial.id, dict())
+            for worker, uri in trial.web_access_uris.items():
+                web_access_uris[trial.id][worker] = uri
+        return web_access_uris
+
+    def _log_web_access_uris(self):
+        """Helper method to log the web access uris of the hyperparameter job"""
+
+        for (trial_id, trial_web_access_uris) in self.web_access_uris.items():
```
sasha-gitg, line 1691:
```suggestion
        for trial_id, trial_web_access_uris in self._get_web_access_uris.items():
```

PR #710, opened 2021-09-15
feat: Updating the Tensorboard uploader to use the new batch write API so it runs more efficiently

yfang1:
b/190431961.
go/tb-uploader-speedup

- The BatchedRequestSenders were sending requests for each run - now changed to batch for all runs in the experiment.
- Generalizing the existing _TimeSeriesResourceManager to _OnePlatformResourceManager such that it manages caching for runs as well, in addition to time series.
- Updating the scalar and blob request sizes to 128kb so that it's in sync with the OSS TB (https://github.com/tensorflow/tensorboard/blob/29cfcec9d15c47d1bf54b15df24e65bc38c188e6/tensorboard/uploader/server_info.py#L51)
- Changing uploader_main to call uploader.start_uploading() only, which inherently handles whether the upload was one_shot or not.

Ran the new code for scalar, tensor, blob data. Looks fine to me.

### DIFF
--- a/google/cloud/aiplatform/tensorboard/uploader.py
+++ b/google/cloud/aiplatform/tensorboard/uploader.py
@@ -16,6 +16,7 @@
 #
 """Uploads a TensorBoard logdir to TensorBoard.gcp."""
 import abc
+from collections import defaultdict
 import contextlib
 import functools
 import json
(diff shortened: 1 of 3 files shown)


google/cloud/aiplatform/tensorboard/uploader.py
```diff
@@ -678,25 +623,82 @@ def dispatch_requests(
...
-        self._run_resource_id = run_resource_id
+        self._experiment_resource_name = experiment_resource_name
         self._api = api
-        self._tag_to_time_series_proto: Dict[
-            str, tensorboard_time_series.TensorboardTimeSeries
-        ] = {}
+        self._run_name_to_run_resource_name: Dict[str, str] = {}
+        self._run_tag_name_to_time_series_name: Dict[(str, str), str] = {}
+
+    def get_run_resource_name(self, run_name: str):
```
karthikv2k, line 642: better to rename it as create_or_get_run_resource_name ?

karthikv2k (approved):

google/cloud/aiplatform/tensorboard/uploader.py
yfang1: It doesn't actually "create" a resource name though, but I see what you're saying.

Gonna leave the name as is, but will add doc string to describe what it does.

vinnysenthil (approved):

PR #717, opened 2021-09-17
fix: use the project id from BQ dataset instead of the default project id

geraint0923:
It is possible that the user specify a different project than the current project as the BQ output destination for batch prediction job.


ivanmkc (approved):
I think this makes sense.

LGTM

geraint0923:
Thanks for approving the PR. Do you mind adding the `auto-merge` tag to the PR so that it can be merged automatically since I don't have the permission to merge?

geraint0923:
@ivanmkc kindly ping:)

PR #719, opened 2021-09-20
fix(PipelineJob): use name as output only field

ji-yaqi:
Previously, `create_time` is used to identify whether a job has been created. Now since we have `job_id` fixed, we could leave `name` out for output only and identify whether a job has been created using `name`.

### DIFF
--- a/google/cloud/aiplatform/pipeline_jobs.py
+++ b/google/cloud/aiplatform/pipeline_jobs.py
@@ -48,9 +48,6 @@ _PIPELINE_ERROR_STATES = set(
     [gca_pipeline_state_v1beta1.PipelineState.PIPELINE_STATE_FAILED]
 )
 
-# Vertex AI Pipelines service API job name relative name prefix pattern.
-_JOB_NAME_PATTERN = "{parent}/pipelineJobs/{job_id}"
-
 # Pattern for valid names used as a Vertex resource name.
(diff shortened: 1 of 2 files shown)


google/cloud/aiplatform/pipeline_jobs.py
```diff
@@ -178,19 +175,18 @@ def __init__(
         )
 
         pipeline_name = pipeline_job["pipelineSpec"]["pipelineInfo"]["name"]
-        job_id = job_id or "{pipeline_name}-{timestamp}".format(
+        self.job_id = job_id or "{pipeline_name}-{timestamp}".format(
```
ivanmkc, line 178: I know this is existing code, but why isn't there a way to get the `job_id` directly from the service instead of deriving it in the client SDK. 

This seems like it could break if the service changes how job ID's are generated.

ivanmkc (approved):
LGTM

google/cloud/aiplatform/pipeline_jobs.py
ji-yaqi: I think `job_id` was provided by the user, or generated by the system and only used during `run` command but not the instance generation process.  That's why we need to store it somewhere, since it wasn't used during the instance creation. 

`Name` shouldn't be passed in since it was an output-only field and was wrongly passed in previously. Although the services generated the name using the same way.

PR #727, opened 2021-09-23
fix: add async logging for pipeline job

ji-yaqi:


google/cloud/aiplatform/pipeline_jobs.py
```diff
@@ -240,6 +240,12 @@ def run(
 
         _LOGGER.log_create_with_lro(self.__class__)
 
+        if not sync:
```
sasha-gitg, line 243: This block will not be executed because `sync` is currently not passed by the `base.optional_sync` decorator.

`sync` in the function signature for these methods is mainly provided as documentation for the caller and not currently intended to be used internally.

However, it should be easily supported by changing this line: 

https://github.com/googleapis/python-aiplatform/blob/ac181824ee5afd64640e1790e7ac70809c7cdb1f/google/cloud/aiplatform/base.py#L655

to 

```Python
sync = kwargs.get("sync", True)
```

google/cloud/aiplatform/pipeline_jobs.py
```diff
@@ -240,6 +240,12 @@ def run(
...
+            _LOGGER.info(
+                "Pipeline Job is scheduled to run and will be available to view once created:\n%s"
+                % self._dashboard_uri()
```
sasha-gitg, line 246: This line will always throw because `_dashboard_uri` references `resource_name` which references `self._gca_resource.name` which doesn't exist at this point because the resource hasn't been created. Perhaps it could be synthesized from `self.job_id` but let's discuss offline on why `_dashboard_uri` below isn't printing.

sasha-gitg (changes requested):

PR #734, opened 2021-09-29
feat: add flaky test diagnostic script and fix flaky test

nicain:
Fixes #https://github.com/googleapis/python-aiplatform/issues/580


samples/snippets/helpers.py
```diff
@@ -63,3 +65,21 @@ def wait_for_job_state(
         "\nTry increasing the timeout in sample test"
         f"\nLast recorded state: {response.state}"
     )
+
+def flaky_test_diagnostic(file_name, test_name, N=20):
+    
+    import pytest
```
nicain, line 71: This is the only non-standard-library dependency for this helper function, so I am hiding the import in here so that pytest does not become a dependency for the whole module. I can also inject the module as an arg, or not worry about pytest as a dependency of the module at put it top-level.

samples/snippets/pipeline_service/create_training_pipeline_tabular_regression_sample_test.py
```diff
@@ -34,6 +34,9 @@ def teardown(teardown_training_pipeline):
 
 def test_ucaip_generated_create_training_pipeline_sample(capsys, shared_state):
 
+    # The return of the cancellation can be flaky; max of 20 runs was 215 sec.
+    shared_state["cancel_batch_prediction_job_timeout"] = 300
```
nicain, line 38: Max execution time was 215 seconds out of 20 runs xref: https://github.com/googleapis/python-aiplatform/issues/580#issuecomment-930500944

vinnysenthil (approved):
Looks good - thanks Nick!

PR #741, opened 2021-09-30
feat: enable reduction server

morgandu:
Fixes #<b/197889114> 🦕


google/cloud/aiplatform/utils/worker_spec_utils.py
```diff
@@ -152,17 +161,52 @@ def pool_specs(
...
             else:
-                break
+                last_non_empty_seen = True
         return specs
 
+    def reduction_server_pool(
+        self,
+        reduction_server_replica_count: int = 0,
+        reduction_server_machine_type: str = "n1-highcpu-16",
+    ):
```
ivanmkc, line 183: Return type hint missing

google/cloud/aiplatform/utils/worker_spec_utils.py
```diff
@@ -152,17 +161,52 @@ def pool_specs(
...
+                if not last_non_empty_seen:
+                    specs.pop()
+                else:
+                    specs[i] = {}
             else:
-                break
+                last_non_empty_seen = True
         return specs
 
+    def reduction_server_pool(
```
ivanmkc, line 179: Prefer a method name that starts with verb.

google/cloud/aiplatform/utils/worker_spec_utils.py
```diff
@@ -152,17 +161,52 @@ def pool_specs(
...
+        reduction_server_spec = _WorkerPoolSpec(
+            replica_count=reduction_server_replica_count,
+            machine_type=reduction_server_machine_type,
+            container_uri=REDUCTION_SERVER_CONTAINER_URI,
+        )
+
+        if self.chief_spec.is_empty:
+            raise ValueError("reduction server can not be set without chief worker.")
+
+        return _DistributedTrainingSpec(
```
ivanmkc, line 204: Hm, you are returning a new _DistributedTrainingSpec, but it will hold pointers to the existing `chief_spec`, `chief_spec` since Python passes objects by reference. This seems potentially dangerous and can cause hard to debug situations down the road.

I would say either mutate the existing instance or do a deep copy to create a new instance of `_DistributedTrainingSpec` with no shared pointers.

@sasha-gitg may know a canonical Python way to handle this.

google/cloud/aiplatform/utils/worker_spec_utils.py
```diff
@@ -152,17 +161,52 @@ def pool_specs(
...
+        reduction_server_machine_type: str = "n1-highcpu-16",
+    ):
+        """Parameterizes Config to support only reduction server replicas.
+
+        Args:
+            reduction_server_replica_count (int):
+                The number of reduction server replicas, default is 0.
+            reduction_server_machine_type (str):
+                The type of machine to use for reduction server, default is `n1-highcpu-16`.
+
```
ivanmkc, line 191: Return section missing.

google/cloud/aiplatform/utils/worker_spec_utils.py
```diff
@@ -152,17 +161,52 @@ def pool_specs(
...
+                    specs[i] = {}
             else:
-                break
+                last_non_empty_seen = True
         return specs
 
+    def reduction_server_pool(
+        self,
+        reduction_server_replica_count: int = 0,
+        reduction_server_machine_type: str = "n1-highcpu-16",
```
ivanmkc, line 182: Does the service allow a value of None for this? If so, prefer to defer the default to the service instead of encoding it in the client SDK.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -2498,6 +2521,8 @@ def run(
         accelerator_count: int = 0,
         boot_disk_type: str = "pd-ssd",
         boot_disk_size_gb: int = 100,
+        reduction_server_replica_count: int = 0,
+        reduction_server_machine_type: str = "n1-highcpu-16",
```
ivanmkc, line 2525: ditto

google/cloud/aiplatform/training_jobs.py
```diff
@@ -1352,7 +1358,15 @@ def _prepare_and_validate_run(
...
-        ).pool_specs
+        )
+
+        if reduction_server_replica_count > 0:
+            worker_pool_specs = worker_pool_specs.reduction_server_pool(
+                reduction_server_replica_count=reduction_server_replica_count,
+                reduction_server_machine_type=reduction_server_machine_type,
+            )
+
+        worker_pool_specs = worker_pool_specs.pool_specs
```
ivanmkc, line 1369: This changes the type of the variable. Although allowable in Python, prefer not to do that for readability purposes.

google/cloud/aiplatform/utils/worker_spec_utils.py
```diff
@@ -86,6 +91,10 @@ def spec_dict(self) -> Dict[str, Union[int, str, Dict[str, Union[int, str]]]]:
             spec["machine_spec"]["accelerator_type"] = accelerator_type
             spec["machine_spec"]["accelerator_count"] = self.accelerator_count
 
+        if self.container_uri == REDUCTION_SERVER_CONTAINER_URI:
```
sasha-gitg, line 94: For now, I think it may be more extendable to add `container_spec` whenever `self.container_uri` is populated.

sasha-gitg (changes requested):

google/cloud/aiplatform/jobs.py
```diff
@@ -1061,6 +1061,9 @@ def from_local_script(
         accelerator_count: int = 0,
         boot_disk_type: str = "pd-ssd",
         boot_disk_size_gb: int = 100,
+        reduction_server_replica_count: Optional[int] = 0,
```
sasha-gitg, line 1064: Since this defaults `0` is doesn't need to be `Optional`.

google/cloud/aiplatform/utils/worker_spec_utils.py
```diff
@@ -21,6 +21,13 @@
     accelerator_type as gca_accelerator_type_compat,
 )
 
+SPEC_ORDERS = {
```
sasha-gitg, line 24:
```suggestion
_SPEC_ORDERS = {
```
Also add a comment above about the purpose of this mapping.

sasha-gitg (changes requested):

PR #746, opened 2021-10-04
feat: Cache Components Arguments

### LINKED ISSUE #745: Granularity for Caching Components
There is not direct way to specify which components should be cached in a pipeline run. the existing components will either all be cached or none of them would be cached.

For the sake of better experimentation, there should be a method/argument to explicitly specify which components to cache, without manually having to edit the compiled json.

dhargopala:
Addeded an Optional `cache_components` argument (in PipelineJob class initializer method) to explicitly specify which components to be cached, instead of caching the entire pipeline.
This is useful for testing especially where caching all components of a pipleline can become a redundant process.

This argument will work only when `enable_caching` is set to `True` , otherwise it will be ignored. In case `enable_caching` is set to `True` then only the component names specified in this list will be cached, rest will not be cached.

Fixes #745


google/cloud/aiplatform/pipeline_jobs.py
```diff
@@ -67,13 +69,26 @@ def _set_enable_caching_value(
           Required. The dictionary of pipeline spec.
      enable_caching (bool):
           Required. Whether to enable caching.
+     cache_components (List[str]):
```
ji-yaqi, line 72: `cache_components` was optional in the PipelineJob instantiation phase, but required here. Could you update this to be optional? Otherwise it would break the existing users.

google/cloud/aiplatform/pipeline_jobs.py
```diff
@@ -67,13 +69,26 @@ def _set_enable_caching_value(
...
+    else:
+        for component in [pipeline_spec["root"]] + list(
+            pipeline_spec["components"].values()
+        ):
+            if "dag" in component:
+                for task_name, task in component["dag"]["tasks"].items():
+                    if task_name in cache_components:
+                        task["cachingOptions"] = {"enableCache": True}
+                    else:
+                        task["cachingOptions"] = {"enableCache": False}
```
ji-yaqi, line 91: Same code was repeated here. Could you make it so we only use the same code once?

ji-yaqi (changes requested):
Thanks for your PR! Left a few comments.

google/cloud/aiplatform/pipeline_jobs.py
dhargopala: Reduced the redundant code a bit, however cannot keep the inner `for` loop common because in one scenario we're using `values` method and in the other `items` method.
The way to deal with further logic is also distinct, and if we try to keep the conditional inside the `for` loop, it will use up additional compute each time we iterate through the loop, checking the same condition which essentially won't change.
Do have a look, and let me know if there are any further requirements/changes.

chensun:
> Thank you for your response @chensun. I am using KFP v2 SDK for compilation and I was not aware of this functionality.
> 
> I do understand that tasks and components are not the same thing, I think It used it as a misnomer here.
> 
> Thank you again for the clarification.
> 
> Would that mean there is no need for this feature?

Yes, I think the `set_caching_options` feature from KFP SDK should cover your use case, and it's more robust -- you don't need to know the name of a task (which could be confusing if a component is used multiple times, you'll get `component`, `component-1`, `component-2`, etc. as task names).

So I suggest we drop this PR.

PR #750, opened 2021-10-05
feat: PipelineJob switch to v1 API from v1beta1 API

chensun:


sasha-gitg (approved):
LGTM. 

There is one more level to expose the types as default types here:
https://github.com/googleapis/python-aiplatform/blob/main/google/cloud/aiplatform/compat/__init__.py

This allows importing without explicitly stating the API version which makes switching between versions during dev easier :

ie: 

```Python
from google.cloud.compat.types import pipeline_job
from google.cloud.compat.types import pipeline_state
```

Won't block on this.

chensun:
> LGTM.
> 
> There is one more level to expose the types as default types here: https://github.com/googleapis/python-aiplatform/blob/main/google/cloud/aiplatform/compat/__init__.py
> 
> This allows importing without explicitly stating the API version which makes switching between versions during dev easier :
> 
> ie:
> 
> ```python
> from google.cloud.compat.types import pipeline_job
> from google.cloud.compat.types import pipeline_state
> ```
> 
> Won't block on this.

Thanks! For the safety of the change, will do this next time if we switch version again.

PR #763, opened 2021-10-12
feat: Add training_utils folder and environment_variables

mkovalski:
Add the environment variables from the dev branch and create a training_utils folder to house them. These variables will be used for any new functionality added for use in training scripts.

- Add the training_utils folder
- Create separate environment_varaibles.py file
- Add tests

### DIFF
--- a/samples/model-builder/create_endpoint_sample.py
+++ b/samples/model-builder/create_endpoint_sample.py
@@ -17,7 +17,7 @@ from google.cloud import aiplatform
 
 #  [START aiplatform_sdk_create_endpoint_sample]
 def create_endpoint_sample(
-    project: str, display_name: str, location: str, sync: bool = True,
+    project: str, display_name: str, location: str,
 ):
     aiplatform.init(project=project, location=location)
 


google/cloud/aiplatform/training_utils/environment_variables.py
```diff
@@ -0,0 +1,144 @@
...
+# limitations under the License.
+#
+
+import json
+import os
+
+from typing import Dict, Optional
+
+
+class EnvironmentVariables:
```
sasha-gitg, line 24: I think it may slightly less cumbersome to make these environment variables module level attributes. Right now the usage requires instantiating `EnvironmentVariables`:

```Python
from google.cloud import training_utils

env_var = training_utils.EnvironmentVariables()
env_var.training_data_uri
```

If may be more straight forward to do this:
```Python
from google.cloud.training_utils import environment_variables

environment_variables.training_data_uri
```

After the initial import of `os`, `environ` doesn't reflect any new changes unless `environ` is changed directly: https://docs.python.org/3/library/os.html#os.environ

If we envision changing `environ` as part of the implementation in the future then we should take the class based approach but if not it will be less lines of code to go with module attributes.

What are you thoughts?

mkovalski: I've updated the variables to be module level attributes, let me know your thoughts.

PR #776, opened 2021-10-18
chore: update feature store sample code to use v1 SDK

lclc19:


samples/snippets/feature_store_service/batch_create_features_sample_test.py
```diff
@@ -15,8 +15,8 @@
 import os
 from uuid import uuid4
 
-import batch_create_features_sample
-import create_entity_type_sample
+import batch_create_features_sample_v1
+import create_entity_type_sample_v1
```
morgandu, line 19:
```
import batch_create_features_sample
import create_entity_type_sample
```

lclc19: Thanks for the comments. Instead of directly upgrading the existing snippets, we actually need to keep the current v1beta1 snippets and add another v1 version. This is to ensure that our customers using the v1beta1 API can still reference the old API docs.

morgandu (approved):

PR #805, opened 2021-10-29
feat: Add prediction container URI builder method

vinnysenthil:
Implements a prediction container URI helper as described [in this doc](https://docs.google.com/document/d/1aXvxE1DTbftFNG7q9LdvmnQB0XUL171lQIEpQNoWCik/edit?resourcekey=0-xGAv9NWWM6PDpS-dwjfYhg#) and [b/197875529](http://b/197875529)


google/cloud/aiplatform/helpers/container_uri_builders.py
```diff
@@ -0,0 +1,136 @@
...
+
+from google.cloud.aiplatform.constants import prediction
+from google.cloud.aiplatform.initializer import global_config
+
+
+def get_prediction_container_uri(
+    framework: str,
+    framework_version: str,
+    region: str = None,
+    with_accelerator: bool = False,
```
vinnysenthil, line 25: I'm unsure about using a flag here, since [with PyTorch containers](https://cloud.google.com/vertex-ai/docs/training/pre-built-containers#pytorch) - there are three flavors `xla`, `cpu`, and `gpu`. But as this seems version dependent, the method could handle searching for the correct URI.

google/cloud/aiplatform/helpers/container_uri_builders.py
```diff
@@ -0,0 +1,136 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import re
+
+from google.cloud.aiplatform.constants import prediction
+from google.cloud.aiplatform.initializer import global_config
+
+
+def get_prediction_container_uri(
```
sasha-gitg, line 21:
```suggestion
def get_prebuilt_prediction_container_uri(
```

google/cloud/aiplatform/helpers/container_uri_builders.py
```diff
@@ -0,0 +1,136 @@
...
+    )
+
+    raise ValueError(
+        f"No serving container for {framework} {framework_version} "
+        f"{'with accelerator ' if with_accelerator else ''}found. "
+        f"{ALTERNATE_VERSIONS_MESSAGE}{DOCS_URI_MESSAGE}"
+    )
+
+
+def get_training_container_uri(
```
sasha-gitg, line 130: Preference to remove this for now.

google/cloud/aiplatform/constants/prediction.py
```diff
@@ -0,0 +1,138 @@
...
+    TF2: ("gpu", "cpu"),
+    XGBOOST: (None, "cpu"),
+}
+
+SERVING_CONTAINER_URIS = (
+    SKLEARN_CONTAINER_URIS + TF_CONTAINER_URIS + XGBOOST_CONTAINER_URIS
+)
+
+# All Vertex AI Prediction first-party prediction containers as a string
+_SERVING_CONTAINER_URIS_STR = " ".join(SERVING_CONTAINER_URIS)
```
sasha-gitg, line 134: It seems like it may be simpler to build a map based on the match which will avoid having to maintain `ACCELERATOR_TO_URI_REF`, the framework constants, and this string in favor of the map.

Something like:

```Python
container_map = defaultdict(lambda: defaultdict(lambda: defaultdict(dict)))
for container_uri in SERVING_CONTAINER_URIS:
  m =  CONTAINER_URI_PATTERN.match(container_uri)
  container_map[m['region']][d['framework']][d['accelerator']][d['version']] = container_uri
```

The TF2 bookkeeping can be done here as well.

Updating with a new container uri or framework will only require adding to the list.

google/cloud/aiplatform/constants/prediction.py
```diff
@@ -0,0 +1,138 @@
...
+    "europe-docker.pkg.dev/vertex-ai/prediction/tf-cpu.1-15:latest",
+    "asia-docker.pkg.dev/vertex-ai/prediction/tf-cpu.1-15:latest",
+    "us-docker.pkg.dev/vertex-ai/prediction/tf-gpu.1-15:latest",
+    "europe-docker.pkg.dev/vertex-ai/prediction/tf-gpu.1-15:latest",
+    "asia-docker.pkg.dev/vertex-ai/prediction/tf-gpu.1-15:latest",
+]
+
+# Pattern:
+# {user_provided_framework: framework_name_in_uri}
+_FRAMEWORK_TO_URI_REF = {
```
sasha-gitg, line 109: I like the UX of this approach but it may be less ambiguous to provide a values, based on the container uri framework keyword.

google/cloud/aiplatform/helpers/container_uri_builders.py
sasha-gitg: I think we should accept `cpu`, `gpu` for now to make this more extendable if we support other accelerators in the future.

google/cloud/aiplatform/helpers/container_uri_builders.py
```diff
@@ -0,0 +1,136 @@
...
+    ```
+
+    Args:
+        framework (str):
+            The ML framework of the pre-built container. For example,
+            "tensorflow", "xgboost", or "sklearn"
+        framework_version (str):
+            The version of the specified ML framework as a string.
+        region (str):
+            A Vertex AI region or multi-region. Used to select the correct
```
sasha-gitg, line 53:
```suggestion
            Optional. A Vertex AI region or multi-region. Used to select the correct
```

google/cloud/aiplatform/helpers/container_uri_builders.py
```diff
@@ -0,0 +1,136 @@
...
+            Artifact Registry multi-region repository and reduce latency.
+            Must start with "us", "asia" or "europe". If not set, defaults
+            to location set by `aiplatform.init()`.
+        with_accelerator (bool):
+            If set to `True`, return container URI that supports GPU usage.
```
sasha-gitg, line 58:
```suggestion
            Optional. If set to `True`, return container URI that supports GPU usage.
```

google/cloud/aiplatform/helpers/container_uri_builders.py
```diff
@@ -0,0 +1,136 @@
...
+        model = aiplatform.Model.upload(
+            display_name="boston_housing_",
+            artifact_uri="gs://my-bucket/my-model/",
+            serving_container_image_uri=uri
+        )
+    ```
+
+    Args:
+        framework (str):
+            The ML framework of the pre-built container. For example,
```
sasha-gitg, line 48:
```suggestion
            Required. The ML framework of the pre-built container. For example,
```

google/cloud/aiplatform/helpers/container_uri_builders.py
```diff
@@ -0,0 +1,136 @@
...
+            "tensorflow", "xgboost", or "sklearn"
+        framework_version (str):
+            The version of the specified ML framework as a string.
```
sasha-gitg, line 51:
```suggestion
            Required. The version of the specified ML framework as a string.
```

google/cloud/aiplatform/helpers/__init__.py
```diff
@@ -1,3 +1,20 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
 from google.cloud.aiplatform.helpers import value_converter
+from google.cloud.aiplatform.helpers.container_uri_builders import (
+    get_prediction_container_uri,
+)
 
-__all__ = (value_converter,)
+__all__ = ("get_prediction_container_uri" "value_converter",)
```
sasha-gitg, line 20:
```suggestion
__all__ = ("get_prediction_container_uri", "value_converter",)
```


I'm uncertain of exposing `value_converter` as well as it's not a module we currently expose or provide documentation/usage on. It mainly used by the enhanced portions of the library.

sasha-gitg (changes requested):

google/cloud/aiplatform/helpers/__init__.py
vinnysenthil: Could we potentially move `value_converter` into `aiplatform.utils` so that `aiplatform.helpers` contains only methods intended for external use?

sasha-gitg: SGTM

google/cloud/aiplatform/helpers/container_uri_builders.py
vinnysenthil: Removed

google/cloud/aiplatform/helpers/container_uri_builders.py
```diff
@@ -0,0 +1,113 @@
...
+
+    if not URI_MAP[region]:
+        del URI_MAP[region]
+        raise ValueError(
+            f"Unsupported container region `{region}`, supported regions are "
+            f"{', '.join(URI_MAP.keys())}. "
+            f"{DOCS_URI_MESSAGE}"
+        )
+
+    if not URI_MAP[region][framework]:
```
sasha-gitg, line 89: Similarly can use `URI_MAP.get(region, {}).get(framework)` and the same pattern below.

sasha-gitg (approved):
LGTM! Thanks!

google/cloud/aiplatform/helpers/container_uri_builders.py
vinnysenthil: Using `URI_MAP[region].get(framework)` pattern since `region` is a key that is known to exist at this line.

morgandu (approved):

google/cloud/aiplatform/constants/prediction.py
```diff
@@ -0,0 +1,129 @@
...
+    "asia-docker.pkg.dev/vertex-ai/prediction/xgboost-cpu.1-1:latest",
+    "us-docker.pkg.dev/vertex-ai/prediction/xgboost-cpu.0-90:latest",
+    "europe-docker.pkg.dev/vertex-ai/prediction/xgboost-cpu.0-90:latest",
+    "asia-docker.pkg.dev/vertex-ai/prediction/xgboost-cpu.0-90:latest",
+    "us-docker.pkg.dev/vertex-ai/prediction/xgboost-cpu.0-82:latest",
+    "europe-docker.pkg.dev/vertex-ai/prediction/xgboost-cpu.0-82:latest",
+    "asia-docker.pkg.dev/vertex-ai/prediction/xgboost-cpu.0-82:latest",
+]
+
+SKLEARN_CONTAINER_URIS = [
```
abcdefgs0324, line 51: Could we add sklearn-cpu.1-0? It's already available on Vertex now.

google/cloud/aiplatform/constants/prediction.py
```diff
@@ -0,0 +1,129 @@
...
+    "asia-docker.pkg.dev/vertex-ai/prediction/sklearn-cpu.0-23:latest",
+    "us-docker.pkg.dev/vertex-ai/prediction/sklearn-cpu.0-22:latest",
+    "europe-docker.pkg.dev/vertex-ai/prediction/sklearn-cpu.0-22:latest",
+    "asia-docker.pkg.dev/vertex-ai/prediction/sklearn-cpu.0-22:latest",
+    "us-docker.pkg.dev/vertex-ai/prediction/sklearn-cpu.0-20:latest",
+    "europe-docker.pkg.dev/vertex-ai/prediction/sklearn-cpu.0-20:latest",
+    "asia-docker.pkg.dev/vertex-ai/prediction/sklearn-cpu.0-20:latest",
+]
+
+TF_CONTAINER_URIS = [
```
abcdefgs0324, line 66: Same here for tf2-cpu.2-7 and tf2-gpu.2-7.

abcdefgs0324 (changes requested):

abcdefgs0324 (approved):

leahecole (approved):
given that the rest is approved by team members, i'm approving the codeowners change

google/cloud/aiplatform/helpers/container_uri_builders.py
```diff
@@ -0,0 +1,109 @@
...
+            f"are {', '.join(URI_MAP[region][framework].keys())}. {DOCS_URI_MESSAGE}"
+        )
+
+    final_uri = URI_MAP[region][framework][accelerator].get(framework_version)
+
+    if not final_uri:
+        raise ValueError(
+            f"No serving container for `{framework}` version `{framework_version}` "
+            f"with accelerator `{accelerator}` found. Supported versions "
+            f"include {', '.join(URI_MAP[region][framework][accelerator].keys())}. {DOCS_URI_MESSAGE}"
```
Ark-kun, line 106: This code seems to raise error when the user calls the function with version which is already supported by backend, but not yet supported by the installed SDK.
I was asked to mark such case as not an error:
https://github.com/googleapis/python-aiplatform/pull/779#discussion_r737706240

PR #811, opened 2021-11-01
feat: Update constants.py

wanyixu199:
Add new regions of Vertex Tables in the SUPPORTED_REGIONS set so users can use the SDK in these 3 new regions.
While the implementation of these new regions is complete the change is required for the E2E tests of these regions.


wanyixu199:
@googlebot I signed it!

wanyixu199:
@googlebot I signed it!

sasha-gitg (changes requested):
Please add `us-west2`

sasha-gitg (approved):

PR #829, opened 2021-11-08
chore: pin `proto-plus<1.19.7`

### LINKED ISSUE #830: Working dependencies are not fully specified
`proto-plus>=1.19.7` breaks `google-api-core<2.x.x` compatibility by
requiring `protobuf>=1.19.0`. The compatibility break is detailed here:
googleapis/python-api-core#276 and so results in a broken set of
dependencies which many dependency managers cannot resolve.

judahrand:
`proto-plus>=1.19.7` breaks `google-api-core<2.x.x` compatibility by
requiring `protobuf>=1.19.0`. The compatibility break is detailed here:
https://github.com/googleapis/python-api-core/issues/276

Fixes #830 🦕

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -69,7 +69,7 @@ setuptools.setup(
         # Until this issue is closed
         # https://github.com/googleapis/google-cloud-python/issues/10566
         "google-api-core[grpc] >= 1.26.0, <3.0.0dev",
-        "proto-plus >= 1.10.1",
+        "proto-plus >= 1.10.1, <1.19.7",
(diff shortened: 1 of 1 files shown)


busunkim96:
Many thanks for noticing this and opening an issue @judahrand! I've opened https://github.com/googleapis/python-api-core/pull/305 with a more generic fix that should work for all the google-cloud-* libraries.

judahrand:
Sounds good, I'll close this then.

PR #833, opened 2021-11-08
feat: default to custom job display name if experiment name looks like a custom job ID

yfang1:
The training backend doesn't store custom job display names, so the uploader would need to fetch the job display name on its own.

This allows experiments to be more recognizable in the Cloud Console experiment list UI.

b/203591169

### DIFF
--- a/google/cloud/aiplatform/tensorboard/uploader_main.py
+++ b/google/cloud/aiplatform/tensorboard/uploader_main.py
@@ -28,8 +28,10 @@ from tensorboard.plugins.hparams import metadata as hparams_metadata
 from tensorboard.plugins.image import metadata as images_metadata
 from tensorboard.plugins.graph import metadata as graphs_metadata
 
+from google.api_core import exceptions
 from google.cloud import storage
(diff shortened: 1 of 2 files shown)


nicain:
@yfang1: I am assigning as owner of this PR; feel free to ping reviewers as needed to make sure the review process progresses in a timely fashion, or provide guidance on a who might better own the process of getting the PR reviewed, passing continuous testing, and merged. Reach out if you have questions.

sasha-gitg (commented):
LGTM. Does this have a unit test that can be updated?

This mock can be copied/reused: https://github.com/googleapis/python-aiplatform/blob/main/tests/unit/aiplatform/test_custom_job.py#L147

yfang1:
> LGTM. Does this have a unit test that can be updated?
> 
> This mock can be copied/reused: https://github.com/googleapis/python-aiplatform/blob/main/tests/unit/aiplatform/test_custom_job.py#L147

Yes, I added new unit tests for the change.

sasha-gitg (approved):

PR #849, opened 2021-11-17
feat: Upgrade Tensorboard from v1beta1 to v1

sasha-gitg:
- Added integration test for `aiplatform.Tensorboard`
- [b/206650730](b/206650730)


google/cloud/aiplatform/compat/types/__init__.py
```diff
@@ -152,6 +158,13 @@
     prediction_service_v1,
     specialist_pool_v1,
     specialist_pool_service_v1,
+    tensorboard_v1,
+    tensorboard_service_v1,
+    tensorboard_data_v1,
+    tensorboard_experiment_v1,
+    tensorboard_run_v1,
+    tensorboard_service_v1,
```
yfang1, line 166: dup'ed above

yfang1 (approved):

google/cloud/aiplatform/compat/types/__init__.py
sasha-gitg: done

tests/unit/aiplatform/test_custom_job.py
```diff
@@ -260,24 +241,25 @@ def create_custom_job_mock_with_enable_web_access():
...
     with mock.patch.object(
-        job_service_client_v1beta1.JobServiceClient, "create_custom_job"
+        job_service_client.JobServiceClient, "create_custom_job"
     ) as create_custom_job_mock:
-        create_custom_job_mock.return_value = _get_custom_job_proto(
-            name=_TEST_CUSTOM_JOB_NAME,
-            state=gca_job_state_compat.JobState.JOB_STATE_PENDING,
-            version="v1beta1",
-        )
+        create_custom_job_mock.side_effect = RuntimeError("Mock fail")
```
morgandu, line 262: What is this "Mock fail" side effect for?

morgandu (approved):
All changes look good, only have a question on Mock fail.

tests/unit/aiplatform/test_custom_job.py
sasha-gitg: This seems to be git diff surfacing the change in a confusing way.

`create_custom_job_mock_fail` hasn't changed, reference line 263 on the previous version. It's used to test that we handle failure from the service.

`create_custom_job_v1beta1_mock` was changed to `create_custom_job_mock_with_tensorboard` above.

PR #850, opened 2021-11-17
feat: add featurestore module including Featurestore, EntityType, and Feature classes; add get, update, delete, list methods in all featurestore classes; add search method in Feature class

morgandu:
feat: add featurestore module including Featurestore, EntityType, and Feature classes; add get, update, delete, list methods in all featurestore classes; add search method in Feature class


google/cloud/aiplatform/_featurestores/entity_type.py
```diff
@@ -0,0 +1,456 @@
...
+
+    _is_client_prediction_client = False
+    _resource_noun = None
+    _getter_method = "get_entity_type"
+    _list_method = "list_entity_types"
+    _delete_method = "delete_entity_type"
+
+    def __init__(
+        self,
+        entity_type_name: str,
```
diemtvu, line 50: Is this a common pattern in other resource? I personally find this a bit confuse for overload this arg with both short form and long form, which make a weird case when full path is used, but project/location/featurestore_id also provided. Is this possible to have overloading constructor in Python?

google/cloud/aiplatform/_featurestores/entity_type.py
```diff
@@ -0,0 +1,456 @@
...
+        featurestore_id: Optional[str] = None,
```
diemtvu, line 51: it's probably easier to keep the project , location, featurestore_id order

google/cloud/aiplatform/_featurestores/entity_type.py
```diff
@@ -0,0 +1,456 @@
...
+                featurestore_id='my_featurestore_id',
+            )
+
+        Args:
+            entity_type_name (str):
+                Required. A fully-qualified entityType resource name or an entity_type ID.
+                Example: "projects/123/locations/us-central1/featurestores/my_featurestore_id/entityTypes/my_entity_type_id"
+                or "my_entity_type_id" when project and location are initialized or passed, with featurestore_id passed.
+            featurestore_id (str):
+                Optional. Featurestore to retrieve entityType from.
```
diemtvu, line 76: What is the value if not set?

google/cloud/aiplatform/_featurestores/entity_type.py
```diff
@@ -0,0 +1,456 @@
...
+
+    def get_featurestore(self) -> _featurestores.Featurestore:
+        """Retrieves the managed featurestore in which this EntityType is.
+
+        Returns:
+            featurestores.Featurestore - The managed featurestore in which this EntityType is.
+        """
+        return _featurestores.Featurestore(self._featurestore_name)
+
+    def get_feature(self, feature_id: str) -> "_featurestores.Feature":
```
diemtvu, line 140: Do we expect the feature resource exist (i.e it was created before)? If not, then should the (Feature) update function throw error in that case?

Similar for get_entity_type for featurestore class.

google/cloud/aiplatform/utils/featurestore_utils.py
```diff
@@ -0,0 +1,73 @@
...
+ENTITY_TYPE_NAME_PATTERN_REGEX = (
+    FEATURESTORE_NAME_PATTERN_REGEX
+    + r"\/entityTypes\/(?P<entity_type_id>"
+    + RESOURCE_ID_PATTERN_REGEX
+    + r")"
+)
+FEATURE_NAME_PATTERN_REGEX = (
+    ENTITY_TYPE_NAME_PATTERN_REGEX
+    + r"\/features\/(?P<feature_id>"
+    + RESOURCE_ID_PATTERN_REGEX
```
lclc19, line 36: Can you add more check for feature_id -- it cannot be one of reserved words: entity_id, feature_timestamp, arrival_timestamp.

google/cloud/aiplatform/_featurestores/entity_type.py
```diff
@@ -0,0 +1,456 @@
...
+        return _featurestores.Feature.list(
+            entity_type_name=self.resource_name, filter=filter, order_by=order_by,
+        )
+
+    @base.optional_sync()
+    def delete_features(self, feature_ids: List[str], sync: bool = True,) -> None:
+        """Deletes feature resources in this EntityType given their feature IDs.
+        WARNING: This deletion is permanent.
+
+        Args:
```
lclc19, line 359: Add feature_ids description

google/cloud/aiplatform/_featurestores/entity_type.py
morgandu: One of the usage pattern for Vertex SDK is user to initialize the project, see usage example https://github.com/googleapis/python-aiplatform#initialization. And user does not need to specify `project` and `location` unless overwritten. That's why Vertex SDK normally ranks `project` and `location` to bottom among other arguments and mark as optional.

Also `Optional` arguments can not go before the `Required` arguments.

google/cloud/aiplatform/_featurestores/entity_type.py
morgandu: Python does not support overloading constructor, that's why we have this as the common pattern for all other Vertex SDK resource classes, for example this is Vertex AI SDK [Dataset class constructor](https://github.com/googleapis/python-aiplatform/blob/main/google/cloud/aiplatform/datasets/dataset.py#L61).

Agree on the confusion of the usage of optional arguments, but that's the language limitation we are working with. We try to compensate by emphasizing the Docstring and Example Usage throughout the SDK.

google/cloud/aiplatform/_featurestores/featurestore.py
```diff
@@ -0,0 +1,525 @@
...
+
+    _is_client_prediction_client = False
+    _resource_noun = "featurestores"
+    _getter_method = "get_featurestore"
+    _list_method = "list_featurestores"
+    _delete_method = "delete_featurestore"
+
+    def __init__(
+        self,
+        featurestore_name: str,
```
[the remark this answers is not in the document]
morgandu: `project` and `location` are `Optional` arguments, whereas `featurestore_name` is `Required` argument. The required arguments have to be listed before the optional arguments. There was another [discussion](https://github.com/googleapis/python-aiplatform/pull/850#discussion_r752859581) of why our `project` and `location` are marked as optional.

google/cloud/aiplatform/_featurestores/entity_type.py
morgandu: The optional `featurestore_id` defaults to `None`, see couple of lines above. 
```
featurestore_id: Optional[str] = None,
```

google/cloud/aiplatform/_featurestores/entity_type.py
morgandu: Both get_* methods will pass in the fully-qualified resource name into resource class constructor. If the resource exists, the method will return the constructed resource class, if the resource does not exist, it will actually hit the service client's get_* method and throw service error.

google/cloud/aiplatform/utils/featurestore_utils.py
morgandu: Vertex SDK is relying on the service for this type of value validation, most validation we are doing is to assist certain usage within the SDK which is to prepare the value to be passed into the service, i.e. these pattern validations essentially are to assist the limitation due to lack of overloading constructor, and we are validating the arguments and form the final resource name to pass into the service. 

However, if the service does not throw an error on the validation you proposed and will cause un alerted error, we can definitely add these in the SDK layer.

Please let me know!

google/cloud/aiplatform/_featurestores/featurestore.py
```diff
@@ -0,0 +1,525 @@
...
+            ValueError if the provided featurestore_name is not in form of a fully-qualified
+            featurestore resource name nor a featurestore ID.
+        """
+        if not (
+            featurestore_utils.validate_featurestore_name(featurestore_name)
+            or featurestore_utils.validate_id(featurestore_name)
+        ):
+            raise ValueError(
+                f"{featurestore_name} is not in form of a fully-qualified "
+                f"featurestore resource name nor a featurestore ID."
```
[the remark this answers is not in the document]
morgandu: Refactored this raise ValueError into `featurestore_utils.py` and it does not involve project and location.

google/cloud/aiplatform/_featurestores/featurestore.py
```diff
@@ -0,0 +1,525 @@
...
+            entity_type_id (str):
+                Required. The managed entityType resource ID in this Featurestore.
+        Returns:
+            featurestores.EntityType - The managed entityType resource object.
+
+        Raises:
+            ValueError if the provided entity_type_id is not in form of a entity_type ID.
+        """
+        if not featurestore_utils.validate_id(entity_type_id):
+            raise ValueError(f"{entity_type_id} is not in form of a entity_type ID.")
```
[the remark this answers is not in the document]
morgandu: Done!

google/cloud/aiplatform/_featurestores/entity_type.py
morgandu: Good catch! Thanks!

google/cloud/aiplatform/utils/featurestore_utils.py
```diff
@@ -0,0 +1,188 @@
...
+    + r")"
+)
+
+
+def validate_id(resource_id: str) -> bool:
+    """Validates feature store resource ID pattern."""
+    return bool(re.compile(r"^" + RESOURCE_ID_PATTERN_REGEX + r"$").match(resource_id))
+
+
+def validate_featurestore_name(featurestore_name: str) -> Dict[str, str]:
```
sasha-gitg, line 46: Preference to name these `parse` instead of `validate`.

Also, is additional RESOURCE_ID validation the main reason we are not relying on the GAPIC parse methods[1]?

1. https://github.com/googleapis/python-aiplatform/blob/main/google/cloud/aiplatform_v1/services/featurestore_service/client.py#L230

google/cloud/aiplatform/utils/featurestore_utils.py
```diff
@@ -0,0 +1,188 @@
...
+    """Gets composite resource noun for entity_type resource."""
+    return f"featurestores/{featurestore_id}/entityTypes"
+
+
+def get_feature_resource_noun(featurestore_id: str, entity_type_id: str) -> str:
+    """Gets composite resource noun for feature resource."""
+    return f"featurestores/{featurestore_id}/entityTypes/{entity_type_id}/features"
+
+
+def validate_and_get_featurestore_resource_id(featurestore_name: str,):
```
sasha-gitg, line 76:
```suggestion
def validate_and_get_featurestore_resource_id(featurestore_name: str,) -> str:
```

google/cloud/aiplatform/utils/featurestore_utils.py
```diff
@@ -0,0 +1,188 @@
...
+        raise ValueError(
+            f"{featurestore_name} is not in form of a fully-qualified featurestore resource name nor an featurestore ID."
+        )
+
+    return featurestore_id
+
+
+def validate_and_get_entity_type_resource_ids(
+    entity_type_name: str, featurestore_id: Optional[str] = None,
+):
```
sasha-gitg, line 108:
```suggestion
) -> Tuple[str, str]:
```

google/cloud/aiplatform/utils/featurestore_utils.py
```diff
@@ -0,0 +1,188 @@
...
+            f"nor an entity_type ID with featurestore_id passed."
+        )
+    return (featurestore_id, entity_type_id)
+
+
+def validate_and_get_feature_resource_ids(
+    feature_name: str,
+    featurestore_id: Optional[str] = None,
+    entity_type_id: Optional[str] = None,
+):
```
sasha-gitg, line 149:
```suggestion
) -> Tuple[str, str, str]:
```

sasha-gitg (changes requested):

google/cloud/aiplatform/utils/__init__.py
```diff
@@ -453,6 +461,23 @@ class EndpointClientWithOverride(ClientWithOverride):
...
 
 
+class FeaturestoreClientWithOverride(ClientWithOverride):
+    _is_temporary = True
+    _default_version = compat.DEFAULT_VERSION
+    _version_map = (
+        (compat.V1, featurestore_service_client_v1.FeaturestoreServiceClient),
+        (compat.V1BETA1, featurestore_service_client_v1beta1.FeaturestoreServiceClient),
+        (
+            "online_" + compat.V1,
```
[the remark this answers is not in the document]
morgandu: added an `FeaturestoreOnlineServingClientWithOverride` matching the current pattern

google/cloud/aiplatform/_featurestores/entity_type.py
```diff
@@ -0,0 +1,436 @@
...
+                set in aiplatform.init will be used.
+            credentials (auth_credentials.Credentials):
+                Optional. Custom credentials to use to list entityTypes. Overrides
+                credentials set in aiplatform.init.
+
+        Returns:
+            List[EntityTypes] - A list of managed entityType resource objects
+        """
+
+        cls._featurestore_id = featurestore_utils.validate_and_get_featurestore_resource_id(
```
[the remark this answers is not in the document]
morgandu: I have already changed the base `_list()` method to include an optional `parent` field for `search_feature`. Updated couple of unnecessary `cls._* =*` which is causing the confusion on mutating the class, those are only to parse some ids and parent resource name for retrieving child resources. Let me know if this resolve your concern. 

However, one thing I would need more guidance is the forming of `_resource_noun` for the nested resource, which require the ancestor IDs as part of the _resource_noun, which can not be pre defined.

google/cloud/aiplatform/_featurestores/featurestore.py
```diff
@@ -0,0 +1,525 @@
...
+        )
+
+    @base.optional_sync()
+    def delete_entity_types(
+        self, entity_type_ids: List[str], sync: bool = True,
+    ) -> None:
+        """Deletes entity_type resources in this Featurestre given their entity_type IDs.
+        WARNING: This deletion is permanent.
+
+        Args:
```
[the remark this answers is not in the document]
morgandu: Done

google/cloud/aiplatform/_featurestores/featurestore.py
```diff
@@ -0,0 +1,527 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+from typing import Dict, List, Optional, Sequence, Tuple, Union
+
+import pandas as pd
```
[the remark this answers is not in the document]
morgandu: We can replace the 
```import pandas as pd``` 
with
```
    import pandas as pd
except ImportError:
    raise ImportError(...)
 ```
But what we normally do with the type hints of the methods? Placing the import inside the method will require a different type hint?

google/cloud/aiplatform/_featurestores/featurestore.py
```diff
@@ -0,0 +1,527 @@
...
+        Args:
+            entity_type_id (str):
+                Required. The managed entityType resource ID in this Featurestore.
+        Returns:
+            featurestores.EntityType - The managed entityType resource object.
+
+        Raises:
+            ValueError if the provided entity_type_id is not in form of an entity_type ID.
+        """
+        if not featurestore_utils.validate_id(entity_type_id):
```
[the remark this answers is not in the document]
morgandu: The `featurestore.get_entity_type` requires an `entity_type_id` only that's why there is an explicit `featurestore_utils.validate_id(entity_type_id):` in use. For the `EntityType` constructor, since it requires an `entity_type_name` and an optional `featurestore_id`, the validation is inside the `featurestore_utils.validate_and_get_entity_type_resource_ids`.

google/cloud/aiplatform/utils/featurestore_utils.py
morgandu: Yes, I am still evaluating the restrictions on Feature Store's resource ids, since we are enabling resource creation/ingestion through user provided df. My gut feeling is we may need some validation in the SDK layer but I think I will have clearer idea when implementing the df related methods. 

I will use the GAPIC methods for now and revisit later if they are needed.

sasha-gitg (commented):
Please add a more informative PR title because it will be added to the release notes.

morgandu:
- Addressed feedback to remove parent resource as property, instead utilize `gca_resource` i.e. `self.resource_name` when needed. 
- Cleaned up unit tests according to the changes. 
- Added integration test, though in this PR we can only cover `list()` and `search_features()` methods in `Featurestore` class.

google/cloud/aiplatform/_featurestores/entity_type.py
diemtvu: I mean if it's None, does it mean the value from the entity_type_name name will be used, which implies entity_type_name must be the fully qualified path. And when it's None, then entity_type_name must be the short ID form. Probably adding some explanation to be clear

diemtvu (approved):

google/cloud/aiplatform/_featurestores/feature.py
```diff
@@ -0,0 +1,306 @@
...
+            featurestore_id,
+            entity_type_id,
+            _,
+        ) = featurestore_utils.validate_and_get_feature_resource_ids(
+            feature_name=feature_name,
+            entity_type_id=entity_type_id,
+            featurestore_id=featurestore_id,
+        )
+
+        self._resource_noun = (
```
sasha-gitg, line 96: Add a comment this this is a workaround until the base class supports nested features.

google/cloud/aiplatform/_featurestores/feature.py
```diff
@@ -0,0 +1,306 @@
...
+            location=location,
+            credentials=credentials,
+            resource_name=feature_name,
+        )
+        self._gca_resource = self._get_gca_resource(resource_name=feature_name)
+
+    @property
+    def featurestore_name(self) -> str:
+        """Full qualified resource name of the managed featurestore in which this Feature is."""
+        return "/".join(self.resource_name.split("/")[:-4])
```
sasha-gitg, line 111: Preference to use GAPIC's parsers here

google/cloud/aiplatform/_featurestores/feature.py
```diff
@@ -0,0 +1,306 @@
...
+
+        Returns:
+            featurestores.Featurestore - The managed featurestore in which this Feature is.
+        """
+        return _featurestores.Featurestore(featurestore_name=self.featurestore_name)
+
+    @property
+    def entity_type_name(self) -> str:
+        """Full qualified resource name of the managed entityType in which this Feature is."""
+        return "/".join(self.resource_name.split("/")[:-2])
```
sasha-gitg, line 124: Preference to use GAPIC's parsers here.

google/cloud/aiplatform/_featurestores/featurestore.py
```diff
@@ -0,0 +1,457 @@
...
+                Required. The list of entity_type IDs to be deleted.
+            sync (bool):
+                Optional. Whether to execute this deletion synchronously. If False, this method
+                will be executed in concurrent Future and any downstream object will
+                be immediately returned and synced when the Future has completed.
+        """
+        for entity_type_id in entity_type_ids:
+            entity_type = self.get_entity_type(entity_type_id=entity_type_id)
+            entity_type.delete(sync=sync)
+            if not sync:
```
sasha-gitg, line 315: Same comment here on concurrent deletions.

google/cloud/aiplatform/_featurestores/entity_type.py
```diff
@@ -0,0 +1,368 @@
...
+                Required. The list of feature IDs to be deleted.
+            sync (bool):
+                Optional. Whether to execute this deletion synchronously. If False, this method
+                will be executed in concurrent Future and any downstream object will
+                be immediately returned and synced when the Future has completed.
+        """
+        for feature_id in feature_ids:
+            feature = self.get_feature(feature_id=feature_id)
+            feature.delete(sync=sync)
+            if not sync:
```
sasha-gitg, line 367: To ensure deletions are as concurrent as possible and reduce the time to delete all features, `wait` should be called after `delete` is invoked on all features. Currently features are deleted sequentially.

google/cloud/aiplatform/_featurestores/featurestore.py
```diff
@@ -0,0 +1,457 @@
...
+
+        Returns:
+            List[Features] - A list of managed feature resource objects
+        """
+        resource = cls._empty_constructor(
+            project=project, location=location, credentials=credentials
+        )
+
+        # Fetch credentials once and re-use for all `_empty_constructor()` calls
+        creds = initializer.global_config.credentials
```
sasha-gitg, line 433: This should default to the credentials passed in then the `global_config` credentials. This is handled by the construction of the resource.

```suggestion
        creds = resource.credentials
```

google/cloud/aiplatform/_featurestores/featurestore.py
```diff
@@ -0,0 +1,457 @@
...
+
+        if page_size:
+            search_features_request["page_size"] = page_size
+
+        feature_list = (
+            resource.api_client.search_features(request=search_features_request) or []
+        )
+
+        return [
+            _featurestores.Feature(
```
sasha-gitg, line 450: Similar to `_list`, this should use`cls._construct_sdk_resource_from_gapic` to avoid the additional get call on construction.

google/cloud/aiplatform/utils/featurestore_utils.py
```diff
@@ -0,0 +1,149 @@
...
+
+RESOURCE_ID_PATTERN_REGEX = r"[a-z_][a-z0-9_]{0,59}"
+
+
+def validate_id(resource_id: str) -> bool:
+    """Validates feature store resource ID pattern."""
+    return bool(re.compile(r"^" + RESOURCE_ID_PATTERN_REGEX + r"$").match(resource_id))
+
+
+def validate_and_get_featurestore_resource_id(featurestore_name: str) -> str:
```
sasha-gitg, line 33: This seems to only be used in the `FeatureStore` constructor `EntityType.list`. Both also rely on `utils.full_resource_name` which does similar validation. What is the benefit of using this approach? Is OK to just rely on `utils.full_resource_name`.

sasha-gitg (changes requested):

google/cloud/aiplatform/_featurestores/featurestore.py
morgandu: Using `cls._construct_sdk_resource_from_gapic` in `Featurestore.search_features()` will return `Featurestore` class. Move `Featurestore.search_features()` to `Feature.search()` and utilize the class methods to retrieve searched feature list.

google/cloud/aiplatform/_featurestores/featurestore.py
```diff
@@ -0,0 +1,322 @@
...
+        Returns:
+            List[EntityTypes] - A list of managed entityType resource objects.
+        """
+        return _featurestores.EntityType.list(
+            featurestore_name=self.resource_name, filter=filter, order_by=order_by,
+        )
+
+    @base.optional_sync()
+    def delete_entity_types(
+        self, entity_type_ids: List[str], sync: Optional[bool] = True,
```
sasha-gitg, line 304:
```suggestion
        self, entity_type_ids: List[str], sync: bool = True,
```

google/cloud/aiplatform/_featurestores/featurestore.py
```diff
@@ -0,0 +1,322 @@
...
+    ) -> None:
+        """Deletes entity_type resources in this Featurestre given their entity_type IDs.
```
sasha-gitg, line 306:
```suggestion
        """Deletes entity_type resources in this Featurestore given their entity_type IDs.
```

google/cloud/aiplatform/_featurestores/featurestore.py
```diff
@@ -0,0 +1,322 @@
...
+                -  ``update_time``
+
+        Returns:
+            List[EntityTypes] - A list of managed entityType resource objects.
+        """
+        return _featurestores.EntityType.list(
+            featurestore_name=self.resource_name, filter=filter, order_by=order_by,
+        )
+
+    @base.optional_sync()
```
sasha-gitg, line 302: Just realized that this is decorated. I think this is new usage pattern for the SDK and there are three options on usage here to consider. This is worth discussion because this pattern should be used throughout.

**1**:  if this method is decorated then the usage is:

```Python
 feature_store.delete_entity_types(..., sync=False)
 feature_store.wait()  # will throw if any of the deletions had issues
 ```

In this case there is no need to pass `sync` to the underlying deletions and instead we can default to asynchronous to make deletions concurrent. Also note that the method needs to be changed to wait on all entity_type deletes not just the last one as it is currently handled.

```Python 
entity_types = []
for entity_type_id in entity_type_ids:
  entity_type = self.get_entity_type(entity_type_id=entity_type_id)
  entity_type.delete(sync=False)
  entity_types.append(entity_type)

for entity_type in entity_types:
  entity_type.wait()
```

**2**: other option is to pass `sync` directly to the underlying delete calls but not associate them with the FeatureStore:

```Python
 feature_store.delete_entity_types(..., sync=False)
 feature_store.wait()  # will NOT throw if any of the deletions had issues
```

In this case this method should not be decorated with `base.optional_sync`. There is no need to call wait here and the implementation would look like:

```Python 
for entity_type_id in entity_type_ids:
  entity_type = self.get_entity_type(entity_type_id=entity_type_id)
  entity_type.delete(sync=sync)
``` 

**3**: is similar to the second option but instead we can return the list of deleted entity types so the user can wait on each one. So usage look like:

```Python
deleted_entity_types: List[EntityType] = feature_store.delete_entity_types(..., sync=False)
feature_store.wait()  # will NOT throw if any of the deletions had issues

for entity_type in deleted_entity_types:
  entity_type.wait()
```

The implementation would look similar to (2) but we compose the list and return it.

```Python 
entity_types = []
for entity_type_id in entity_type_ids:
  entity_type = self.get_entity_type(entity_type_id=entity_type_id)
  entity_type.delete(sync=sync)
  entity_types.append(entity_type)
return entity_types
``` 

Also to unblock this PR, feel free to remove the async support for now and follow up. Up to you.

morgandu: Thanks for the options. I would vote for option 1 over 2 and 3. The main reason is when we delete child resources from the parent resource, I do want to provide sync option to prevent user to retrieve the delete pending child from the parent, i.e. Featurestore.get_entity_type() when Featurestore.delete_entity_types(). So add a Featurestore.wait() during deleting kind of make sense, and if any of the deletion had issue, it is captured.

sasha-gitg: SGTM

sasha-gitg (approved):
LGTM! Thanks!

PR #851, opened 2021-11-18
feat: updating Tensorboard related code to use v1

yfang1:
Fallout of #849

### DIFF
--- a/google/cloud/aiplatform/tensorboard/plugins/tf_profiler/profile_uploader.py
+++ b/google/cloud/aiplatform/tensorboard/plugins/tf_profiler/profile_uploader.py
@@ -38,20 +38,14 @@ from tensorboard.util import tb_logging
 import tensorflow as tf
 
 from google.cloud import storage
-from google.cloud.aiplatform.compat.services import tensorboard_service_client_v1beta1
-from google.cloud.aiplatform.compat.types import (
-    tensorboard_data_v1beta1 as tensorboard_data,
-)
-from google.cloud.aiplatform.compat.types import (
-    tensorboard_service_v1beta1 as tensorboard_service,
-)
-from google.cloud.aiplatform.compat.types import (
-    tensorboard_time_series_v1beta1 as tensorboard_time_series,
-)
+from google.cloud.aiplatform.compat.services import tensorboard_service_client
+from google.cloud.aiplatform.compat.types import tensorboard_data
+from google.cloud.aiplatform.compat.types import tensorboard_service
+from google.cloud.aiplatform.compat.types import tensorboard_time_series
 from google.cloud.aiplatform.tensorboard import uploader_utils
 from google.protobuf import timestamp_pb2 as timestamp
 
-TensorboardServiceClient = tensorboard_service_client_v1beta1.TensorboardServiceClient
+TensorboardServiceClient = tensorboard_service_client.TensorboardServiceClient
 
 logger = tb_logging.get_logger()
 
--- a/google/cloud/aiplatform/tensorboard/uploader_utils.py
+++ b/google/cloud/aiplatform/tensorboard/uploader_utils.py
@@ -29,18 +29,12 @@ from tensorboard.util import tb_logging
 
 from google.api_core import exceptions
 from google.cloud import storage
-from google.cloud.aiplatform.compat.types import (
-    tensorboard_run_v1beta1 as tensorboard_run,
-)
-from google.cloud.aiplatform.compat.types import (
-    tensorboard_service_v1beta1 as tensorboard_service,
-)
-from google.cloud.aiplatform.compat.types import (
-    tensorboard_time_series_v1beta1 as tensorboard_time_series,
-)
-from google.cloud.aiplatform.compat.services import tensorboard_service_client_v1beta1
+from google.cloud.aiplatform.compat.types import tensorboard_run
+from google.cloud.aiplatform.compat.types import tensorboard_service
+from google.cloud.aiplatform.compat.types import tensorboard_time_series
+from google.cloud.aiplatform.compat.services import tensorboard_service_client
 
-TensorboardServiceClient = tensorboard_service_client_v1beta1.TensorboardServiceClient
+TensorboardServiceClient = tensorboard_service_client.TensorboardServiceClient
 
 logger = tb_logging.get_logger()
 logger.setLevel(logging.WARNING)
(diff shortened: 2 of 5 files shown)


google/cloud/aiplatform/tensorboard/plugins/tf_profiler/profile_uploader.py
```diff
@@ -38,20 +38,18 @@
 import tensorflow as tf
 
 from google.cloud import storage
-from google.cloud.aiplatform.compat.services import tensorboard_service_client_v1beta1
+from google.cloud.aiplatform.compat.services import tensorboard_service_client_v1
+from google.cloud.aiplatform.compat.types import tensorboard_data_v1 as tensorboard_data
```
morgandu, line 42: We append the `*_v1beta1` i.e. `tensorboard_service_client_v1beta1` only when the `v1beta1` version is in use which is against compat's `DEFAULT_VERSION = V1`. 

For all `compat` supporting `types` and `services`, you can go with i.e 
`from google.cloud.aiplatform.compat.types import tensorboard_data` where `types.tensorboard_data = types.tensorboard_data_v1` see config here: https://github.com/googleapis/python-aiplatform/blob/main/google/cloud/aiplatform/compat/__init__.py#L24

The lines below should be sufficient for calling v1 as the default version:
```
from google.cloud.aiplatform.compat.services import tensorboard_service_client
from google.cloud.aiplatform.compat.types import tensorboard_data
```

morgandu (changes requested):

morgandu:
Left a comment which applies to all the changes. So the rule of thumb is: 

if your expected version aligns with the `DEFAULT_VERSION` in the `compact` module, no need to specify version i.e. `from google.cloud.aiplatform.compat.types import tensorboard_data`; but if your expected version is different than the `DEFAULT_VERSION`, import the particular version. i.e. `from google.cloud.aiplatform.compat.types import tensorboard_data_v1beta1`.

google/cloud/aiplatform/tensorboard/plugins/tf_profiler/profile_uploader.py
yfang1: ok changed. Thanks!

google/cloud/aiplatform/tensorboard/uploader_utils.py
```diff
@@ -201,7 +195,7 @@ def _create_or_get_run_resource(
                 Required. The display name of this run.
 
         Returns:
-            tb_run (google.cloud.aiplatform_v1beta1.types.TensorboardRun):
+            tb_run (google.cloud.aiplatform_v1.types.TensorboardRun):
```
morgandu, line 198: ```tb_run (tensorboard_run.TensorboardRun):```

tests/unit/aiplatform/test_uploader.py
```diff
@@ -46,22 +46,18 @@
 from google.cloud.aiplatform.tensorboard.plugins.tf_profiler import profile_uploader
 import google.cloud.aiplatform.tensorboard.uploader as uploader_lib
 from google.cloud import storage
-from google.cloud.aiplatform.compat.services import tensorboard_service_client_v1beta1
-from google.cloud.aiplatform_v1beta1.services.tensorboard_service.transports import (
+from google.cloud.aiplatform.compat.services import tensorboard_service_client
```
morgandu, line 49: ```from google.cloud.aiplatform_v1.services import tensorboard_service_client```

tests/unit/aiplatform/test_uploader.py
```diff
@@ -46,22 +46,18 @@
...
+from google.cloud.aiplatform_v1.services.tensorboard_service.transports import (
     grpc as transports_grpc,
 )
+from google.cloud.aiplatform.compat.types import tensorboard_data
+from google.cloud.aiplatform.compat.types import tensorboard_service
```
morgandu, line 54:
```
from google.cloud.aiplatform_v1.types import tensorboard_data
from google.cloud.aiplatform_v1.types import tensorboard_service
```

tests/unit/aiplatform/test_uploader.py
```diff
@@ -46,22 +46,18 @@
...
+from google.cloud.aiplatform.compat.types import tensorboard_run as tensorboard_run_type
 from google.cloud.aiplatform.compat.types import (
-    tensorboard_experiment_v1beta1 as tensorboard_experiment_type,
-)
-from google.cloud.aiplatform.compat.types import (
-    tensorboard_run_v1beta1 as tensorboard_run_type,
-)
-from google.cloud.aiplatform.compat.types import (
-    tensorboard_time_series_v1beta1 as tensorboard_time_series_type,
+    tensorboard_time_series as tensorboard_time_series_type,
```
morgandu, line 60: dito

tests/unit/aiplatform/test_uploader.py
```diff
@@ -46,22 +46,18 @@
...
     grpc as transports_grpc,
 )
+from google.cloud.aiplatform.compat.types import tensorboard_data
+from google.cloud.aiplatform.compat.types import tensorboard_service
 from google.cloud.aiplatform.compat.types import (
-    tensorboard_data_v1beta1 as tensorboard_data,
-    tensorboard_service_v1beta1 as tensorboard_service,
+    tensorboard_experiment as tensorboard_experiment_type,
 )
+from google.cloud.aiplatform.compat.types import tensorboard_run as tensorboard_run_type
```
morgandu, line 58: dito

tests/unit/aiplatform/test_uploader.py
```diff
@@ -46,22 +46,18 @@
...
+from google.cloud.aiplatform.compat.services import tensorboard_service_client
+from google.cloud.aiplatform_v1.services.tensorboard_service.transports import (
     grpc as transports_grpc,
 )
+from google.cloud.aiplatform.compat.types import tensorboard_data
+from google.cloud.aiplatform.compat.types import tensorboard_service
 from google.cloud.aiplatform.compat.types import (
-    tensorboard_data_v1beta1 as tensorboard_data,
-    tensorboard_service_v1beta1 as tensorboard_service,
+    tensorboard_experiment as tensorboard_experiment_type,
```
morgandu, line 56: dito

morgandu (approved):
Sorry I probably missed the `test_uploader.py` in the previous review. For unit tests, we don't rely on the `compat` module, instead we specify the expected version. Because we don't have tests when `compat` may or may not select different version within different resource class methods, this way we can ensure class methods calling `compat` returning the right version we expect.

We only discussed this when `compat` was introduced, but did not enforce the implementation across all other module, so there is some inconsistency in the code base. But prefer start to enforce it for the reason mentioned above.

Issue #854, opened 2021-11-19
Error when working with dataset when BigQuery source URI contains `:`


Ark-kun:
I receive error when I try working with dataset that has `:` in the BigQuery source URI:
`BadRequest: 400 GET https://bigquery.googleapis.com/bigquery/v2/projects/140626129697/datasets/bigquery-public-data:chicago_taxi_trips/tables/taxi_trips?prettyPrint=false: Invalid dataset ID "bigquery-public-data:chicago_taxi_trips". Dataset IDs must be alphanumeric (plus underscores and dashes) and must be at most 1024 characters long.`

The Vertex Backend supports the dataset and the UX shows columns and types.

#### Environment details

  - OS type and version:
  - `google-cloud-aiplatform` version: `1.4.0`

#### Steps to reproduce
```python
aiplatform.TabularDataset('projects/140626129697/locations/us-central1/datasets/4730389293762084864').column_names
```

#### Stack trace
```
---------------------------------------------------------------------------
BadRequest                                Traceback (most recent call last)
<ipython-input-153-50eba4b09d21> in <module>
----> 1 bug_dataset.column_names

/opt/conda/lib/python3.7/site-packages/google/cloud/aiplatform/datasets/tabular_dataset.py in column_names(self)
     85                 No more than 64 user labels can be associated with one Tensorboard
     86                 (System labels are excluded).
---> 87                 See https://goo.gl/xmQnxf for more information and examples of labels.
     88                 System reserved label keys are prefixed with "aiplatform.googleapis.com/"
     89                 and are immutable.

/opt/conda/lib/python3.7/site-packages/google/cloud/aiplatform/datasets/tabular_dataset.py in _retrieve_bq_source_columns(project, bq_table_uri, credentials)

/opt/conda/lib/python3.7/site-packages/google/cloud/bigquery/client.py in get_table(self, table, retry, timeout)
    695         table_ref = _table_arg_to_table_ref(table, default_project=self.project)
    696         api_response = self._call_api(
--> 697             retry, method="GET", path=table_ref.path, timeout=timeout
    698         )
    699         return Table.from_api_repr(api_response)

/opt/conda/lib/python3.7/site-packages/google/cloud/bigquery/client.py in _call_api(self, retry, **kwargs)
    572         if retry:
    573             call = retry(call)
--> 574         return call()
    575 
    576     def get_dataset(self, dataset_ref, retry=DEFAULT_RETRY, timeout=None):

[...]

PR #855, opened 2021-11-19
fix: Fixed BigQuery datasets that have colon in URI

Ark-kun:
The colon-based "project:dataset.table" format is no longer supported.

Fixes: https://github.com/googleapis/python-aiplatform/issues/854


ivanmkc (approved):
LGTM

google/cloud/aiplatform/datasets/column_names_dataset.py
```diff
@@ -237,6 +237,12 @@ def _retrieve_bq_source_columns(
         if bq_table_uri.startswith(prefix):
             bq_table_uri = bq_table_uri[len(prefix) :]
 
+        # The colon-based "project:dataset.table" format is no longer supported:
+        # Invalid dataset ID "bigquery-public-data:chicago_taxi_trips".
+        # Dataset IDs must be alphanumeric (plus underscores and dashes) and must be at most 1024 characters long.
+        # Using dot-based "project.dataset.table" format instead.
+        bq_table_uri.replace(":", ".")
```
imrehg, line 244: `.replace()` returns the modified copy of the string, rather than applying changes in place, right? This line has no actual effect on the value of `bq_table_uri`, if I read things correctly.

This should be the correct form?
```suggestion
        bq_table_uri = bq_table_uri.replace(":", ".")
```

Ark-kun: Thank you!
I'm don't know how I overlooked this bug.

PR #869, opened 2021-11-29
fix: Import error for cloud_profiler

### LINKED ISSUE #867: The build failed
This test failed!

To configure my behavior, see [the Flaky Bot documentation](https://github.com/googleapis/repo-automation-bots/tree/main/packages/flakybot).

If I'm commenting on this issue too often, add the `flakybot: quiet` label and
I will stop commenting.

---

commit: 6d5c7c42d1c352f161e4738c6dbbf540a032017b
buildURL: [Build Status](https://source.cloud.google.com/results/invocations/3cb6b038-12b5-44e7-a09b-f61ed5b514ea), [Sponge](http://sponge2/3cb6b038-12b5-44e7-a09b-f61ed5b514ea)
status: failed

mkovalski:
Circular imports involving absolute imports with binding a submodule to a name was not supported in python 3.6, as documented [here](https://docs.python.org/3/whatsnew/3.7.html#other-language-changes). Needed to update import.

- Import initializer without alias 
- Move the ImportError to the tf profiler module.

Fixes #867


google/cloud/aiplatform/training_utils/cloud_profiler/__init__.py
```diff
@@ -15,13 +15,7 @@
 # limitations under the License.
 #
 
-try:
```
sasha-gitg, line 18: It seems like removing this may throw if werkzeug is not installed but it won't raise the same informative message.

Also as a follow up, should werkzeung be added to setup.py?

sasha-gitg (changes requested):

google/cloud/aiplatform/training_utils/cloud_profiler/__init__.py
mkovalski: Added werkzeug within the try/except and updated setup.py

google/cloud/aiplatform/training_utils/cloud_profiler/plugins/tensorflow/tf_profiler.py
```diff
@@ -17,6 +17,15 @@
 
 """A plugin to handle remote tensoflow profiler sessions for Vertex AI."""
 
+try:
+    from tensorboard_plugin_profile.profile_plugin import ProfilePlugin
+    from werkzeug import Response
```
sasha-gitg, line 22: `initializer.py` imports `werkzeug` before importing `tf_profiler`: https://github.com/googleapis/python-aiplatform/blob/main/google/cloud/aiplatform/training_utils/cloud_profiler/initializer.py#L21 

Recommend catching on all optional imports or walking through all the imports to determine when they are first imported.

tests/unit/aiplatform/test_cloud_profiler.py
```diff
@@ -130,6 +130,12 @@ class TestProfilerPlugin(unittest.TestCase):
     def setUp(self):
         setupProfilerEnvVars()
 
+    def testImportError(self):
+        for mock_module in ["tensorboard_plugin_profile.profile_plugin", "werkzeug"]:
+            with self.subTest():
+                with mock.patch.dict("sys.modules", {mock_module: None}):
+                    self.assertRaises(ImportError, reload, tf_profiler)
```
sasha-gitg, line 137: Should this test importing `cloud_profiler` as that's the prescribed usage?

Also, may want to use the [context manager](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaises) to get the exception object and ensure the raised exception includes the intended verbiage.

sasha-gitg (changes requested):

setup.py
```diff
@@ -36,7 +36,11 @@
 tensorboard_extra_require = ["tensorflow >=2.3.0, <=2.5.0"]
 metadata_extra_require = ["pandas >= 1.0.0"]
 xai_extra_require = ["tensorflow >=2.3.0, <=2.5.0"]
-profiler_extra_require = ["tensorboard-plugin-profile", "tensorflow >=2.4.0"]
+profiler_extra_require = [
+    "tensorboard-plugin-profile",
+    "werkzeug",
```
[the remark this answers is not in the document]
mkovalski: Done.

google/cloud/aiplatform/training_utils/cloud_profiler/plugins/tensorflow/tf_profiler.py
mkovalski: Done, I went with catching the imports at where they were first imported. The tests should catch the verbiage as well.

tests/unit/aiplatform/test_cloud_profiler.py
mkovalski: Done.

sasha-gitg (approved):

PR #872, opened 2021-12-01
feat: add create in Featurestore, EntityType, Feature; add create_entity_type in Featurestore; add create_feature, batch_create_features in EntityType; add ingest_from_* for bq and gcs in EntityType; add and update delete with force delete nested resources

morgandu:
- [x] add create method in Featurestore, EntityType, Feature classes; 
- [x] add create_entity_type in Featurestore; 
- [x] add create_feature, batch_create_features in EntityType; 
- [x] add ingest_from_bq and ingest_from_gcs methods in EntityType class
- [x] add and update delete methods with force delete nested resources
- [x] add unit tests
- [x] update integration test

### DIFF
--- a/google/cloud/aiplatform/featurestore/entity_type.py
+++ b/google/cloud/aiplatform/featurestore/entity_type.py
@@ -484,11 +484,11 @@ class EntityType(base.VertexAiResourceNounWithFutureManager):
                 System reserved label keys are prefixed with
                 "aiplatform.googleapis.com/" and are immutable.
             project (str):
-                Optional. Project to create EntityType in. If not set, project
-                set in aiplatform.init will be used.
+                Optional. Project to create EntityType in if `featurestore_name` is passed an featurestore ID.
+                If not set, project set in aiplatform.init will be used.
             location (str):
-                Optional. Location to create EntityType in. If not set, location
-                set in aiplatform.init will be used.
+                Optional. Location to create EntityType in if `featurestore_name` is passed an featurestore ID.
+                If not set, location set in aiplatform.init will be used.
             credentials (auth_credentials.Credentials):
                 Optional. Custom credentials to use to create EntityTypes. Overrides
                 credentials set in aiplatform.init.
@@ -511,14 +511,23 @@ class EntityType(base.VertexAiResourceNounWithFutureManager):
             location=location,
         )
 
+        gapic_entity_type = gca_entity_type.EntityType()
+
         if labels:
             utils.validate_labels(labels)
+            gapic_entity_type.labels = labels
 
-        gapic_entity_type = gca_entity_type.EntityType(
-            description=description, labels=labels,
+        if description:
+            gapic_entity_type.description = description
+
+        api_client = cls._instantiate_client(
+            location=featurestore_utils.CompatFeaturestoreServiceClient.parse_featurestore_path(
+                path=featurestore_name
+            )[
+                "location"
+            ],
+            credentials=credentials,
         )
-
-        api_client = cls._instantiate_client(location=location, credentials=credentials)
 
         created_entity_type_lro = api_client.create_entity_type(
             parent=featurestore_name,
@@ -741,12 +750,9 @@ class EntityType(base.VertexAiResourceNounWithFutureManager):
     def _validate_and_get_import_feature_values_request(
         self,
         feature_ids: List[str],
+        feature_time: Union[str, datetime.datetime],
+        data_source: Union[gca_io.AvroSource, gca_io.BigQuerySource, gca_io.CsvSource],
         feature_source_fields: Optional[Dict[str, str]] = None,
-        avro_source: Optional[gca_io.AvroSource] = None,
-        bigquery_source: Optional[gca_io.BigQuerySource] = None,
-        csv_source: Optional[gca_io.CsvSource] = None,
-        feature_time_field: Optional[str] = None,
-        feature_time: Optional[datetime.datetime] = None,
         entity_id_field: Optional[str] = None,
         disable_online_serving: Optional[bool] = None,
         worker_count: Optional[int] = None,
@@ -757,6 +763,18 @@ class EntityType(base.VertexAiResourceNounWithFutureManager):
                 Required. IDs of the Feature to import values
                 of. The Features must exist in the target
                 EntityType, or the request will fail.
+            feature_time (Union[str, datetime.datetime]):
+                Required. The feature_time can be one of:
+                    - The source column that holds the Feature
+                    timestamp for all Feature values in each entity.
+                    - A single Feature timestamp for all entities
+                    being imported. The timestamp must not have
+                    higher than millisecond precision.
+            data_source (Union[gca_io.AvroSource, gca_io.BiqQuerySource, gca_io.CsvSource]):
+                Required. The data_source can be one of:
+                    - AvroSource
+                    - BiqQuerySource
+                    - CsvSource
             feature_source_fields (Dict[str, str]):
                 Optional. User defined dictionary to map ID of the Feature for importing values
                 of to the source column for getting the Feature values from.
@@ -782,24 +800,6 @@ class EntityType(base.VertexAiResourceNounWithFutureManager):
                      feature_source_fields = {
                         'my_feature_id_1': 'my_feature_id_1_source_field',
                      }
-
-            avro_source (gca_io.AvroSource):
-                Optional. This field is a member of `oneof`_ ``source``.
-            bigquery_source (gca_io.BigQuerySource):
-                Optional. This field is a member of `oneof`_ ``source``.
-            csv_source (gca_io.CsvSource):
-                Optional. This field is a member of `oneof`_ ``source``.
-            feature_time_field (str):
-                Optional. Source column that holds the Feature
-                timestamp for all Feature values in each entity.
-
-                This field is a member of `oneof`_ ``feature_time_source``.
-            feature_time (datetime.datetime):
-                Optional. Single Feature timestamp for all entities
-                being imported. The timestamp must not have
-                higher than millisecond precision.
-
-                This field is a member of `oneof`_ ``feature_time_source``.
             entity_id_field (str):
                 Optional. Source column that holds entity IDs. If not provided, entity
                 IDs are extracted from the column named ``entity_id``.
@@ -821,8 +821,8 @@ class EntityType(base.VertexAiResourceNounWithFutureManager):
         Returns:
             gca_featurestore_service.ImportFeatureValuesRequest - request message for importing feature values
         Raises:
-            ValueError if no source or more than one source is provided
-            ValueError if no feature_time_source or more than one feature_time_source is provided
+            ValueError if data_source type is not supported
+            ValueError if feature_time type is not supported
         """
         feature_specs = []
         for feature_id in set(feature_ids):
@@ -845,28 +845,29 @@ class EntityType(base.VertexAiResourceNounWithFutureManager):
             entity_type=self.resource_name, feature_specs=feature_specs,
         )
 
-        # oneof source
-        if avro_source and not bigquery_source and not csv_source:
-            import_feature_values_request.avro_source = avro_source
-        elif not avro_source and bigquery_source and not csv_source:
-            import_feature_values_request.bigquery_source = bigquery_source
-        elif not avro_source and not bigquery_source and csv_source:
-            import_feature_values_request.csv_source = csv_source
+        if isinstance(data_source, gca_io.AvroSource):
+            import_feature_values_request.avro_source = data_source
+        elif isinstance(data_source, gca_io.BigQuerySource):
+            import_feature_values_request.bigquery_source = data_source
+        elif isinstance(data_source, gca_io.CsvSource):
+            import_feature_values_request.csv_source = data_source
         else:
             raise ValueError(
-                "One and only one of `avro_source`, `bigquery_source`, and `csv_source` need to be passed. "
+                f"The type of `data_source` field should be: "
+                f"`gca_io.AvroSource`, `gca_io.BigQuerySource`, or `gca_io.CsvSource`, "
+                f"get {type(data_source)} instead. "
             )
 
-        # oneof feature_time_source
-        if feature_time_field and not feature_time:
-            import_feature_values_request.feature_time_field = feature_time_field
-        elif not feature_time_field and feature_time:
+        if isinstance(feature_time, str):
+            import_feature_values_request.feature_time_field = feature_time
+        elif isinstance(feature_time, datetime.datetime):
             import_feature_values_request.feature_time = utils.get_timestamp_proto(
                 time=feature_time
             )
         else:
             raise ValueError(
-                "One and only one of `feature_time_field` and `feature_time` need to be passed. "
+                f"The type of `feature_time` field should be: `str` or `datetime.datetime`, "
+                f"get {type(feature_time)} instead. "
             )
 
         if entity_id_field is not None:
@@ -926,14 +927,13 @@ class EntityType(base.VertexAiResourceNounWithFutureManager):
 
     def ingest_from_bq(
         self,
-        bq_source_uri: str,
         feature_ids: List[str],
+        feature_time: Union[str, datetime.datetime],
+        bq_source_uri: str,
         batch_create_feature_configs: Optional[
             Dict[str, Dict[str, Union[bool, int, Dict[str, str], str]]]
         ] = None,
         feature_source_fields: Optional[Dict[str, str]] = None,
-        feature_time_field: Optional[str] = None,
-        feature_time: Optional[datetime.datetime] = None,
         entity_id_field: Optional[str] = None,
         disable_online_serving: Optional[bool] = None,
         worker_count: Optional[int] = None,
@@ -943,14 +943,21 @@ class EntityType(base.VertexAiResourceNounWithFutureManager):
         """Ingest feature values from BigQuery.
 
         Args:
-            bq_source_uri (str):
-                Required. BigQuery URI to the input table.
-                Example:
-                    'bq://project.dataset.table_name'
             feature_ids (List[str]):
                 Required. IDs of the Feature to import values
                 of. The Features must exist in the target
                 EntityType, or the request will fail.
+            feature_time (Union[str, datetime.datetime]):
+                Required. The feature_time can be one of:
+                    - The source column that holds the Feature
+                    timestamp for all Feature values in each entity.
+                    - A single Feature timestamp for all entities
+                    being imported. The timestamp must not have
+                    higher than millisecond precision.
+            bq_source_uri (str):
+                Required. BigQuery URI to the input table.
+                Example:
+                    'bq://project.dataset.table_name'
             batch_create_feature_configs (Dict[str, Dict[str, Union[bool, int, Dict[str, str], str]]]):
                 Optional. A user defined Dict containing configurations for feature creation if to create features before ingest.
                 Default to None when features exist in the Featurestore.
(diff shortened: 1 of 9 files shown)


google/cloud/aiplatform/featurestore/feature.py
```diff
@@ -461,3 +461,141 @@ def search(
...
+                underscores and dashes. International characters
+                are allowed.
+                See https://goo.gl/xmQnxf for more information
+                on and examples of labels. No more than 64 user
+                labels can be associated with one Feature
+                (System labels are excluded)."
+                System reserved label keys are prefixed with
+                "aiplatform.googleapis.com/" and are immutable.
+            project (str):
+                Optional. Project to create Feature in. If not set, project
```
diemtvu, line 534: Would it be ignored, or throw error if it is set and fully-qualified entity_type_name is used (may have different value for project)?

Same for location and featurestore id

google/cloud/aiplatform/featurestore/feature.py
```diff
@@ -461,3 +461,141 @@ def search(
...
+        ) = featurestore_utils.validate_and_get_entity_type_resource_ids(
+            entity_type_name=entity_type_name, featurestore_id=featurestore_id,
+        )
+
+        # TODO(b/208269923): Temporary workaround, update when base class supports nested resource
+        cls._resource_noun = (
+            f"featurestores/{featurestore_id}/entityTypes/{entity_type_id}/features"
+        )
+
+        entity_type_name = utils.full_resource_name(
```
diemtvu, line 565: What is the behavior of this function if entity_type_name is fully-qualified but does not consistent with project and location or resource noun?

google/cloud/aiplatform/featurestore/featurestore.py
```diff
@@ -301,24 +302,253 @@ def list_entity_types(
...
     ) -> None:
-        """Deletes entity_type resources in this Featurestore given their entity_type IDs.
+        """Deletes entity_type resources in this Featurestore.
         WARNING: This deletion is permanent.
 
         Args:
             entity_type_ids (List[str]):
-                Required. The list of entity_type IDs to be deleted.
+                Optional. The list of entity_type IDs to be deleted. If entity_type_ids is not set,
+                all entityTypes in this Featurestore will be deleted.
```
diemtvu, line 313: This is convenient, but sounds scary :). I wonder if we should do smt to prevent user accidentally delete their data. Probably we shouldn't add this option.

google/cloud/aiplatform/featurestore/featurestore.py
```diff
@@ -301,24 +302,253 @@ def list_entity_types(
...
+            ]
+
+        for entity_type in entity_types:
+            entity_type.delete(force=force, sync=False)
 
         for entity_type in entity_types:
             entity_type.wait()
+
+    @base.optional_sync()
+    def delete(self, force: bool = None, sync: bool = True) -> None:
```
diemtvu, line 337: nit. None -> False ?

google/cloud/aiplatform/featurestore/featurestore.py
```diff
@@ -301,24 +302,253 @@ def list_entity_types(
...
+        )
+        lro.result()
+        _LOGGER.log_action_completed_against_resource("deleted.", "", self)
+
+    @classmethod
+    @base.optional_sync()
+    def create(
+        cls,
+        featurestore_id: str,
+        online_store_fixed_node_count: Optional[int] = 1,
```
diemtvu, line 373: does the Optional mean it can be omit in the function call (and take the default value of 1), or it can be None? If the later, then we probably remove the Optional (but keep defaut value)?

google/cloud/aiplatform/featurestore/featurestore.py
```diff
@@ -301,24 +302,253 @@ def list_entity_types(
...
+                Required. The ID to use for this Featurestore, which will
+                become the final component of the Featurestore's resource
+                name.
+
+                This value may be up to 60 characters, and valid characters
+                are ``[a-z0-9_]``. The first character cannot be a number.
+
+                The value must be unique within the project and location.
+            online_store_fixed_node_count (int):
+                Required. Config for online serving resources.
```
diemtvu, line 407: Then I guess we should remove the Optional wrapper, right?

google/cloud/aiplatform/featurestore/featurestore.py
```diff
@@ -301,24 +302,253 @@ def list_entity_types(
...
+            Featurestore - Featurestore resource object
+
+        """
+        if labels:
+            utils.validate_labels(labels)
+
+        gapic_featurestore = gca_featurestore.Featurestore(
+            labels=labels,
+            online_serving_config=gca_featurestore.Featurestore.OnlineServingConfig(
+                fixed_node_count=online_store_fixed_node_count
```
diemtvu, line 455: Just a head up, we are planing to have auto-scaling to public preview. Which mean, setting this count will need to change to setting scaling_config (which may need 2 numbers). We can revisit this when it's done, just fyi for now.

tests/system/aiplatform/test_featurestore.py
```diff
@@ -15,24 +15,172 @@
...
+        shared_state["resources"] = [featurestore]
+
+        get_featurestore = aiplatform.Featurestore(
+            featurestore_name=featurestore.resource_name
+        )
+        assert featurestore.resource_name == get_featurestore.resource_name
 
         list_featurestores = aiplatform.Featurestore.list()
-        assert len(list_featurestores) >= 0
+        assert len(list_featurestores) - base_list_featurestores == 1
```
diemtvu, line 71: I'm not sure what is the style guide with OSS, should we follow one assert per test function or use subtest?

google/cloud/aiplatform/featurestore/featurestore.py
morgandu: I totally agree. I was hoping to get some feedback on this function. I can remove the delete all by default, and add an example in the docstring on how to get `entity_type_ids` from `list_entity_types()` to achieve the delete all functionality. If later there is more usage request on the delete all child resources, I can add a separate `delete_all_entity_types()`.

google/cloud/aiplatform/utils/featurestore_utils.py
```diff
@@ -115,3 +126,119 @@ def validate_and_get_feature_resource_ids(
...
+            )
+        else:
+            create_feature_request = gca_featurestore_service.CreateFeatureRequest(
+                feature=gapic_feature, feature_id=self._get_feature_id()
+            )
+
+        return create_feature_request
+
+
+def get_timestamp_proto(
```
sasha-gitg, line 229: Can we move this into general utilities since it's not featurestore specific?

tests/system/aiplatform/e2e_base.py
```diff
@@ -104,7 +104,9 @@ def teardown(self, shared_state: Dict[str, Any]):
 
         for resource in shared_state["resources"]:
             try:
-                if isinstance(resource, aiplatform.Endpoint):
+                if isinstance(resource, aiplatform.Endpoint) or isinstance(
+                    resource, aiplatform.Featurestore
+                ):
```
sasha-gitg, line 109:
```suggestion
                if isinstance(resource, (aiplatform.Endpoint, aiplatform.Featurestore)):
```

sasha-gitg (changes requested):

google/cloud/aiplatform/featurestore/featurestore.py
morgandu: Made it `Optional` to be consistent with defined in the the Args, where removed the 1 value, added more description to indicate 1 node count will be launched without specifying anything.

google/cloud/aiplatform/featurestore/featurestore.py
morgandu: Noted

google/cloud/aiplatform/featurestore/featurestore.py
morgandu: Updated the `online_store_fixed_node_count` to `Required` with default to 1, also added more docstring description.
Service api at this moment requires `Node count must be at least 1.`

google/cloud/aiplatform/featurestore/featurestore.py
morgandu: For the current setting, the `online_store_fixed_node_count` should set to `Required` with 1 as default. But since there are some pending updates i.e. no online store launched by default when creating fs, I make it Optional to None, but in docstring, indicated when not specified 1 is in use.

google/cloud/aiplatform/featurestore/featurestore.py
morgandu: See comment/question above.

google/cloud/aiplatform/utils/featurestore_utils.py
```diff
@@ -115,3 +126,119 @@ def validate_and_get_feature_resource_ids(
...
+
+    def _get_value_type_enum(self) -> int:
+        """Validates value_type and returns the enum of the value type.
+
+        Returns:
+            int - valid value type enum.
+        """
+
+        # Raises ValueError if invalid value_type
+        validate_value_type(value_type=self.value_type)
```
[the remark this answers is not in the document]
morgandu: The current `validate_id` is more of a if it is an id, to support and validate if user correctly provided the resource id with its ancestors' resource id/full path or, only the resource full path. Changed the function to `is_resource_id` to save the `validate_id` for later use.

google/cloud/aiplatform/featurestore/feature.py
```diff
@@ -461,3 +461,141 @@ def search(
...
+            feature_id=feature_id,
+            value_type=value_type,
+            description=description,
+            labels=labels,
+        )
+        create_feature_request = feature_config.get_create_feature_request(
+            parent=entity_type_name
+        )
+
+        api_client = cls._instantiate_client(location=location, credentials=credentials)
```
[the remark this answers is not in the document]
morgandu: Updated the `location` using the full resource name of the creating resource's parent in `create` of both `Feature` and `EntityType` class.

google/cloud/aiplatform/featurestore/feature.py
morgandu: If `entity_type_name` is provided as fully-qualified resource name, the `project` and `location` args will be ignored.
Added description to indicate `project` and `location` are needed only when direct parent resource name arg is provided as a resource ID for both `create` methods in `EntityType` and `Feature`.

google/cloud/aiplatform/featurestore/feature.py
morgandu: If the `resource_name` arg for `utils.full_resource_name` is identified as a valid fully-qualified resource name, the arg value will be the return value of the method, and the overwrite `project` and `location` args will be ignored. 

The `resource_noun` is optional and only needed when the the `resource_name` arg value is identified as an invalid fully-qualified resource name, then a project and a location plus the `resource_noun` arg will be required to form the fully-qualified resource name. If `project` and `location` args can still be optional, if `project` and `location` set through `aiplatform.init`.

tests/system/aiplatform/test_featurestore.py
morgandu: Double checking the integration test standard (we did plan to have simple end to end test, but I can see this getting complicated), will update upon confirmation.

google/cloud/aiplatform/featurestore/entity_type.py
```diff
@@ -381,3 +386,869 @@ def delete_features(self, feature_ids: List[str], sync: bool = True,) -> None:
...
+
+        """
+        if batch_create_feature_configs:
+            self.batch_create_features(
+                feature_configs=batch_create_feature_configs,
+                request_metadata=request_metadata,
+                sync=sync,
+            )
+
+            if not sync:
```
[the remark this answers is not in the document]
morgandu: Tested when move the `base.optional_sync` from `_import_feature_values` to outer `ingest_*` methods. When the ingestion (with batch create) started, it hangs, nothing come back. Is that possible that the `batch_create_features` is already wrapped and `batch_create_features` is inside the `ingest_*`. 

It works fine when changing back to wrap the `_import_feature_values` and made the `batch_create_features` inside the `ingest_*` sync=True by default.

google/cloud/aiplatform/featurestore/entity_type.py
```diff
@@ -381,3 +386,836 @@ def delete_features(self, feature_ids: List[str], sync: bool = True,) -> None:
...
+        )
+
+        return self
+
+    def ingest_from_bq(
+        self,
+        feature_ids: List[str],
+        feature_time: Union[str, datetime.datetime],
+        bq_source_uri: str,
+        batch_create_feature_configs: Optional[
```
diemtvu, line 925: I thought we decide not to do this yet?

And if we do, then it's probably better to use a boolean arg, e.g create_if_missing, which, if turned on, we will compare the existing schema vs the columns in the data source and create features accordingly. It will be filled with some default description, with the name and type best match with data type (there could be cases we map to int but user want double etc, so may not be 100% robust). 

Another option is adding API to infer schema from the data source. It works the same as above, but only return the feature creation request (or feature configs) so that user can patch (e.g update the description, labels) and call batch_features to complete.


Anyway, both these are really fancy and beyond the core for the first delivery. Wdyt?

tests/unit/aiplatform/test_utils.py
```diff
@@ -360,6 +363,15 @@ def test_client_w_override_select_version():
     )
 
 
+def test_get_timestamp_proto():
```
diemtvu, line 366: Can we use (or based on) the `get_timestamp_proto` in util?

google/cloud/aiplatform/featurestore/entity_type.py
```diff
@@ -381,3 +386,836 @@ def delete_features(self, feature_ids: List[str], sync: bool = True,) -> None:
...
+        )
+
+        _LOGGER.log_action_started_against_resource_with_lro(
+            "Batch create Features",
+            "entityType",
+            self.__class__,
+            batch_created_features_lro,
+        )
+
+        batch_created_features_lro.result()
```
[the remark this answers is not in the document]
morgandu: The `sync` arg does not work directly toward the `*_lro.result()`. The `_lro.result()` will essentially wait no matter what depends on the service response, and the `_LOGGER` with capture when the result is ready.

With the default behavior `sync=True`, the `return self` will not happen till the result is ready. But when `sync=False`, where it actually gets affect is the decorator method `@base.optional_sync(return_input_arg="self")`. So the user can continue with other class while this particular class is waiting for lro result.

tests/unit/aiplatform/test_utils.py
morgandu: `assert true_timestamp_proto == utils.get_timestamp_proto(time)`

This unit test is testing the `utils.get_timestamp_proto`? I am not quiet sure I understand the question.

google/cloud/aiplatform/featurestore/entity_type.py
morgandu: My impression of what we decided not to do (through the design meeting) is not allow users to add, delete, or update/maintain feature configs before sending the batch create feature request. The goal is to ensure stateless within the SDK. But we were discussing utilizing python's native data structure i.e. dictionary, so user can maintain their creating feature id as key without duplicating cases, and SDK can take cares of certain validations, i.e. correct format of `feature_id` (no reserved keywords) and correct type of `value_type`, and `labels`. Also `labels` and `description` are all default to None. We try to avoid map any `type` in general but force user to define, for exactly the reason you mentioned.

For the `batch_create_feature_configs`, users need to specify correctly the `feature_id`. And, no matter what type is used in the BigQuery, users need to specify the `value_type` for the feature in Feature Store.

The implementation for validation and generate the `requests` for `batch_create_features` are in two places:
`entity_type.EntityType._validate_and_get_create_feature_requests` and `featurestore_utils._FeatureConfig.get_create_feature_request`.

May be it's easier to have a call if I misunderstand your question. 

Apart from that. I think having an API to print all features' information is very useful for users. We have so far `list_features` but users need to retrieve those feature level info by iterate through all the listed Feature classes. It will be much simpler at EntityType level to have a glance of all features. However, I would prefer to implement this after the initial scope is implemented. What do you think? Also tagging @sasha-gitg

sasha-gitg (changes requested):
One high level comment: if possible, please avoid force updating the commit history as some of the previous commits are no longer accessible and it becomes difficult to follow previous comments. It also seems to break "Show changes since last review" view. Thanks!

google/cloud/aiplatform/featurestore/entity_type.py
morgandu: sounds good

google/cloud/aiplatform/featurestore/entity_type.py
```diff
@@ -381,3 +386,836 @@ def delete_features(self, feature_ids: List[str], sync: bool = True,) -> None:
...
+
+        return requests
+
+    @base.optional_sync(return_input_arg="self")
+    def batch_create_features(
+        self,
+        feature_configs: Dict[str, Dict[str, Union[bool, int, Dict[str, str], str]]],
+        request_metadata: Optional[Sequence[Tuple[str, str]]] = (),
+        sync: bool = True,
+    ) -> "EntityType":
```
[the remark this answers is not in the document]
morgandu: I think the main reason is if returning a list of Features after a batch creation is useful or not for users. The operation I can think of that require accessing the Feature class is checking Feature properties and updating it. But it's not likely to perform it on all Features/after a batch creation, also, iterating through a list of Feature classes is not as clear as to retrieve an individual Feature by ID.

google/cloud/aiplatform/featurestore/entity_type.py
diemtvu: +1. I think it's cleaner to separate ingest and creation for now, until we gather more feedback.

google/cloud/aiplatform/featurestore/entity_type.py
```diff
@@ -381,3 +386,836 @@ def delete_features(self, feature_ids: List[str], sync: bool = True,) -> None:
...
+        feature_time: Union[str, datetime.datetime],
+        bq_source_uri: str,
+        batch_create_feature_configs: Optional[
+            Dict[str, Dict[str, Union[bool, int, Dict[str, str], str]]]
+        ] = None,
+        feature_source_fields: Optional[Dict[str, str]] = None,
+        entity_id_field: Optional[str] = None,
+        disable_online_serving: Optional[bool] = None,
+        worker_count: Optional[int] = None,
+        request_metadata: Optional[Sequence[Tuple[str, str]]] = (),
```
diemtvu, line 932: nit. Should the default be None ?

google/cloud/aiplatform/featurestore/entity_type.py
```diff
@@ -381,3 +386,836 @@ def delete_features(self, feature_ids: List[str], sync: bool = True,) -> None:
...
+            entity_type_id (str):
+                Required. The ID to use for the EntityType, which will
+                become the final component of the EntityType's resource
+                name.
+
+                This value may be up to 60 characters, and valid characters
+                are ``[a-z0-9_]``. The first character cannot be a number.
+
+                The value must be unique within a featurestore.
+            featurestore_name (str):
```
diemtvu, line 458: nit. Should we mentioned that featurestore (or parent resource in general) should exist before this call? Same for other create function.

google/cloud/aiplatform/featurestore/entity_type.py
morgandu: We carried the default value `()` from GAPIC, https://github.com/googleapis/python-aiplatform/blob/main/google/cloud/aiplatform_v1/services/featurestore_service/client.py#L430, that allows us avoid change the `None` to `()` (requires an iterable data type) if not set across all SDK call.

sasha-gitg (approved):
LGTM! Thanks!

diemtvu (approved):

Issue #853, opened 2021-11-18
Make display_name parameter optional for all resources


Ark-kun:
I think we should try to reduce the number of required parameters that the user has to specify to be able to call a given API. The `display_name` is easy to generate in a way that will satisfy most users.

I propose to mark the `display_name` parameter as optional and generate it as `<Resource> 2021-11-18 15:28:01` by default.

PR #882, opened 2021-12-03
feat: Made display_name parameter optional for most calls

Ark-kun:

Fixes https://github.com/googleapis/python-aiplatform/issues/853 🦕


google/cloud/aiplatform/datasets/image_dataset.py
```diff
@@ -126,6 +127,8 @@ def create(
             image_dataset (ImageDataset):
                 Instantiated representation of the managed image dataset resource.
         """
+        if not display_name:
+            display_name = cls.name + " " + datetime.datetime.now().isoformat(sep=" ")
```
sasha-gitg, line 131: Please capture `display_name` generation as a shared utility

sasha-gitg (changes requested):
Please add/update unit tests to test this change.

google/cloud/aiplatform/datasets/dataset.py
```diff
@@ -105,12 +106,12 @@ def _validate_metadata_schema_uri(self) -> None:
     @classmethod
     def create(
         cls,
-        display_name: str,
```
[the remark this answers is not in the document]
Ark-kun: Unfortunately, required arguments cannot  follow optional arguments.
Would it be OK to make `metadata_schema_uri` optional as well and validate it in code?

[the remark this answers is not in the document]
sasha-gitg: I don't think there is a change that could make `display_name` optional and not introduce breaking changes. Current guidance is to leave the signatures where this is the case as-is and we can:

1. Makes these changes in the next major version rev (v2)
2. We can consider the tradeoffs of enforcing keyword-only parameters in the next major version rev to provide greater flexibility for these types of changes.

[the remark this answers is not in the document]
Ark-kun: Note that the class in question (`_Dataset`) is private...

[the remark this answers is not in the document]
Ark-kun: I've reverted this change as requested.

google/cloud/aiplatform/datasets/image_dataset.py
Ark-kun: Good idea. Done.

Ark-kun:
I think I've resolved all feedback.

Ark-kun:
What are the remaining action items for me here?

Ark-kun:
> @Ark-kun [the change of argument order in public methods](https://github.com/googleapis/python-aiplatform/pull/882/files#diff-21789052f0ef42ae38b04240e8e9bdc213aaa8fce97a946ae0e17858a5b018f0R6356-R6363) is considered a breaking change and cannot be merged

Thank you for the link. I've missed this one. I'll restore the ordering. Thank you.

google/cloud/aiplatform/datasets/dataset.py
```diff
@@ -208,7 +208,8 @@ def create(
             dataset (Dataset):
                 Instantiated representation of the managed dataset resource.
         """
-
+        if not display_name:
```
sasha-gitg, line 211: We should explicitly check if display_name is `None` to not override in the case of an empty string.

google/cloud/aiplatform/jobs.py
```diff
@@ -344,7 +344,7 @@ def completion_stats(self) -> Optional[gca_completion_stats.CompletionStats]:
     @classmethod
     def create(
         cls,
-        job_display_name: str,
+        job_display_name: Optional[str],
```
sasha-gitg, line 347: Preference to remove this pattern as valid usage would require passing `None` explicitly.

Please remove throughout where `display_name` is a required argument and cannot be changed to have a default argument because it breaks order of arguments.

google/cloud/aiplatform/jobs.py
```diff
@@ -1010,9 +1012,9 @@ class CustomJob(_RunnableJob):
 
     def __init__(
         self,
-        display_name: str,
         worker_pool_specs: Union[List[Dict], List[aiplatform.gapic.WorkerPoolSpec]],
         base_output_dir: Optional[str] = None,
+        display_name: Optional[str] = None,
```
sasha-gitg, line 1017: Breaking change on order of arguments.

google/cloud/aiplatform/jobs.py
```diff
@@ -1494,6 +1498,7 @@ def __init__(
         max_failed_trial_count: int = 0,
         search_algorithm: Optional[str] = None,
         measurement_selection: Optional[str] = "best",
+        display_name: Optional[str] = None,
```
sasha-gitg, line 1501: Breaking change as well.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -1013,7 +1015,8 @@ class _CustomTrainingJob(_TrainingJob):
 
     def __init__(
         self,
-        display_name: str,
+        # TODO: Make display_name parameter fully optional in next major release
```
sasha-gitg, line 1018: Please create and add a buganizer ticket.

sasha-gitg (changes requested):

google/cloud/aiplatform/datasets/dataset.py
Ark-kun: Note that resources with empty display name are very hard to use in Pantheon.
In Pantheon the user must click on the resource name in the first column. But that's impossible when the name is empty.
I've logged a bug about this, but I'm not sure whether it's going to be fixed since it's a common Pantheon UX control.

google/cloud/aiplatform/jobs.py
Ark-kun: This is a type annotation. Type annotations specify what argument types are valid for a parameter.
In Python `Optional[T]` is the same as `Union[T, None]`. In Google codebases I've also seen tha `Optional[str]` means `str or None`. Such annotation states that `None` is a valid argument for a parameter. Otherwise, when the user passes `display_name=None`, some type checkers can complain that `None` is not a valid argument type for this parameter.

google/cloud/aiplatform/training_jobs.py
Ark-kun: Done.

google/cloud/aiplatform/jobs.py
Ark-kun: Fixed

google/cloud/aiplatform/jobs.py
Ark-kun: Fixed

google/cloud/aiplatform/datasets/dataset.py
sasha-gitg: Got it, thanks for the context. I agree with your approach.

google/cloud/aiplatform/jobs.py
sasha-gitg: Understand. My preference is to not recommend users pass `None` explicitly as a workaround to our inability to change the argument ordering. This could have downstream effect on samples, notebooks, and other ancillary artifacts being created with this type of usage:

```
job = BatchPredictionJob.create(job_display_name=None, model_name='projects/.../my_model')
``` 

Which, to me, seems to indicate that display_name is not set.

Perhaps we can update the docstrings to indicate the format of the name when passed `None`.

@morgandu @kweinmeister Thoughts? If you think this usage is OK then we can merge in. I'm open to changing this view.

kweinmeister: I have not seen this style of explicitly passing in `None`, and then having the value overridden in the function.

In the example above, anyone reading the code without knowledge of the `BatchPredictionJob` internals would assume that the display name is set to `None`.

I would suggest we start with applying this change only to functions where we can specify `Optional[str] = None`. It will have a clear usability benefit. We can always come back later with another PR for cases in which `display_name` is required, after we have given it more thought.

Ark-kun: Having `Optional[T]` parameter without a default value is pretty common even in the Vertex SDK.

However I can remove the annotation changes as asked.

https://github.com/googleapis/python-aiplatform/blob/89e321246b6223a2355947d8dbd0161b84523478/google/cloud/aiplatform/initializer.py#L102

https://github.com/googleapis/python-aiplatform/blob/5735545c647075d00c0d18f4a3319c76be8dc110/google/cloud/aiplatform/models.py#L327

https://github.com/googleapis/python-aiplatform/blob/5735545c647075d00c0d18f4a3319c76be8dc110/google/cloud/aiplatform/models.py#L328

https://github.com/googleapis/python-aiplatform/blob/ac0c4e79bc2882480f5e7a0dc4d8c051a6759bbf/google/cloud/aiplatform/training_jobs.py#L1676

https://github.com/googleapis/python-aiplatform/blob/ac0c4e79bc2882480f5e7a0dc4d8c051a6759bbf/google/cloud/aiplatform/training_jobs.py#L2134

https://github.com/googleapis/python-aiplatform/blob/ac0c4e79bc2882480f5e7a0dc4d8c051a6759bbf/google/cloud/aiplatform/training_jobs.py#L3267

https://github.com/googleapis/python-aiplatform/blob/81c403a6b833aae3836f3b3a98f8f09602839c05/google/cloud/aiplatform/models.py#L352

https://github.com/googleapis/python-aiplatform/blob/93acea098e01cc9931a403037c309217cd391be2/google/cloud/aiplatform/training_jobs.py#L557

https://github.com/googleapis/python-aiplatform/blob/57b12412ff5b79bf5819f82cd807491a48cfcf93/google/cloud/aiplatform/datasets/dataset.py#L257

https://github.com/googleapis/python-aiplatform/blob/b613b264524aaab2cb65e63a5487770736faa7c8/google/cloud/aiplatform/tensorboard/plugins/tf_profiler/profile_uploader.py#L81

Ark-kun: As asked, I've removed the `Optional[str]` annotation and docstring changes in cases where specifying the default value is not possible.

kweinmeister (approved):

sasha-gitg (approved):

PR #886, opened 2021-12-06
fix: add clarity to parameters per user feedback

andrewferlitsch:

### DIFF
--- a/samples/model-builder/predict_tabular_classification_sample.py
+++ b/samples/model-builder/predict_tabular_classification_sample.py
@@ -29,8 +29,8 @@ def predict_tabular_classification_sample(
     Args
         project: Your project ID or project number.
         location: Region where Endpoint is located. For example, 'us-central1'.
-        endpoint: A fully qualified endpoint name or endpoint ID. Example: "projects/123/locations/us-central1/endpoints/456" or
- |              "456" when project and location are initialized or passed.
+        endpoint_name: A fully qualified endpoint name or endpoint ID. Example: "projects/123/locations/us-central1/endpoints/456" or
+               "456" when project and location are initialized or passed.
         instances: A list of one or more instances (examples) to return a prediction for.
     '''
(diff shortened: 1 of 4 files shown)


samples/model-builder/predict_tabular_classification_sample.py
```diff
@@ -22,9 +22,17 @@
 def predict_tabular_classification_sample(
     project: str, location: str, endpoint: str, instances: List[Dict],
 ):
+    '''
+    Args
+        project: Your project ID or project number.
+        location: Region where Endpoint is located. For example, 'us-central1'.
+        endpoint: A fully qualified endpoint name or endpoint ID. Example: "projects/123/locations/us-central1/endpoints/456" or
```
morgandu, line 29: Docstring also need to reflect the change from `endpoint` to `endpoint_name`

samples/model-builder/predict_tabular_classification_sample.py
```diff
@@ -22,9 +22,17 @@
 def predict_tabular_classification_sample(
     project: str, location: str, endpoint: str, instances: List[Dict],
```
morgandu, line 23: based on the change of this PR, `endpoint` should also be changed to `endpoint_name`

morgandu (changes requested):

samples/model-builder/predict_tabular_classification_sample.py
andrewferlitsch: fixed

samples/model-builder/predict_tabular_classification_sample.py
andrewferlitsch: fixed

samples/model-builder/predict_tabular_classification_sample.py
andrewferlitsch: fixed

morgandu (approved):

PR #898, opened 2021-12-08
fix: Improve handling of undeploying model without redistributing remaining traffic

vinnysenthil:
- Add informative error when undeploying a model with traffic from an Endpoint with multiple deployed models, without providing a new traffic split.
- Improve accuracy of docstring for `Endpoint.undeploy()`
- Add tests to cover CUJs

Fixes [b/198290421](http://b/198290421) 🦕

### DIFF
--- a/google/cloud/aiplatform/models.py
+++ b/google/cloud/aiplatform/models.py
@@ -993,22 +993,23 @@ class Endpoint(base.VertexAiResourceNounWithFutureManager):
         sync=True,
     ) -> None:
         """Undeploys a deployed model.
-
-        Proportionally adjusts the traffic_split among the remaining deployed
-        models of the endpoint.
+        
+        The model to be undeployed should have no traffic or user must provide
+        a new traffic_split with the remaining deployed models. Refer
+        to `Endpoint.traffic_split` for the current traffic split mapping.
 
         Args:
             deployed_model_id (str):
                 Required. The ID of the DeployedModel to be undeployed from the
                 Endpoint.
             traffic_split (Dict[str, int]):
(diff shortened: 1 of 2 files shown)


google/cloud/aiplatform/models.py
```diff
@@ -1023,6 +1024,20 @@ def undeploy(
                     "Sum of all traffic within traffic split needs to be 100."
                 )
 
+        elif (
+            len(self.traffic_split) > 1  # Two or more models deployed to Endpoint
```
sasha-gitg, line 1028: Shouldn't this check for the case where all other deployed models' traffic split sums to 0? Or at least that's how I understood the bug.

sasha-gitg: Also, shouldn't `undeploy_all` sort the models so that the lowest traffic split, potentially 0%, is undeployed first so we don't misconfigure the endpoint.

vinnysenthil: You're right, this covers a different case:

```py
ep.traffic_split
# { 'm1': 35, 'm2': 65 }

ep.undeploy('m1'). # Raises error about "would leave the remaining traffic split at 65%"
```

shall I leave this case in while adding the following two changes:

```py
ep.traffic_split
# { 'm1': 0, 'm2': 0, 'm3':100 }

ep.undeploy('m3'). # Raises error about needing to undeploy 'm1', 'm2' first ​
```

```py
ep.traffic_split
# { 'm1': 0, 'm2': 100, 'm3':0 }


ep.undeploy_all()  # Undeploys 'm1', 'm3' first, then 'm2'
```

sasha-gitg (approved):

PR #908, opened 2021-12-14
fix: add support for API base path overriding

sasha-gitg:
Fixes [b/210643836](b/210643836) 🦕 and failing unit tests.


mkovalski (approved):
LGTM from profiler side, but might need to update the tensorboard uploader as well. https://github.com/googleapis/python-aiplatform/blob/main/google/cloud/aiplatform/tensorboard/uploader_main.py#L34

yfang1 (approved):

vinnysenthil (approved):

PR #919, opened 2021-12-17
feat: enable feature store batch serve to BigQuery and GCS for csv and tfrecord

morgandu:
- [x] added `batch_serve_to_bq` and `batch_serve_to_gcs`
- [x] added unit tests
- [x] added integration tests


google/cloud/aiplatform/featurestore/featurestore.py
```diff
@@ -557,3 +562,597 @@ def create_entity_type(
...
+                read_instances = [read_instances]
+
+        return gca_io.CsvSource(gcs_source=gca_io.GcsSource(uris=read_instances))
+
+    @base.optional_sync(return_input_arg="self")
+    def batch_serve_to_bq(
+        self,
+        bq_destination_output_uri: str,
+        entity_type_ids: List[str],
+        entity_type_destination_fields: Optional[
```
diemtvu, line 845: This argument name is a bit confusing. It is to specify what features to read, and (optional) the output column name, right? May be change it to smt related to feature than entity type, e.g `feature_to_read` or some sort.

Alternatively, we can have a separate struct to define the entity_type and what feature to read within that entity type, and provide a list of that to this function, instead of the separate entity_type_ids and this entity_type_destination_field.


Another option is, we can have a flat list of feature to read, using the full-resource path name. Internally, we can group them by entity type and construct the read request. This could be a little bit repetitive on the calling side, but I think the interface is more intuitive. We'll need to use different struct for 'read_instance' though (e.g Dict instead of List)

google/cloud/aiplatform/featurestore/featurestore.py
```diff
@@ -557,3 +562,597 @@ def create_entity_type(
...
+                            'my_entity_type_id_1': {},
+                            'my_entity_type_id_2': [],
+                            'my_entity_type_id_3': None,
+                         }
+
+                     - In case selected features will be batch served and using Feature ID as destination field name:
+
+                         entity_type_ids = ['my_entity_type_id_1', 'my_entity_type_id_2', 'my_entity_type_id_3']
+
+                         feature_source_fields = {
```
diemtvu, line 704: Do you mean `entity_type_destination_fields`?

google/cloud/aiplatform/featurestore/featurestore.py
morgandu: Isn't `read_instances` the bq uri or gcs uri(s)? How will dict work here? 

Regarding renaming the `entity_type_destination_fields`, how about `feature_destination_fields` to match the `feature_source_fields` for ingestion in EntityType?

Regarding the data type and usage of the `entity_type_destination_fields`:
1) I think the current implementation is trying to implement the first alternative suggested here. The purpose for having a separate `entity_type_ids` and `entity_type_destination_fields` is that for some `entity_type` that user want to batch serve all features in it, without a different output feature name, user won't need to include a list of features or a mapping (feature to output name) for that entity_type in the `entity_type_destination_fields`. We can remove the separated `entity_type_ids` field and make it required to have all `entity_type_id` specified in the `entity_type_destination_fields`'s key, with ["*"] value if all features are to be served. For example,
```
{
    "entity_type_1": ["*"], 
    "entity_type_2": {"feature_1": "feature_1_col"}, 
    "entity_type_3": {"feature_1", "feature_2"}, 
}
```

2) the second option is a bit redundant but it is indeed very clear and straight forward
{"...path/to/fs/fs1/et/e1/f/f1": "f11_col", "...path/to/fs/fs1/et/e2/f/f1": "f21_col"}

Will the modification of the current implementation in 1) be less confusing and more appealing for usage?

diemtvu: Regarding [1], you also have the option to support mapping the output column name to something different than feature name. I feel this overloaded argument is hard for users to understand (though I may agree it could be convenient once user get a good grasp of it)

I think it is easier to understand to split into 2 arguments: one to define the feature to read, the other one to define the  (overridden) output column name. This also allow user to define the output column only for the feature they want to (i.e they want to read */all features from entity_type_1 and output to the column with the same name, except feature `foo` will be output to column `bar`) 

With the flat-list, the entity type IDs can be extracted from the feature path name. However, unless we put some constraints that the list have to group by entity type, the order of the entity type in the `read_instance` could be ambiguous. Thus I think using Dict (e.g `{'entity_type_1' : 'user_1234', 'entity_type_2': 'product_xyz' }` is clearer than using array (`['user_1234', 'product_xyz']`). The flat list option also requires all to be explicitly listed, i.e "*" is not supported. This is the biggest cons for that option.

google/cloud/aiplatform/featurestore/featurestore.py
morgandu: fixed the docstring

google/cloud/aiplatform/featurestore/featurestore.py
```diff
@@ -556,3 +561,461 @@ def create_entity_type(
...
+                Optional. When not empty, the specified fields in the
+                read_instances source will be joined as-is in the output,
+                in addition to those fields from the Featurestore Entity.
+
+                For BigQuery source, the type of the pass-through values
+                will be automatically inferred. For CSV source, the
+                pass-through values will be passed as opaque bytes.
+
+        Returns:
+            Featurestore - The featurestore resource object batch read feature values from.
```
sasha-gitg, line 854:
```suggestion
            Featurestore: The featurestore resource object batch read feature values from.
```

This formats correctly on Google Cloud docs.

google/cloud/aiplatform/featurestore/featurestore.py
```diff
@@ -556,3 +561,461 @@ def create_entity_type(
...
```
sasha-gitg, line 984:
```suggestion
            Featurestore: The featurestore resource object batch read feature values from.
```

google/cloud/aiplatform/featurestore/featurestore.py
morgandu: removed ```entity_type_ids``` and ```entity_type_destination_fields``` to avoid confusion, added two dict ```serving_feature_ids``` for an entity_type / features mapping and ```feature_destination_fields ``` for feature_id / feature_destination_field_name mapping.

diemtvu (approved):
LGTM

PR #923, opened 2021-12-21
docs(samples): add AutoML image classification sample

andrewferlitsch:

Fixes #202042079🦕


andrewferlitsch:
The build is failing:

import tensorflow as tf
E   ModuleNotFoundError: No module named 'tensorflow'

That seems to be part of the testing pipeline, so not sure how to fix.

morgandu:
Please change the PR title from "fix: ..." to "samples: ...".

"fix: ..." and "feat: ..." are reserved for Vertex SDK and the title content will be propagated to the Release Note.

morgandu (changes requested):
Please change the PR title from "fix: ..." to "samples: ...".

"fix: ..." and "feat: ..." are reserved for Vertex SDK and the title content will be propagated to the Release Note.

kweinmeister:
Also, to add to @morgandu's comment, it's preferred if we don't refer to internal bug numbers that some users may not have access to.

So, could the title also be changed to include a summary of what was actually fixed? That will make the Release Notes for this fix more useful.

samples/model-builder/predict_image_classification_sample.py
```diff
@@ -0,0 +1,56 @@
...
+
+from typing import List
+
+from google.cloud import aiplatform
+
+import tensorflow as tf
+
+
+#  [START aiplatform_sdk_predict_image_classification_sample]
+def predict_tabular_classification_sample(
```
kweinmeister, line 26: Should this read `predict_image_classification_sample`?

andrewferlitsch: yes, that looks like a typo.

andrewferlitsch: I fixed it.

kweinmeister (approved):

morgandu (approved):

PR #943, opened 2022-01-06
fix: Fixed getting project ID when running on Vertex AI; Fixes #852

### LINKED ISSUE #852: The SDK does not pick up the correct user project when running on Vertex Pipelines or CustomJobs
Usually the Vertex SDK gets the project ID automatically (by calling `google.auth.default()`). This works when running on Kubeflow Pipelines.
However when running on Vertex Pipelines or CustomJobs, the detected project is not the user project and is not usable.
This leads to failure when trying to create any resource in the project:
>google.api_core.exceptions.PermissionDenied: 403 Permission 'aiplatform.models.upload' denied on resource '//aiplatform.googleapis.com/projects/gbd40bc90c7804989-tp/locations/us-central1' (or it may not exist).

Fortunately there is a way to get project number from the Vertex environment. There is also a way to get project ID from the project number.

Deducing project number
```python
project_number = os.environ.get("CLOUD_ML_PROJECT_ID")
```

Getting project ID:
```python
    if not project:
        project_number = os.environ.get("CLOUD_ML_PROJECT_ID")
        if project_number:
            print(f"Inferred project number: {project_number}")
            project = project_number
            # To improve the naming we try to convert the project number into the user project ID.
            try:
                from googleapiclient import discovery

                cloud_resource_manager_service = discovery.build(
                    "cloudresourcemanager", "v3"
                )
                project_id = (
                    cloud_resource_manager_service.projects()
                    .get(name=f"projects/{project_number}")
                    .execute()["projectId"]
                )
                if project_id:
                    print(f"Inferred project ID: {project_id}")
                    project = project_id
            except Exception as e:
                print(e)
```

Perhaps we should improve the Vertex SDK initializer to use the project mentioned in `CLOUD_ML_PROJECT_ID` if it's available by default.

Ark-kun:

Fixes #852 🦕


setup.py
```diff
@@ -89,6 +89,7 @@
         "packaging >= 14.3",
         "google-cloud-storage >= 1.32.0, < 2.0.0dev",
         "google-cloud-bigquery >= 1.15.0, < 3.0.0dev",
+        "google-api-python-client >= 2.29",  # for API discovery
```
vinnysenthil, line 92: The Python discovery client is something we'd want to avoid adding, could we remove the project number to project ID conversion for now? We can aim to alter `CLOUD_ML_PROJECT_ID` service-side or add another env var in Custom Jobs / Pipelines that explicitly contains the project ID

vinnysenthil (changes requested):

setup.py
Ark-kun: >The Python discovery client is something we'd want to avoid adding

Could you share the main issues making adding this package undesirable compared to other packages? 

>could we remove the project number to project ID conversion for now?

This is possible, but some consequences would be not quite desirable. Namely, the staging bucket names are currently in the form `<project-id>-vertex-staging-<region>`. Using project numbers would result in less readable names.

>We can aim to alter CLOUD_ML_PROJECT_ID service-side or add another env var in Custom Jobs / Pipelines that explicitly contains the project ID

I agree. That solution would be the best.

Ark-kun: I've removed this dependency. Let's revisit this decision later.

vinnysenthil (approved):

sasha-gitg (changes requested):
Please add a unit test that tests this block. Preferably, also an integration test that executes a custom job with the SDK.

google/cloud/aiplatform/initializer.py
```diff
@@ -149,6 +149,35 @@ def project(self) -> str:
...
+                cloud_resource_manager_service = discovery.build(
+                    "cloudresourcemanager", "v3"
+                )
+                project_id = (
+                    cloud_resource_manager_service.projects()
+                    .get(name=f"projects/{project_number}")
+                    .execute()["projectId"]
+                )
+                self._project = project_id
+            except ImportError:
```
sasha-gitg, line 173: Preference to reduce the scope of this block to just the import statement.

google/cloud/aiplatform/initializer.py
```diff
@@ -149,6 +149,35 @@ def project(self) -> str:
...
+                logging.debug(
+                    "Need google-api-python-client to convert project number to project ID."
```
sasha-gitg, line 175:
```suggestion
                    "Please install google-api-python-client to convert project number to project ID."
```

google/cloud/aiplatform/initializer.py
```diff
@@ -149,6 +149,35 @@ def project(self) -> str:
...
+        # See https://github.com/googleapis/python-aiplatform/issues/852
+        # See https://github.com/googleapis/google-auth-library-python/issues/924
+        # TODO: Remove when google.auth.default() learns the
+        # CLOUD_ML_PROJECT_ID env variable or Vertex AI starts setting GOOGLE_CLOUD_PROJECT env variable.
+        project_number = os.environ.get("CLOUD_ML_PROJECT_ID")
+        if project_number:
+            self._project = project_number
+            # Try to convert project number to project ID which is more readable.
+            try:
+                from googleapiclient import discovery
```
sasha-gitg, line 162: Preference to move the conversion of project number to project ID to shared utility so it can used elsewhere in the future. This could be useful in scenarios where project number is not supported, like BQ.

cc: @morgandu

google/cloud/aiplatform/initializer.py
```diff
@@ -149,6 +149,35 @@ def project(self) -> str:
...
+
+                cloud_resource_manager_service = discovery.build(
+                    "cloudresourcemanager", "v3"
+                )
+                project_id = (
+                    cloud_resource_manager_service.projects()
+                    .get(name=f"projects/{project_number}")
+                    .execute()["projectId"]
+                )
+                self._project = project_id
```
sasha-gitg, line 172: I'm a bit hesitant to assign to `self._project` because it's inconsistent with how we handle retrieving the default project below. I understand this is useful because it will remove the additional API calls when accessing the property in the future. Perhaps if this is broken out into a separate utility we can use `functools.cache` or `functoosl.lru_cache` to capture the mapping of project number to project ID.

sasha-gitg (changes requested):

morgandu:
Hi @Ark-kun , PTAL of the #1068 for extracting project ID given a project number.

google/cloud/aiplatform/initializer.py
Ark-kun: Hmm. How would you do it?
Are you proposing something like this?
```python
try:
    from googleapiclient import discovery
except ImportError:
    discovery = None
if discovery is not None:
    ...
```

Ark-kun: Done

google/cloud/aiplatform/initializer.py
Ark-kun: Done.

google/cloud/aiplatform/initializer.py
Ark-kun: Done.

Ark-kun:
I've switched to the new utility function.

Ark-kun:
>Presubmit - Unit Tests Expected — Waiting for status to be reported

Is there some issues with the tests? Most tests are not running.

Ark-kun:
> Please add a unit test that tests this block. Preferably, also an integration test that executes a custom job with the SDK.

I've added both a unit test and an integration test that executes a custom job with the SDK.

sasha-gitg (approved):
Thanks Alexey!

PR #960, opened 2022-01-19
feat: Added LocalModel for creating models with existing images and uploading models.

abcdefgs0324:

It's for custom prediction routine.


google/cloud/aiplatform/prediction/local_model.py
```diff
@@ -0,0 +1,291 @@
...
+
+        Returns:
+            model: Instantiated representation of the uploaded model resource.
+
+        Raises:
+            ValueError: If only `explanation_metadata` or `explanation_parameters`
+                is specified.
+                Also if model directory does not contain a supported model file.
+        """
+        envs = {}
```
sasha-gitg, line 260: It seems like a dict comphension and a list comprehension below would be more Pythonic.

sasha-gitg (approved):

google/cloud/aiplatform/prediction/local_model.py
```diff
@@ -0,0 +1,285 @@
...
+        cls,
+        serving_container_image_uri: str,
+        serving_container_predict_route: Optional[str] = None,
+        serving_container_health_route: Optional[str] = None,
+        serving_container_command: Optional[Sequence[str]] = None,
+        serving_container_args: Optional[Sequence[str]] = None,
+        serving_container_environment_variables: Optional[Dict[str, str]] = None,
+        serving_container_ports: Optional[Sequence[int]] = None,
+    ) -> "LocalModel":
+        """Creates a local model from a built image and given container spec.
```
kthytang, line 57: nit - "... from an existing image..."

kthytang (approved):
LGTM! Great work!

google/cloud/aiplatform/prediction/local_model.py
```diff
@@ -0,0 +1,285 @@
...
+from google.cloud.aiplatform import explain
+from google.cloud.aiplatform import models
+
+from google.cloud.aiplatform.compat.types import (
+    model as gca_model_compat,
+    env_var as gca_env_var_compat,
+)
+
+
+class LocalModel:
```
Ark-kun, line 32: Is "local model" the best term here?
Usually, when people talk about a local model, they mean a local model file (XGBoost, Scikit-learn, Tensorflow).
Is this class related to local model files?

abcdefgs0324: It aims to be analogous to the local version of Vertex model resource. It will have artifacts while we deploy is to a local endpoint.

Ark-kun: How would the CUJ for "deploy local model file to local endpoint" scenario look like for the user?

Currently we support a very simple and user-friendly interface for Vertex Models:

```python
model = aiplatform.Model.upload_tensorflow_saved_model("iris.tensorflow_model.SavedModel")
```

abcdefgs0324: You could see this PR https://github.com/googleapis/python-aiplatform/pull/1005.

An example workflow should be like.
```
# Builds a custom container.
local_model = LocalModel.create_cpr_model(
    "path/to/user_code",
    "us-docker.pkg.dev/my-project/my-image:0223",
    predictor=CustomPredictor,
    requirements_path=os.path.join(_USER_CODE_DIR, _REQUIREMENTS_FILE),
)
# Locally tests the container.
with local_model.deploy_to_local_endpoint(
    artifact_uri=_ARTIFACT_URI,
    credential_path=os.environ.get("GOOGLE_APPLICATION_CREDENTIALS"),
) as local_endpoint:
    local_predict_response = local_endpoint.predict(
        request=f'{{"instances": [[1, 2, 3, 4]]}}',
        headers={"Content-Type": "application/json"},
    )
# Pushes the built image, uploads, and deploys.
local_model.push_image()
model = local_model.upload("my_custom_model", artifact_uri=_ARTIFACT_URI)
model.deploy(machine_type="n1-standard-2")
```

The main difference is that in this workflow, users are creating a new image and deploy with that built image. But in the sample you mentioned, users are reusing our pre-built images.

PR #963, opened 2022-01-20
feat: Open LIT with a deployed model

taiseiak:
Add the ability to open LIT with a deployed model. Has support for explainability, if the model has it enabled.

b/210943929


google/cloud/aiplatform/explain/lit.py
```diff
@@ -75,8 +76,103 @@ def spec(self):
...
+            if not deployed_model:
+                raise ValueError(
+                    "A model with id {model_id} was not found in the endpoint {endpoint}.".format(
+                        model_id=model_id, endpoint=model
+                    )
+                )
+            self._explanation_enabled = bool(deployed_model.explanation_spec)
+        # Check if all models in the endpoint have explanation enabled
+        else:
+            self._explanation_enabled = all(
```
sasha-gitg, line 131: Prefer the comprehension `all(model.explanation_spec for model in self._endpoint.list_models())` as it's more Pythonic. Also, I don't think you need to cast to bool, it should automatically evaluate to True if the field if present and False if not.

google/cloud/aiplatform/explain/lit.py
```diff
@@ -75,8 +76,103 @@ def spec(self):
...
+                Required. An OrderedDict of string names matching the labels of the model
+                as the key, and the associated LitType of the label.
+              model_id:
+                Optional. A string of the specific model in the endpoint to create the
+                LIT model from. If this is not set, any usable model in the endpoint is
+                used to create the LIT model.
+            Raises:
+                ValueError if the model_id was not found in the endpoint.
+        """
+        self._endpoint = aiplatform.Endpoint(model)
```
sasha-gitg, line 110: This will only use the default environment creds or creds set in `aiplatform.init`. Accepting an `Endpoint` object will provide greater flexibility by allowing them to set creds at construction time. Is there a reason we wouldn't want to do that?

google/cloud/aiplatform/explain/lit.py
```diff
@@ -75,8 +76,103 @@ def spec(self):
...
+        else:
+            self._explanation_enabled = all(
+                map(
+                    lambda model: bool(model.explanation_spec),
+                    self._endpoint.list_models(),
+                )
+            )
+
+    def predict_minibatch(
+        self, inputs: List[lit_types.JsonDict]
```
sasha-gitg, line 139: docstring

google/cloud/aiplatform/explain/lit.py
```diff
@@ -244,12 +340,17 @@ def create_lit_model(
     input_types: "OrderedDict[str, lit_types.LitType]",  # noqa: F821
     output_types: "OrderedDict[str, lit_types.LitType]",  # noqa: F821
     attribution_method: str = "sampled_shapley",
+    model_id: Optional[str] = None,
```
sasha-gitg, line 343: This function may be getting a overloaded with `model` argument representing both model and endpoints. Recommend breaking out the endpoint support to a new function: `create_lit_model_from_endpoint`.

google/cloud/aiplatform/explain/lit.py
```diff
@@ -260,10 +361,17 @@ def create_lit_model(
...
             or 'integrated_gradients'.
+          model_id:
+            Optional. A string of the specific model in the endpoint to create the
+            LIT model from. If this is not set, any usable model in the endpoint is
+            used to create the LIT model.
         Returns:
             A LIT Model object that has the same functionality as the model provided.
     """
-    return _VertexLitModel(model, input_types, output_types, attribution_method)
+    if os.path.exists(model):
```
sasha-gitg, line 371: Endpoint resource name would be a valid path. Unlikely it would exist, but this provides greater motivation to accept `aiplatform.Endpoint` so we can rely on type checking. Relating to the comment above, separating into a new function would remove the requirement to check.

sasha-gitg (changes requested):

google/cloud/aiplatform/explain/lit.py
```diff
@@ -75,8 +76,103 @@ def spec(self):
...
+    def __init__(
+        self,
+        model: str,
+        input_types: "OrderedDict[str, lit_types.LitType]",  # noqa: F821
+        output_types: "OrderedDict[str, lit_types.LitType]",  # noqa: F821
+        model_id: Optional[str] = None,
+    ):
+        """Construct a VertexLitModel.
+            Args:
+              model:
```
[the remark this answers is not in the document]
taiseiak: If I change this to another function, would it be still called `endpoint_name` if the type was `Union[str, aiplatform.Endpoint]`?

google/cloud/aiplatform/explain/lit.py
taiseiak: Looks much better. Done.

google/cloud/aiplatform/explain/lit.py
taiseiak: We though that having one function that can handle different situations might have been better design, as it lowered the overhead from the user. However, I am now refactoring into another function.

google/cloud/aiplatform/explain/lit.py
taiseiak: Added.

google/cloud/aiplatform/explain/lit.py
taiseiak: Separating to another function.

google/cloud/aiplatform/explain/lit.py
taiseiak: Done.

google/cloud/aiplatform/explain/lit.py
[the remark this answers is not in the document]
sasha-gitg: LGTM on `endpoint`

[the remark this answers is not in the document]
sasha-gitg: Please update the docstring.

google/cloud/aiplatform/explain/lit.py
```diff
@@ -239,6 +347,33 @@ def create_lit_dataset(
...
+def create_lit_model_from_endpoint(
+    endpoint: Union[str, aiplatform.Endpoint],
+    input_types: "OrderedDict[str, lit_types.LitType]",  # noqa: F821
+    output_types: "OrderedDict[str, lit_types.LitType]",  # noqa: F821
+    model_id: Optional[str] = None,
+) -> lit_model.Model:
+    """Creates a LIT Model object.
+        Args:
+          model:
+            Required. The name of the Endpoint resource or an Endpoint class.
```
sasha-gitg, line 359:
```suggestion
            Required. The name of the Endpoint resource or an Endpoint instance.
```

google/cloud/aiplatform/explain/lit.py
```diff
@@ -75,8 +76,109 @@ def spec(self):
...
+        else:
+            self._explanation_enabled = all(
+                model.explanation_spec for model in self._endpoint.list_models()
+            )
+
+    def predict_minibatch(
+        self, inputs: List[lit_types.JsonDict]
+    ) -> List[lit_types.JsonDict]:
+        """Retun predictions based on a batch of inputs.
+            Args:
```
sasha-gitg, line 142: It looks like the Args/Returns section need to be deindented one level throughout.

sasha-gitg (approved):

PR #968, opened 2022-01-24
Remove logging.basicConfig from module level

### LINKED ISSUE #967: importing should not override logging config
Merely importing the client libraries should not override the logging configuration that has been set up.

Sample code:

```python
import logging

logger = logging.getLogger('test_logger')


def main():
    logger.debug('test debug level')
    logger.info('test info level')


if __name__ == "__main__":
    LOG_FORMAT = '%(asctime)s %(name)-20s %(levelname)-8s %(funcName)s: %(message)s'
    logging.basicConfig(level=logging.DEBUG, format=LOG_FORMAT)

    main()
```

The output from this simple program is:

```
INFO:test_logger:test info level
```

If you comment out the `aiplatform` import  the expected logs are produced

```
2022-01-24 17:52:47,678 test_logger          DEBUG    main: test debug level
2022-01-24 17:52:47,678 test_logger          INFO     main: test info level
```

crazystick:
It should be up to users of the library to configure logging how they wish and we should not override it.

Fixes #967


samgoodman:
Addressed by moving aiplatform logging to its own logger (and not the root logger) in https://github.com/googleapis/python-aiplatform/pull/1081. Thanks for the call out, @crazystick!

PR #976, opened 2022-01-25
fix: show logs when TFX pipelines are submitted

sararob:
Fixes an issue where calling `PipelineJob.submit()` with TFX pipelines didn't show any logs. This was happening because when `tfx.v1` is imported, the log level gets set to `WARNING`. I added a check for TFX pipelines and reset the log level to `INFO`.

Fixes b/209602615.

### DIFF
--- a/google/cloud/aiplatform/pipeline_jobs.py
+++ b/google/cloud/aiplatform/pipeline_jobs.py
@@ -16,6 +16,7 @@
 #
 
 import datetime
+import logging
 import time
 import re
 from typing import Any, Dict, List, Optional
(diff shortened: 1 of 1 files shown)


google/cloud/aiplatform/pipeline_jobs.py
```diff
@@ -273,6 +274,10 @@ def submit(
         if network:
             self._gca_resource.network = network
 
+        # Prevents logs from being supressed on TFX pipelines
+        if self._gca_resource.pipeline_spec["sdkVersion"].startswith("tfx"):
```
sasha-gitg, line 278: Please update the unit tests to include a mock tfx pipeline similar to this:
https://github.com/googleapis/python-aiplatform/blob/main/tests/unit/aiplatform/test_pipeline_jobs.py#L76

It can then be added to parameterize the current tests here:
https://github.com/googleapis/python-aiplatform/blob/main/tests/unit/aiplatform/test_pipeline_jobs.py#L240

You may need to update the expected call in the test to add `sdkVersion`: https://github.com/googleapis/python-aiplatform/blob/main/tests/unit/aiplatform/test_pipeline_jobs.py#L285

sasha-gitg (changes requested):

sasha-gitg (approved):

ji-yaqi (approved):

PR #977, opened 2022-01-25
feat: enable ingest from pd.DataFrame

morgandu:
- [x] added ingest_from_df
- [x] added unit tests
- [x] added system tests

### DIFF
--- a/google/cloud/aiplatform/featurestore/entity_type.py
+++ b/google/cloud/aiplatform/featurestore/entity_type.py
@@ -17,6 +17,7 @@
 
 import datetime
 from typing import Dict, List, Optional, Sequence, Tuple, Union
+import uuid
 
 from google.auth import credentials as auth_credentials
 from google.protobuf import field_mask_pb2
@@ -34,6 +35,7 @@ from google.cloud.aiplatform import initializer
 from google.cloud.aiplatform import utils
 from google.cloud.aiplatform.utils import featurestore_utils
 
+from google.cloud import bigquery
 
 _LOGGER = base.Logger(__name__)
 _ALL_FEATURE_IDS = "*"
@@ -1000,6 +1002,7 @@ class EntityType(base.VertexAiResourceNounWithFutureManager):
             EntityType - The entityType resource object with feature values imported.
 
         """
+
         bigquery_source = gca_io.BigQuerySource(input_uri=bq_source_uri)
 
         import_feature_values_request = self._validate_and_get_import_feature_values_request(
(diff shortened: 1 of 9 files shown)


google/cloud/aiplatform/featurestore/entity_type.py
```diff
@@ -1146,6 +1149,148 @@ def ingest_from_gcs(
...
+            :1024
+        ]
+        temp_bq_table_id = f"{temp_bq_dataset_id}.{entity_type_id}"
+
+        temp_bq_dataset = bigquery.Dataset(dataset_ref=temp_bq_dataset_id)
+        temp_bq_dataset.location = self.location
+        temp_bq_dataset = bigquery_client.create_dataset(temp_bq_dataset)
+
+        try:
+            job = bigquery_client.load_table_from_dataframe(
```
ivanmkc, line 1271: Doesn't a temp table and id get generated for you if you don't provide one? What's wrong with using that?

google/cloud/aiplatform/featurestore/entity_type.py
```diff
@@ -1146,6 +1149,148 @@ def ingest_from_gcs(
...
+        entity_id_field: Optional[str] = None,
+        disable_online_serving: Optional[bool] = None,
+        worker_count: Optional[int] = None,
+        request_metadata: Optional[Sequence[Tuple[str, str]]] = (),
+    ) -> "EntityType":
+        """Ingest feature values from DataFrame.
+
+        Args:
+            feature_ids (List[str]):
+                Required. IDs of the Feature to import values
```
diemtvu, line 1167: nit. Can this be optional? If not set, all columns (except the ID and timestamp column) will be imported

google/cloud/aiplatform/featurestore/entity_type.py
```diff
@@ -1146,6 +1149,148 @@ def ingest_from_gcs(
...
+                If not provided, the source column need to be the same as the Feature ID.
+
+                Example:
+
+                     feature_ids = ['my_feature_id_1', 'my_feature_id_2', 'my_feature_id_3']
+
+                     In case all features' source field and ID match:
+                     feature_source_fields = None or {}
+
+                     In case all features' source field and ID do not match:
```
diemtvu, line 1204: This is not much different from the purpose of this field, but the 2 examples (this and below) give me the impression that  there are some differentiate at first. I think it's cleaner to have the second example only.

google/cloud/aiplatform/featurestore/entity_type.py
```diff
@@ -1146,6 +1149,148 @@ def ingest_from_gcs(
...
+                        'my_feature_id_2': 'my_feature_id_2_source_field',
+                        'my_feature_id_3': 'my_feature_id_3_source_field',
+                     }
+
+                     In case some features' source field and ID do not match:
+                     feature_source_fields = {
+                        'my_feature_id_1': 'my_feature_id_1_source_field',
+                     }
+            entity_id_field (str):
+                Optional. Source column that holds entity IDs. If not provided, entity
```
diemtvu, line 1216: nit. Does this make sense to have the default value to `entity_id` and drop the optional?

google/cloud/aiplatform/featurestore/entity_type.py
```diff
@@ -1146,6 +1149,148 @@ def ingest_from_gcs(
...
         )
 
+    def ingest_from_df(
+        self,
+        feature_ids: List[str],
+        feature_time: Union[str, datetime.datetime],
+        df_source: "pd.DataFrame",  # noqa: F821 - skip check for undefined name 'pd'
+        feature_source_fields: Optional[Dict[str, str]] = None,
+        entity_id_field: Optional[str] = None,
+        disable_online_serving: Optional[bool] = None,
```
diemtvu, line 1159: Given ingest data from dataframe should not be too big (less than few Gb in most cases, I am thinking we probably want to hide these 2 options (disable_online_serving and worker_count) for simplicity. It also easier to change to use streaming ingestion in the future (which is faster and simpler). Wdyt?

google/cloud/aiplatform/featurestore/entity_type.py
morgandu: It does, updated.

google/cloud/aiplatform/featurestore/entity_type.py
morgandu: Making it optional will lead to breaking changes, assuming we want ingest_from_df, ingest_from_bq, and ingest_from_df has same pattern.

One option is to update `feature_ids: List[str]` to `feature_ids: Union[List[str], str]` and takes `"*"` as value to import all features. 

Though, I think enforcing users to put together a list of `feature_ids` maybe helpful to differentiate created features in the entity_type, versus extra columns in the data source. 

For example, users have some columns does not match to any existing feature_id, and they expect them to be imported. By passing "*" into the service, users will only find out that they need to create the features first after the ingestion. But if they pass in individual feature_id, service will not proceed and throw back features not exist.

google/cloud/aiplatform/featurestore/entity_type.py
morgandu: updated

google/cloud/aiplatform/featurestore/entity_type.py
morgandu: Optional->Required will also lead to breaking change for `ingest_from_bq` and `ingest_from_gcs`.

Added default value as `"entity_id"`, but still Optional, since service can take None and use "entity_id" by default.

google/cloud/aiplatform/featurestore/entity_type.py
morgandu: It maybe easier to keep the optional field in case user want to use it (unless these for sure will not be needed?) than have to add it back later. Also, we have different methods for online serve and batch serve, it may make more sense to have both online ingest and batch ingest. Wdyt?

google/cloud/aiplatform/featurestore/entity_type.py
diemtvu: SG.

google/cloud/aiplatform/featurestore/entity_type.py
diemtvu: I'm not sure if we will provide dual implementation (using batch vs streaming ingestion API). I'm leaning toward just using streaming ingestion as it's both simpler to implement and faster for notebook use-case. With that API, disable_online_serving and number of workers is not relevant. I feel it would be more complex to deprecate it than add it back when really need it.

I vote for leave it out for now, and we can revisit when streaming ingestion is ready for this function.

google/cloud/aiplatform/featurestore/entity_type.py
```diff
@@ -1146,6 +1128,140 @@ def ingest_from_gcs(
...
+                performance.
+            request_metadata (Sequence[Tuple[str, str]]):
+                Optional. Strings which should be sent along with the request as metadata.
+
+        Returns:
+            EntityType - The entityType resource object with feature values imported.
+
+        """
+        try:
+            import pyarrow  # noqa: F401 - skip check for 'pyarrow' which is required when using 'google.cloud.bigquery'
```
sasha-gitg, line 1213: Feature store should have it's own `extras_require` in `setup.py` that includes these dependencies. and these dependencies should be added to `full`.

tests/system/aiplatform/test_featurestore.py
```diff
@@ -217,21 +213,107 @@ def test_batch_create_features_and_ingest_feature_values(
...
+            feature_time=feature_time_column,
+            df_source=movies_df,
             entity_id_field="movie_id",
-            worker_count=2,
+            worker_count=1,
         )
 
-        list_movie_features = movie_entity_type.list_features()
-        assert len(list_movie_features) == 3
+        assert "EntityType feature values imported." in caplog.text
```
sasha-gitg, line 266: Is there anyway we can assert the resource has been updated based on the ingest from df?

google/cloud/aiplatform/featurestore/entity_type.py
```diff
@@ -1146,6 +1149,148 @@ def ingest_from_gcs(
...
+        entity_type_name_components = self._parse_resource_name(self.resource_name)
+        featurestore_id, entity_type_id = (
+            entity_type_name_components["featurestore"],
+            entity_type_name_components["entity_type"],
+        )
+
+        temp_bq_dataset_name = f"temp_{featurestore_id}_{uuid.uuid4()}".replace(
+            "-", "_"
+        )
+        temp_bq_dataset_id = f"{initializer.global_config.project}.{temp_bq_dataset_name}"[
```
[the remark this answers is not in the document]
morgandu: `self.project` returns the project number, but bq dataset need a project ID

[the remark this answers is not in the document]
sasha-gitg: Got it. This is an issue moving forward because the project used to construct `EntityType` might not match the project ID set in `init` or, if project is not set using `init`, the project ID returned from `google.auth.default`.

Additionally, we don't stop the user from using a project number when setting `init`.

We shouldn't block this PR as there is a workaround but we should capture this issue in a ticket.

google/cloud/aiplatform/featurestore/entity_type.py
morgandu: added

tests/system/aiplatform/test_featurestore.py
morgandu: updated with the `read` method which produce a pd from online reading, to validate the feature values before and after ingestion, using a raw data created dataframe. This in turn validated both `read` and `ingest_from_df`method.

google/cloud/aiplatform/featurestore/entity_type.py
morgandu: Sounds good. Removed the two args.

google/cloud/aiplatform/featurestore/entity_type.py
morgandu: The `destination` requires a `TableReference` or a string for the table, and it requires an existing `dataset`.

google/cloud/aiplatform/featurestore/entity_type.py
```diff
@@ -1146,6 +1149,148 @@ def ingest_from_gcs(
...
+        feature_ids: List[str],
+        feature_time: Union[str, datetime.datetime],
+        df_source: "pd.DataFrame",  # noqa: F821 - skip check for undefined name 'pd'
+        feature_source_fields: Optional[Dict[str, str]] = None,
+        entity_id_field: Optional[str] = None,
+        disable_online_serving: Optional[bool] = None,
+        worker_count: Optional[int] = None,
+        request_metadata: Optional[Sequence[Tuple[str, str]]] = (),
+    ) -> "EntityType":
+        """Ingest feature values from DataFrame.
```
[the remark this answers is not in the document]
morgandu: Added a `Note:`

google/cloud/aiplatform/featurestore/entity_type.py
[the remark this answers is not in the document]
morgandu: sounds good, will address this in a separate ticket/PR

google/cloud/aiplatform/featurestore/entity_type.py
```diff
@@ -1146,6 +1128,129 @@ def ingest_from_gcs(
...
 
+    def ingest_from_df(
+        self,
+        feature_ids: List[str],
+        feature_time: Union[str, datetime.datetime],
+        df_source: "pd.DataFrame",  # noqa: F821 - skip check for undefined name 'pd'
+        feature_source_fields: Optional[Dict[str, str]] = None,
+        entity_id_field: Optional[str] = None,
+        request_metadata: Optional[Sequence[Tuple[str, str]]] = (),
+    ) -> "EntityType":
```
diemtvu, line 1139: Can you add a comment to mention whether this function wait for ingestion complete (in other words, do users expected    data is available once the function return)

diemtvu (commented):
lgtm. Thanks

google/cloud/aiplatform/featurestore/entity_type.py
morgandu: added

sasha-gitg (approved):

PR #981, opened 2022-01-27
docs(samples): replace deprecated fields in create_training_pipeline_tabular_forecasting_sample.py

dwkk-google:
They have been deprecated since the Public preview since last May.


ivanmkc:
LGTM but can you provide context in the PR description? Have all of those parameters been renamed? Since what version?

dwkk-google:
> LGTM but can you provide context in the PR description? Have all of those parameters been renamed? Since what version?

It is like this since the Public preview since last May. we have deprecate those fields back then.

ivanmkc (approved):
LGTM

PR #982, opened 2022-01-28
feat: Initialized predictor, serializer, and model server in the 

abcdefgs0324:
prediction folder for custom prediction routine and added extra
requirements for prediction.

### DIFF
--- a/google/cloud/aiplatform/prediction/handler.py
+++ b/google/cloud/aiplatform/prediction/handler.py
@@ -15,6 +15,8 @@
 # limitations under the License.
 #
 
+from typing import Optional, Type
+
 try:
     from fastapi import Request
     from fastapi import Response
@@ -31,12 +33,17 @@ from google.cloud.aiplatform.prediction.serializer import DefaultSerializer
 class Handler:
     """Interface for Handler class to handle prediction requests."""
 
-    def __init__(self, predictor: Predictor):
+    def __init__(
+        self, gcs_artifacts_uri: str, predictor: Optional[Type[Predictor]] = None,
+    ):
         """Initializes a Handler instance.
 
         Args:
-            predictor (Predictor):
-                Required. The Predictor instance this handler consumes.
+            gcs_artifacts_uri (str):
+                Required. The value of the environment variable AIP_STORAGE_URI.
+            predictor (Type[Predictor]):
+                Optional. The Predictor class this handler uses to initiate predictor
+                instance if given.
         """
         pass
 
@@ -54,16 +61,27 @@ class Handler:
 
 
 class PredictionHandler(Handler):
-    """Default handler for handling the requests sent to the application."""
+    """Default prediction handler for the prediction requests sent to the application."""
 
-    def __init__(self, predictor: Predictor):
-        """Initializes a PredictionHandler instance.
+    def __init__(
+        self, gcs_artifacts_uri: str, predictor: Optional[Type[Predictor]] = None,
+    ):
+        """Initializes a Handler instance.
 
         Args:
-            predictor (Predictor):
-                Required. The Predictor instance this handler consumes.
+            gcs_artifacts_uri (str):
+                Required. The value of the environment variable AIP_STORAGE_URI.
+            predictor (Type[Predictor]):
+                Optional. The Predictor class this handler uses to initiate predictor
+                instance if given.
         """
-        self._predictor = predictor
+        if predictor is None:
+            raise ValueError(
+                "PredictionHandler must have a predictor class passed to the init function."
+            )
+
+        self._predictor = predictor()
+        self._predictor.load(gcs_artifacts_uri)
 
     async def handle(self, request: Request) -> Response:
         """Handles a prediction request.
--- a/google/cloud/aiplatform/prediction/handler.py
+++ b/google/cloud/aiplatform/prediction/handler.py
@@ -26,8 +26,8 @@ except ImportError:
         'Please install the SDK using "pip install python-aiplatform[prediction]"'
     )
 
+from google.cloud.aiplatform.prediction import handler_utils
 from google.cloud.aiplatform.prediction.predictor import Predictor
-from google.cloud.aiplatform.prediction.serializer import APPLICATOIN_JSON
 from google.cloud.aiplatform.prediction.serializer import DefaultSerializer
 
 
@@ -95,13 +95,13 @@ class PredictionHandler(Handler):
             The response of the prediction request.
         """
         request_body = await request.body()
-        prediction_input = DefaultSerializer.deserialize(
-            request_body, request.headers.get("content-type", APPLICATOIN_JSON)
-        )
+        content_type = handler_utils.get_content_type_from_headers(request.headers)
+        prediction_input = DefaultSerializer.deserialize(request_body, content_type)
+
         prediction_results = self._predictor.postprocess(
             self._predictor.predict(self._predictor.preprocess(prediction_input))
         )
-        data = DefaultSerializer.serialize(
-            prediction_results, request.headers.get("accept", APPLICATOIN_JSON)
-        )
-        return Response(content=data, media_type=request.headers.get("accept"))
+
+        accept = handler_utils.get_content_type_from_headers(request.headers)
+        data = DefaultSerializer.serialize(prediction_results, accept)
+        return Response(content=data, media_type=accept)
--- a/google/cloud/aiplatform/prediction/handler.py
+++ b/google/cloud/aiplatform/prediction/handler.py
@@ -53,11 +53,11 @@ class Handler:
         pass
 
 
-class DefaultHandler(Handler):
+class PredictionHandler(Handler):
     """Default handler for handling the requests sent to the application."""
 
     def __init__(self, predictor: Predictor):
-        """Initializes a DefaultHandler instance.
+        """Initializes a PredictionHandler instance.
 
         Args:
             predictor (Predictor):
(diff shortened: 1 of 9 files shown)


google/cloud/aiplatform/prediction/predictor.py
```diff
@@ -0,0 +1,70 @@
...
+from typing import Any
+
+
+class Predictor:
+    """Interface for Predictor class that users would be implementing."""
+
+    def __init__(self):
+        pass
+
+    def load(self, gcs_model_uri: str):
```
kthytang, line 27: Let's call this gcs_artifacts_uri.

google/cloud/aiplatform/prediction/handler.py
```diff
@@ -0,0 +1,88 @@
...
+    )
+
+from google.cloud.aiplatform.prediction.predictor import Predictor
+from google.cloud.aiplatform.prediction.serializer import DefaultSerializer
+
+
+class Handler:
+    """Interface for Handler class to handle prediction requests."""
+
+    def __init__(self, predictor: Predictor):
```
kthytang, line 34: Let's actually not require the user provide a Predictor (they may not necessarily use the `Predictor` interface.

We'll need to pass the gcs_artifacts_uri to the handler though. So perhaps have a `@classmethod`, `from_gcs_artifacts_uri` that we can call to instantiate the user's `Handler`? WDYT?

google/cloud/aiplatform/prediction/handler.py
```diff
@@ -0,0 +1,88 @@
...
+                Required. The Predictor instance this handler consumes.
+        """
+        pass
+
+    def handle(self, request: Request) -> Response:
+        """Handles a prediction request.
+
+        Args:
+            request (Request):
+                The request sent to the application.
```
kthytang, line 48: nit: "prediction request"

google/cloud/aiplatform/prediction/model_server.py
```diff
@@ -0,0 +1,105 @@
...
+from google.cloud.aiplatform.prediction.handler import DefaultHandler
+from google.cloud.aiplatform.prediction.handler import Handler
+from google.cloud.aiplatform.prediction.predictor import Predictor
+
+
+class ModelServer:
+    """Base model server to do custom prediction routines."""
+
+    def __init__(
+        self, predictor: Predictor, handler_class: Type[Handler] = DefaultHandler
```
kthytang, line 48: This will only require a `Handler`, and the `ModelServer` will only have a concept of the `Handler`. 

Under the hood, the `Handler` may use a `Predictor`, but we won't require it. The `Handler` will be responsible for interacting with the `Predictor`.

kthytang (commented):
Awesome work!

google/cloud/aiplatform/prediction/handler.py
abcdefgs0324: I'm a bit confused. I think `Predictor` loads a model instead of `Handler`, doesn't it?

Or you mean a `@class method` `from_gcs_artifacts_uri` in `Predictor` as the following code?
```
class Handler:
    def __init__(self, gcs_artifacts_uri):
        self.predictor = CustomPredictor.from_gcs_artifacts_uri(gcs_artifacts_uri)
```
There's another problem here that we need to `import CustomPredictor` before users implement theirs. The current way to pass `predictor` to `Handler` avoids this issue.

We could consider marking `predictor: Optional[Predictor] = None` in `Handler.__init__` but `PredictionHandler` must have `predictor` passed in our design. Therefore, I didn't have strong preference to `Optional`.

google/cloud/aiplatform/prediction/model_server.py
abcdefgs0324: We can let `ModelServer` to have `Handler` only. Then we will put `handler` initialization before we pass it to `ModelServer` (the time that `ModelServer` initialization is in entrypoint.py now). And the time we call `Predictor.load` will be in `handler` initialization.

google/cloud/aiplatform/prediction/handler.py
abcdefgs0324: Updated the code. PTAL.

google/cloud/aiplatform/prediction/model_server.py
abcdefgs0324: Updated the code. PTAL.

google/cloud/aiplatform/prediction/predictor.py
```diff
@@ -0,0 +1,70 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+from typing import Any
+
+
+class Predictor:
```
samthrasher, line 21: If we do expect users to inherit from this class, we should raise NotImplemented instead of returning None in the methods we expect users to implement (i.e. predict, init, load).

google/cloud/aiplatform/prediction/serializer.py
```diff
@@ -0,0 +1,95 @@
...
+        "FastAPI is not installed and is required to build model servers. "
+        'Please install the SDK using "pip install python-aiplatform[prediction]"'
+    )
+
+
+class Serializer:
+    """Interface to implement serialization and deserialization for prediction."""
+
+    @staticmethod
+    def deserialize(data: Any, content_type: Optional[str],) -> Any:
```
samthrasher, line 34: stray comma here and in other Serializer and DefaultSerializer methods

google/cloud/aiplatform/prediction/serializer.py
```diff
@@ -0,0 +1,95 @@
...
+        Args:
+            data (Any):
+                Required. The request data sent to the application.
+            content_type (Optional[str]):
+                Optional. The specified content type of the request.
+        """
+        pass
+
+    @staticmethod
+    def serialize(prediction: Dict, accept: Optional[str],) -> Any:
```
samthrasher, line 46: Is it always a dict? I don't think this will be true in general.

samthrasher (changes requested):
Nice work, left a few comments/questions to address before submitting.

google/cloud/aiplatform/prediction/handler.py
```diff
@@ -0,0 +1,106 @@
...
+            The response of the prediction request.
+        """
+        pass
+
+
+class PredictionHandler(Handler):
+    """Default prediction handler for the prediction requests sent to the application."""
+
+    def __init__(
+        self, gcs_artifacts_uri: str, predictor: Optional[Type[Predictor]] = None,
```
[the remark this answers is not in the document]
abcdefgs0324: Made it a protocol and also made the function default to not implemented error.

google/cloud/aiplatform/prediction/model_server.py
```diff
@@ -0,0 +1,95 @@
...
+    def __init__(self, handler: Handler):
+        """Initializes a fastapi application and sets the configs.
+
+        Args:
+            handler (Handler):
+                Required. The handler to handle requests.
+        """
+        self.handler = handler
+
+        if "AIP_HTTP_PORT" not in os.environ:
```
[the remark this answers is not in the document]
abcdefgs0324: If we add the default values, users can run it locally without providing any values but they can not deploy on our service without values. Therefore, I prefer raising errors here.

With SDK integration, we will provide default values in our LocalModel class. While users use the class to upload models, the default routes will also be used.

tests/unit/aiplatform/test_prediction.py
```diff
@@ -0,0 +1,359 @@
...
+        model_server = ModelServer(handler)
+
+        client = TestClient(model_server.app)
+
+        with mock.patch.object(model_server.handler, "handle") as handle_mock:
+            handle_mock.side_effect = HTTPException(status_code=400)
+
+            response = client.post(_TEST_AIP_PREDICT_ROUTE, json={"x": [1]})
+
+        assert response.status_code == 400
```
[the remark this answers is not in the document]
abcdefgs0324: Good catch! Updated to handle the internal server error and added logging.

google/cloud/aiplatform/prediction/handler.py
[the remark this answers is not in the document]
abcdefgs0324: I just found `typing.Protocol` is new in version 3.8+. I removed Protocol from Predictor class and I think it's acceptable we ask users to inherit from our predictor class.

google/cloud/aiplatform/prediction/handler.py
```diff
@@ -0,0 +1,106 @@
...
+        Args:
+            request (Request):
+                Required. The prediction request sent to the application.
+
+        Returns:
+            The response of the prediction request.
+        """
+        request_body = await request.body()
+        prediction_input = DefaultSerializer.deserialize(
+            request_body, request.headers.get("content-type")
```
kthytang, line 98: Can we default this to json if the content-type is not specified?

And same with accept below.

kthytang (changes requested):

google/cloud/aiplatform/prediction/handler.py
abcdefgs0324: If the body is not in json format, then it will throw json deserialize error. Is it expected?

kthytang: I think it's fine. If the user doesn't provide a content-type, let's treat it as json by default.

Can we make the error message clear that the json deserialization failed? And also something like...

To specify a different type, please set the "content-type" header in the request. Currently supported content types: {...}.

abcdefgs0324: Updated as our offline discussion. PTAL.

kthytang (approved):
Thanks for making the changes!

samthrasher (approved):

PR #983, opened 2022-01-28
feat: enable feature store batch serve to Pandas DataFrame; fix: read instances uri for batch serve

morgandu:
- [x] add batch_to_df
- [x] add unit tests
- [x] add integration tests
- [x] fix bugs

### DIFF
--- a/google/cloud/aiplatform/featurestore/featurestore.py
+++ b/google/cloud/aiplatform/featurestore/featurestore.py
@@ -113,8 +113,18 @@ class Featurestore(base.VertexAiResourceNounWithFutureManager):
             featurestore.EntityType - The managed entityType resource object.
         """
         self.wait()
-        featurestore_name_components = self._parse_resource_name(self.resource_name)
+        return self._get_entity_type(entity_type_id=entity_type_id)
 
+    def _get_entity_type(self, entity_type_id: str) -> "featurestore.EntityType":
+        """Retrieves an existing managed entityType in this Featurestore.
+
+        Args:
+            entity_type_id (str):
+                Required. The managed entityType resource ID in this Featurestore.
+        Returns:
+            featurestore.EntityType - The managed entityType resource object.
+        """
+        featurestore_name_components = self._parse_resource_name(self.resource_name)
         return featurestore.EntityType(
             entity_type_name=featurestore.EntityType._format_resource_name(
                 project=featurestore_name_components["project"],
@@ -225,6 +235,7 @@ class Featurestore(base.VertexAiResourceNounWithFutureManager):
         Returns:
             Featurestore - The updated featurestore resource object.
         """
+        self.wait()
         update_mask = list()
 
         if labels:
@@ -314,6 +325,7 @@ class Featurestore(base.VertexAiResourceNounWithFutureManager):
         Returns:
             List[featurestore.EntityType] - A list of managed entityType resource objects.
         """
+        self.wait()
         return featurestore.EntityType.list(
             featurestore_name=self.resource_name, filter=filter, order_by=order_by,
         )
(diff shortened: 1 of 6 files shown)


google/cloud/aiplatform/featurestore/featurestore.py
```diff
@@ -715,59 +749,14 @@ def _validate_and_get_batch_read_feature_values_request(
...
-        return gca_io.CsvSource(gcs_source=gca_io.GcsSource(uris=read_instances))
-
     @base.optional_sync(return_input_arg="self")
     def batch_serve_to_bq(
         self,
         bq_destination_output_uri: str,
         serving_feature_ids: Dict[str, List[str]],
-        feature_destination_fields: Optional[Dict[str, str]] = None,
-        read_instances: Optional[Union[str, List[str]]] = None,
+        read_instances_uri: str,
```
sasha-gitg, line 757: It looks like this is the case but just want to confirm the method hasn't been released yet. Thanks!

google/cloud/aiplatform/featurestore/featurestore.py
```diff
@@ -999,17 +984,204 @@ def batch_serve_to_gcs(
...
+            raise ImportError(
+                f"Pandas is not installed. Please install pandas to use "
+                f"{self.batch_serve_to_df.__name__}"
+            )
+
+        bigquery_client = bigquery.Client(
+            project=self.project, credentials=self.credentials
+        )
+
+        featurestore_name_components = self._parse_resource_name(self.resource_name)
```
sasha-gitg, line 1123: It seems like this method should include a `self.wait` call because it's possible they could run methods like `create` and `create_entity_type` with `sync=False` and then call this method. That may be the case else where in these classes.

google/cloud/aiplatform/featurestore/featurestore.py
```diff
@@ -999,17 +984,204 @@ def batch_serve_to_gcs(
...
+        featurestore_id = featurestore_name_components["featurestore"]
+
+        temp_bq_dataset_name = f"temp_{featurestore_id}_{uuid.uuid4()}".replace(
+            "-", "_"
+        )
+        temp_bq_dataset_id = f"{initializer.global_config.project}.{temp_bq_dataset_name}"[
```
sasha-gitg, line 1129: Please add a TODO with with the project ID buganizer ID and description in locations where we reference the global config's project.

google/cloud/aiplatform/featurestore/featurestore.py
```diff
@@ -999,17 +984,204 @@ def batch_serve_to_gcs(
...
+                dataset=temp_bq_dataset.dataset_id, delete_contents=True,
+            )
+
+        frames = []
+        for stream in read_session_proto.streams:
+            reader = bigquery_storage_read_client.read_rows(stream.name)
+            for message in reader.rows().pages:
+                frames.append(message.to_dataframe())
+
+        return None if not frames else pd.concat(frames)
```
sasha-gitg, line 1187: Should this use `ignore_index=True` or is the SOT for the index in the table or possibly `message.to_dataframe()` assigns incrementing index values?

sasha-gitg, line 1187: This return type doesn't match the signature. I think we should return an empty `pd.DataFrame` if frames is empty.

sasha-gitg (changes requested):

google/cloud/aiplatform/featurestore/featurestore.py
morgandu: Yes, it has not been released.

google/cloud/aiplatform/featurestore/featurestore.py
morgandu: Yeah, that looks right. If so, there are more methods need to include this self.wait() before parsing for resource_name_components.

i.e.
```
def get_entity_type(self, entity_type_id: str) -> "featurestore.EntityType":
        featurestore_name_components = self._parse_resource_name(self.resource_name)
```

google/cloud/aiplatform/featurestore/featurestore.py
morgandu: Added.

google/cloud/aiplatform/featurestore/featurestore.py
morgandu: updated

sasha-gitg (approved):

diemtvu (approved):
lgtm

PR #1022, opened 2022-02-19
feat: Added a system test for prediction cpr.

abcdefgs0324:


tests/system/aiplatform/test_prediction_cpr.py
```diff
@@ -0,0 +1,83 @@
+# -*- coding: utf-8 -*-
+
+# Copyright 2020 Google LLC
```
samthrasher, line 3: 2022?

tests/system/aiplatform/test_resources/cpr_user_code/predictor.py
```diff
@@ -0,0 +1,59 @@
...
+from typing import Any
+
+from google.cloud import storage
+from google.cloud.aiplatform.prediction.predictor import Predictor
+
+
+class SklearnPredictor(Predictor):
+    """Interface for Predictor class that users would be implementing."""
+
+    def __init__(self):
```
samthrasher, line 29: is it necessary to override `__init__` if it has no effect?

samthrasher (approved):
client code looks great!

samthrasher (changes requested):
please fix build before submitting

tests/system/aiplatform/test_resources/cpr_user_code/predictor.py
abcdefgs0324: `__init__` in predictor template is marked as `NotImplementedError` by default.

kthytang (approved):

PR #1033, opened 2022-02-23
feat: Add PrivateEndpoint class and HTTP methods

### LINKED ISSUE #708: Private endpoints are not supported
Vertex AI supports private endpoints.
https://cloud.google.com/vertex-ai/docs/predictions/using-private-endpoints#create-private-endpoint
The user can specify the `network` for the endpoint.

`Endpoint.create()` method does not allow specifying the network.

Add `Endpoint.create(network=...)` parameter.

vinnysenthil:
## Summary of Changes

- Adds `PrivateEndpoint` class which subclasses `Endpoint`
- New create, deploy methods to specify network
- Adds new `predict`, `explain` and `health_check` methods to make unauthenticated REST HTTP calls to model server within network
- Add `network` to `aiplatform.init()` and associated classes, reading the field in `Model.deploy` & `Endpoint.create`
- Add `urllib3` dependency requirement to library
- Add unit tests and system tests for `PrivateEndpoint`

Fixes #708, [b/194725522](http://b/194725522) , [b/211371627](http://b/211371627) (partial), [b/220936610](http://b/220936610), [b/230003208](https://buganizer.corp.google.com/issues/230003208), [b/233758779](http://b/233758779) 🦕


google/cloud/aiplatform/models.py
```diff
@@ -1259,8 +1299,467 @@ def explain(
...
+                credentials set in aiplatform.init.
+        """
+        super().__init__(
+            endpoint_name=endpoint_name,
+            project=project,
+            location=location,
+            credentials=credentials,
+        )
+
+        self._custom_predict_uri = None
```
sasha-gitg, line 1506: Where are these uris used?

google/cloud/aiplatform/models.py
```diff
@@ -1344,67 +1870,164 @@ def list(
...
                 will be executed in concurrent Future and any downstream object will
                 be immediately returned and synced when the Future has completed.
         """
-        self._sync_gca_resource()
+        self._sync_gca_resource_if_skipped()
 
-        models_to_undeploy = sorted(  # Undeploy zero traffic models first
-            self._gca_resource.traffic_split.keys(),
-            key=lambda id: self._gca_resource.traffic_split[id],
+        if len(self._gca_resource.deployed_models):
```
sasha-gitg, line 1969: This usage could raise in the case where the endpoint was created with `sync=False` as the resource may not have been created. The `_sync_gca_resource_if_skipped` applies only to the special case where an already created Endpoint is initialized with it's resource name. We skip the GET call to remove the additional latency of `Endpoint(resource_name).predict(...)`.

Preference to rely on the service to throw.

google/cloud/aiplatform/models.py
```diff
@@ -1344,67 +1870,164 @@ def list(
...
-            sync (bool):
-                Whether to execute this method synchronously. If False, this method
-                will be executed in concurrent Future and any downstream object will
-                be immediately returned and synced when the Future has completed.
-        Raises:
-            FailedPrecondition: If models are deployed on this Endpoint and force = False.
+            deployed_model_id (str):
+                Required. The ID of the DeployedModel to be undeployed from the
+                private Endpoint. Use PrivateEndpoint.list_models() to get the
+                deployed model ID.
```
sasha-gitg, line 2024: It seems like `sync` was removed in the docstring. Hard to tell with this diff.

google/cloud/aiplatform/models.py
```diff
@@ -2068,6 +2691,15 @@ def deploy(
             explanation_parameters,
         )
 
+        if endpoint and endpoint.__class__ == "PrivateEndpoint"():
```
sasha-gitg: I've never encountered this syntax, how does this string literal invocation work: `"PrivateEndpoint"()`

Preference for `isinstance(endpoint, PrivateEndpoint)`

google/cloud/aiplatform/models.py
```diff
@@ -2101,19 +2734,22 @@ def _deploy(
...
+        explanation_metadata: Optional[aiplatform.explain.ExplanationMetadata] = None,
+        explanation_parameters: Optional[
+            aiplatform.explain.ExplanationParameters
+        ] = None,
         metadata: Optional[Sequence[Tuple[str, str]]] = (),
         encryption_spec_key_name: Optional[str] = None,
+        network: Optional[str] = None,
         sync: bool = True,
         deploy_request_timeout: Optional[float] = None,
     ) -> Endpoint:
```
sasha-gitg, line 2747: Union[Endpoint, PrivateEndpoint]

google/cloud/aiplatform/models.py
```diff
@@ -2046,12 +2661,20 @@ def deploy(
...
+                Endpoints [in the documentation](https://cloud.google.com/vertex-ai/docs/predictions/using-private-endpoints)
             sync (bool):
                 Whether to execute this method synchronously. If False, this method
                 will be executed in concurrent Future and any downstream object will
                 be immediately returned and synced when the Future has completed.
             deploy_request_timeout (float):
                 Optional. The timeout for the deploy request in seconds.
+
         Returns:
             endpoint ("Endpoint"):
```
sasha-gitg, line 2680: Union[Endpoint, PrivateEndpoint]

google/cloud/aiplatform/models.py
```diff
@@ -1962,37 +2589,25 @@ def deploy(
...
+        explanation_metadata: Optional[aiplatform.explain.ExplanationMetadata] = None,
+        explanation_parameters: Optional[
+            aiplatform.explain.ExplanationParameters
+        ] = None,
         metadata: Optional[Sequence[Tuple[str, str]]] = (),
         encryption_spec_key_name: Optional[str] = None,
+        network: Optional[str] = None,
         sync=True,
         deploy_request_timeout: Optional[float] = None,
     ) -> Endpoint:
```
sasha-gitg, line 2602: Union[Endpoint, PrivateEndpoint]

google/cloud/aiplatform/initializer.py
```diff
@@ -50,6 +50,7 @@ def __init__(self):
         self._staging_bucket = None
         self._credentials = None
         self._encryption_spec_key_name = None
+        self._network = None
```
sasha-gitg, line 53: Copyright year update

google/cloud/aiplatform/models.py
```diff
@@ -578,18 +596,20 @@ def _validate_deploy_args(
...
-                raise ValueError("Traffic percentage cannot be negative.")
-
-        elif traffic_split:
-            # TODO(b/172678233) verify every referenced deployed model exists
-            if sum(traffic_split.values()) != 100:
-                raise ValueError(
-                    "Sum of all traffic within traffic split needs to be 100."
-                )
+        # TODO(b/221059294): private Endpoints do not yet support traffic splitting
+        if cls == Endpoint:
```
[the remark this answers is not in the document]
sasha-gitg: Preference to refactor this so we don't make the super class need to be aware of the subclass implementation.

sasha-gitg (changes requested):

google/cloud/aiplatform/__init__.py
```diff
@@ -40,6 +40,7 @@
 )
 from google.cloud.aiplatform.metadata import metadata
 from google.cloud.aiplatform.models import Endpoint
+from google.cloud.aiplatform.models import PrivateEndpoint
```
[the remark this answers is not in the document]
nayaknishant: Fixed

google/cloud/aiplatform/models.py
nayaknishant: I think we can remove them.

google/cloud/aiplatform/models.py
```diff
@@ -1259,8 +1299,467 @@ def explain(
...
+                display_name="my_endpoint_name",
+                network="projects/123/global/networks/my_vpc"
+            )
+
+        Args:
+            display_name (str):
+                Required. The user-defined name of the Endpoint.
+                The name can be up to 128 characters long and can be consist
+                of any UTF-8 characters.
+            project (str):
```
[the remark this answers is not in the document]
nayaknishant: Fixed, changed the order of the arguments

google/cloud/aiplatform/models.py
nayaknishant: Fixed

google/cloud/aiplatform/models.py
```diff
@@ -1259,8 +1299,467 @@ def explain(
...
+        display_name: str,
+        network: Optional[str] = None,
+        description: Optional[str] = None,
+        labels: Optional[Dict[str, str]] = None,
+        project: Optional[str] = None,
+        location: Optional[str] = None,
+        credentials: Optional[auth_credentials.Credentials] = None,
+        encryption_spec_key_name: Optional[str] = None,
+        sync=True,
+    ) -> "Endpoint":
```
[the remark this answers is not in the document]
nayaknishant: Fixed

google/cloud/aiplatform/models.py
```diff
@@ -1259,8 +1299,467 @@ def explain(
...
+
+                Overrides encryption_spec_key_name set in aiplatform.init.
+
+            sync (bool):
+                Whether to execute this method synchronously. If False, this method
+                will be executed in concurrent Future and any downstream object will
+                be immediately returned and synced when the Future has completed.
+
+        Returns:
+            endpoint (endpoint.Endpoint):
```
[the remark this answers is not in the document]
nayaknishant: Fixed

google/cloud/aiplatform/models.py
nayaknishant: Fixed

google/cloud/aiplatform/models.py
nayaknishant: Fixed

google/cloud/aiplatform/models.py
nayaknishant: Fixed

google/cloud/aiplatform/models.py
```diff
@@ -1259,8 +1299,467 @@ def explain(
...
+                The name can be up to 128 characters long and can be consist
+                of any UTF-8 characters.
+            project (str):
+                Required. Project to retrieve endpoint from. If not set, project
+                set in aiplatform.init will be used.
+            location (str):
+                Required. Location to retrieve endpoint from. If not set, location
+                set in aiplatform.init will be used.
+            network (str):
+                Required. The full name of the Compute Engine network to which
```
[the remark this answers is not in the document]
nayaknishant: Fixed

google/cloud/aiplatform/models.py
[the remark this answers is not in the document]
nayaknishant: Hey Sasha, sounds good. I'm going to reassign [b/221059294](https://b.corp.google.com/issues/221059294) to myself and will add the arguments to the PrivateEndpoint call and remove the line that checks if the endpoint is private or not.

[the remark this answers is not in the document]
nayaknishant: Also, I don't think we need to include any extra arguments for traffic. In the docstring for `traffic_percentage` it states `Defaults to 100 if there are no pre-existing deployed models.` and private endpoint can only have one model deployed to it as of now. Similarly, `traffic_split` has `should not be provided if traffic_percentage is provided.` I guess we can explicitly put `traffic_split=None`.

google/cloud/aiplatform/models.py
nayaknishant: Fixed

google/cloud/aiplatform/models.py
```diff
@@ -1259,8 +1299,467 @@ def explain(
...
+        body: Optional[Dict[Any, Any]] = None,
+        headers: Optional[Dict[str, str]] = None,
+    ) -> urllib3.response.HTTPResponse:
+        """Helper function used to perform HTTP requests for private Endpoint.
+
+        Args:
+            method (str):
+                Required. The HTTP request method to use. Example: "POST" or "GET"
+            url (str):
+                Required. Project to construct Endpoint object from. If not set,
```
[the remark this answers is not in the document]
nayaknishant: My mistake, fixed

google/cloud/aiplatform/models.py
```diff
@@ -1259,8 +1299,467 @@ def explain(
...
+                project set in aiplatform.init will be used.
+            body (Dict[Any, Any]):
+                Optional. Data sent to the url in the HTTP request. For private
+                Endpoint, an instance is sent and a prediction response is expected.
+            headers (Dict[str, str]):
+                Optional. Header in the HTTP request.
+
+        Returns:
+            urllib3.response.HTTPResponse:
+                An initialized PrivateEndpoint resource.
```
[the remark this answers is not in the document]
nayaknishant: My mistake, fixed

google/cloud/aiplatform/models.py
```diff
@@ -1259,8 +1299,467 @@ def explain(
...
+
+            if response.status < 300:
+                return response
+            else:
+                raise RuntimeError(
+                    f"{response.status} - Failed to make request, see response: "
+                    + response.data.decode("utf-8")
+                )
+
+        except urllib3.exceptions.MaxRetryError:
```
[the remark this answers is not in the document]
nayaknishant: Fixed

google/cloud/aiplatform/models.py
```diff
@@ -1259,8 +1299,467 @@ def explain(
...
+        self._custom_explain_uri = None
+        self._custom_health_uri = None
+
+        self._http_client = urllib3.PoolManager()
+
+    @property
+    def predict_http_uri(self) -> Optional[str]:
+        """Http(s) path to send prediction requests to, used when calling `PrivateEndpoint.predict()`"""
+        if not self._gca_resource.deployed_models:
+            raise RuntimeError(
```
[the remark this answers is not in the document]
nayaknishant: So the `predict`, `explain` and `health_check` methods all use their corresponding uris. I initially though the same way (putting the exception in the method itself), but if a user calls on the uris without deploying a model, the line
```
self._gca_resource.deployed_models[0].private_endpoints.explain_http_uri
```
in the uri methods will return an index out of range error, instead of the RuntimeError. The way it's written now, both the uri methods and the `predict`, `explain` and `health_check` methods will raise the RuntimeError.

Let me know what approach we should take. Thanks

google/cloud/aiplatform/models.py
nayaknishant: Fixed

google/cloud/aiplatform/models.py
nayaknishant: Hey Sasha, I'm a bit confused on what the fix here is, should we delete the line `self._sync_gca_resource_if_skipped()`? Isn't this line in the deploy() method, so how does it relate to predict()? Also what is the service that we have a preference towards?

Thanks,
Nishant

google/cloud/aiplatform/models.py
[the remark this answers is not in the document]
nayaknishant: Sasha, apologizies for not clarifiying. In the docstring for `traffic_percentage` it states: `Defaults to 100 if there are no pre-existing deployed models.`, and for `traffic_split` it states at the end: `Should not be provided if traffic_percentage is provided.` so I thought that no arguments need to be added since you values you said to pass are the default ones. I did remove exposing `Endpoint` though.

I changed
```python
        # TODO(b/221059294): private Endpoints do not yet support traffic splitting
        if cls == Endpoint:
            if traffic_split is None:
                if traffic_percentage > 100:
                    raise ValueError("Traffic percentage cannot be greater than 100.")
                if traffic_percentage < 0:
                    raise ValueError("Traffic percentage cannot be negative.")

            elif traffic_split:
                # TODO(b/172678233) verify every referenced deployed model exists
                if sum(traffic_split.values()) != 100:
                    raise ValueError(
                    )
 ```
to 
```python
        if traffic_percentage and traffic_split:
            raise ValueError(
                "Optionally define either traffic percentage or traffic split, not both."
            )
        if traffic_percentage:
            if traffic_percentage > 100:
                raise ValueError("Traffic percentage cannot be greater than 100.")
            if traffic_percentage < 0:
                raise ValueError("Traffic percentage cannot be negative.")
        if traffic_split:
            # TODO(b/172678233) verify every referenced deployed model exists
            if sum(traffic_split.values()) != 100:
                raise ValueError(
                )
```
to remove the line that checks if the endpoint is private or not and also makes sure that only one of traffic split and traffic percentage is defined.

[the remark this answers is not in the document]
nayaknishant: I'm going to go ahead and explicitly put `traffic_split=None` and `traffic_percentage=100` as well.

google/cloud/aiplatform/models.py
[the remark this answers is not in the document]
nayaknishant: hey Sasha, I added `None` to the return value of the public properties and moved the exceptions to the methods

google/cloud/aiplatform/models.py
```diff
@@ -1326,15 +1374,474 @@ def explain(
...
+                Optional. Custom credentials to use to upload this model. Overrides
+                credentials set in aiplatform.init.
+        """
+        super().__init__(
+            endpoint_name=endpoint_name,
+            project=project,
+            location=location,
+            credentials=credentials,
+        )
+
```
sasha-gitg, line 1536: There should be validation that this resource is a private endpoint because it's possible to construct this class with a non-private endpoint.

google/cloud/aiplatform/models.py
```diff
@@ -1326,15 +1374,474 @@ def explain(
...
+                Required. The url used to send requests and get responses from.
+            body (Dict[Any, Any]):
+                Optional. Data sent to the url in the HTTP request. For private
+                Endpoint, an instance is sent and a prediction response is expected.
+            headers (Dict[str, str]):
+                Optional. Header in the HTTP request.
+
+        Returns:
+            urllib3.response.HTTPResponse:
+                A HTTP Response container.
```
sasha-gitg, line 1732: Raises section

google/cloud/aiplatform/models.py
```diff
@@ -1326,15 +1374,474 @@ def explain(
...
+            parameters (Dict):
+                The parameters that govern the prediction. The schema of
+                the parameters may be specified via Endpoint's
+                DeployedModels' [Model's
+                ][google.cloud.aiplatform.v1beta1.DeployedModel.model]
+                [PredictSchemata's][google.cloud.aiplatform.v1beta1.Model.predict_schemata]
+                ``parameters_schema_uri``.
+
+        Returns:
+            prediction: Prediction with returned predictions and Model Id.
```
sasha-gitg, line 1787: Raises section.

google/cloud/aiplatform/models.py
```diff
@@ -1371,51 +1881,189 @@ def list(
...
-            deployed_models (Sequence[aiplatform.gapic.DeployedModel]):
-                A list of the models deployed in this Endpoint.
-        """
-        self._sync_gca_resource()
-        return self._gca_resource.deployed_models
+    def deploy(
+        self,
+        model: "Model",
+        deployed_model_display_name: Optional[str] = None,
+        traffic_percentage: Optional[int] = 100,
```
sasha-gitg, line 1902: If I understand correctly, only one Model can be deployed to a PrivateEndpoint. What's the purpose of exposing `traffic_percentage` and `traffic_split`?

google/cloud/aiplatform/models.py
```diff
@@ -1371,51 +1881,189 @@ def list(
...
+            explanation_parameters (aiplatform.explain.ExplanationParameters):
+                Optional. Parameters to configure explaining for Model's predictions.
+                For more details, see `Ref docs <http://tinyurl.com/1an4zake>`
+            metadata (Sequence[Tuple[str, str]]):
+                Optional. Strings which should be sent along with the request as
+                metadata.
             sync (bool):
                 Whether to execute this method synchronously. If False, this method
                 will be executed in concurrent Future and any downstream object will
                 be immediately returned and synced when the Future has completed.
```
sasha-gitg, line 1991: Raises section.

google/cloud/aiplatform/models.py
```diff
@@ -1371,51 +1881,189 @@ def list(
...
+                be immediately returned and synced when the Future has completed.
+        """
+        self._sync_gca_resource_if_skipped()
+
+        self._undeploy(
+            deployed_model_id=deployed_model_id,
+            sync=sync,
+        )
 
     def delete(self, force: bool = False, sync: bool = True) -> None:
```
sasha-gitg, line 2064: Why not rely on the super class(Endpoint) implementation?

google/cloud/aiplatform/models.py
```diff
@@ -1174,7 +1210,16 @@ def _undeploy(
         self._sync_gca_resource_if_skipped()
         current_traffic_split = traffic_split or dict(self._gca_resource.traffic_split)
 
-        if deployed_model_id in current_traffic_split:
+        # TODO(b/211351292): Remove check once traffic split is supported
+        # Skip unallocating traffic and raise if new split is provided
+        if self.network:
+            if traffic_split:
+                raise ValueError(
```
sasha-gitg, line 1217: Preference to allow the service to raise in this case.

sasha-gitg (changes requested):

google/cloud/aiplatform/models.py
```diff
@@ -604,25 +622,25 @@ def _validate_deploy_args(
             raise ValueError("Max replica cannot be negative.")
         if deployed_model_display_name is not None:
             utils.validate_display_name(deployed_model_display_name)
-
-        if traffic_split is None:
+        if traffic_percentage and traffic_split:
```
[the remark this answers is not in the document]
nayaknishant: Fixed, removed the line checking for both and explicitly set the values of `traffic_percentage` and `traffic_split` instead of exposing the arguments.

google/cloud/aiplatform/models.py
```diff
@@ -530,16 +545,19 @@ def _unallocate_traffic(
 
         return new_traffic_split
 
-    @staticmethod
+    @classmethod
```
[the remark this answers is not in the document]
nayaknishant: Fixed, reverted to `@staticmethod`

google/cloud/aiplatform/models.py
```diff
@@ -879,6 +903,7 @@ def _deploy_call(
         endpoint_resource_name: str,
         model: "Model",
         endpoint_resource_traffic_split: Optional[proto.MapField] = None,
+        network: str = "",
```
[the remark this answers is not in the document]
nayaknishant: Fixed, changed to `Optional[str]`

google/cloud/aiplatform/models.py
nayaknishant: Fixed, removing these arguments and explicitly setting `traffic_percentage=100` and `traffic_split=None`

google/cloud/aiplatform/models.py
nayaknishant: The `delete` with `force=True` in the super class(Endpoint) implementation calls undeploy_all() which sorts the models based on their traffic split percentage. It then goes through this traffic split dictionary (that maps model_id to traffic split) and deletes them one by one. Since private Endpoint has no traffic split, this dictionary ends up empty, doesn't delete the model, and raises the error to undeploy your model before deleting which is confusing. I think having this seperate `delete()` function could be more easy to understand instead of introducing an if statement in the Endpoint `delete` function just for private Endpoints

google/cloud/aiplatform/models.py
nayaknishant: Fixed, moved to the `undeploy()` method of private Endpoint

setup.py
```diff
@@ -101,6 +101,7 @@
         "google-api-core[grpc] >= 1.31.5, <3.0.0dev,!=2.0.*,!=2.1.*,!=2.2.*,!=2.3.0",
         "proto-plus >= 1.15.0",
         "packaging >= 14.3",
+        "urllib3 >=1.21.1, <1.27",
```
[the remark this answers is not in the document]
nayaknishant: Fixed, created `private_endpoint_extra_require` and also added a try/except import. Also changed `pipelines_extra_requires` to `pipelines_extra_require` to match the other `*_extra_require`

google/cloud/aiplatform/models.py
```diff
@@ -14,12 +14,14 @@
...
 # limitations under the License.
 #
+import json
 import pathlib
 import proto
 import re
 import shutil
 import tempfile
-from typing import Dict, List, NamedTuple, Optional, Sequence, Tuple, Union
+import urllib3
```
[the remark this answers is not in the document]
nayaknishant: Fixed, added a separate block for the try/except import

google/cloud/aiplatform/models.py
sasha-gitg: SG. Thanks for the detailed explanation!

google/cloud/aiplatform/models.py
```diff
@@ -14,12 +14,23 @@
...
 #
+import json
 import pathlib
 import proto
 import re
 import shutil
 import tempfile
-from typing import Dict, List, NamedTuple, Optional, Sequence, Tuple, Union
+
+try:
```
sasha-gitg, line 24: This should be imported within the methods that use this module so this doesn't throw when aiplatform is imported. I think this is only the PrivateEndpoint constructors.

google/cloud/aiplatform/models.py
```diff
@@ -14,12 +14,23 @@
...
+    import urllib3
+except ImportError:
+    raise ImportError(
+        """cannot import the urllib3 HTTP client.
+    Please ensure the correct version of urllib3 has been installed.
+    Check `setup.py` in the root directory for details."""
```
sasha-gitg, line 30: Preference to recommend installing `google-cloud-aiplatform[private_endpoints]`

google/cloud/aiplatform/models.py
nayaknishant: fixed, added Raises section

google/cloud/aiplatform/models.py
sasha-gitg: We should remove this validation and allow the service to throw when the user attempts to deploy a second model to a PrivateEndpoint.

google/cloud/aiplatform/models.py
```diff
@@ -1326,15 +1360,508 @@ def explain(
...
+
+        return endpoint
+
+    def _http_request(
+        self,
+        method: str,
+        url: str,
+        body: Optional[Dict[Any, Any]] = None,
+        headers: Optional[Dict[str, str]] = None,
+    ):
```
sasha-gitg, line 1721: Return type hint.

google/cloud/aiplatform/models.py
```diff
@@ -1326,15 +1360,508 @@ def explain(
...
+        Returns:
+            prediction: Prediction with returned predictions and Model Id.
+
+        Raises:
+            RuntimeError: If a model has not been deployed a request cannot be made.
+        """
+        self.wait()
+        self._sync_gca_resource_if_skipped()
+
+        if not self._gca_resource.deployed_models:
```
sasha-gitg, line 1809: Remove this validation. Require service to throw this validation.

google/cloud/aiplatform/models.py
```diff
@@ -238,10 +234,16 @@ def create(
...
+                Required. Project to retrieve endpoint from. If not set, project
+                set in aiplatform.init will be used.
+            location (str):
+                Required. Location to retrieve endpoint from. If not set, location
+                set in aiplatform.init will be used.
             credentials (auth_credentials.Credentials):
                 Optional. Custom credentials to use to upload this model. Overrides
                 credentials set in aiplatform.init.
-            encryption_spec_key_name (Optional[str]):
+            encryption_spec_key_name (str):
```
ivanmkc, line 246: Good catch.

google/cloud/aiplatform/models.py
```diff
@@ -721,14 +740,14 @@ def deploy(
...
-            explanation_metadata,
-            explanation_parameters,
+            min_replica_count=min_replica_count,
+            max_replica_count=max_replica_count,
+            accelerator_type=accelerator_type,
+            deployed_model_display_name=deployed_model_display_name,
+            traffic_split=traffic_split,
+            traffic_percentage=traffic_percentage,
+            explanation_metadata=explanation_metadata,
+            explanation_parameters=explanation_parameters,
```
ivanmkc, line 750: Nice

google/cloud/aiplatform/models.py
```diff
@@ -1326,15 +1364,508 @@ def explain(
...
+                Optional. Project to retrieve endpoint from. If not set, project
+                set in aiplatform.init will be used.
+            location (str):
+                Optional. Location to retrieve endpoint from. If not set, location
+                set in aiplatform.init will be used.
+            credentials (auth_credentials.Credentials):
+                Optional. Custom credentials to use to upload this model. Overrides
+                credentials set in aiplatform.init.
+        """
+        try:
```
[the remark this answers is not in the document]
ivanmkc: What's our philosophy regarding extras?

google/cloud/aiplatform/models.py
```diff
@@ -1326,15 +1364,508 @@ def explain(
...
+                Optional. Header in the HTTP request.
+
+        Returns:
+            urllib3.response.HTTPResponse:
+                A HTTP Response container.
+
+        Raises:
+            RuntimeError: If a HTTP request could not be made.
+        """
+        try:
```
ivanmkc, line 1746: ditto

google/cloud/aiplatform/models.py
```diff
@@ -1326,15 +1364,508 @@ def explain(
...
+            raise ImportError(
+                "Cannot import the urllib3 HTTP client. Please install google-cloud-aiplatform[private_endpoints]."
+            )
+
+        try:
+            response = self._http_client.request(
+                method=method, url=url, body=body, headers=headers
+            )
+
+            if response.status < 300:
```
ivanmkc, line 1758: Use a constant

ivanmkc (commented):
Since you added the `network` parameter to aiplatform.init, make sure to update MatchingEngine to use that as a default.

google/cloud/aiplatform/models.py
```diff
@@ -1326,15 +1364,508 @@ def explain(
...
+        """Make a prediction against this private Endpoint using a HTTP request.
+        This method must be called within the network the private Endpoint is peered to.
+        The predict() call will fail otherwise. To check, use `PrivateEndpoint.network`.
+
+        Example usage:
+            response = my_private_endpoint.predict(instances=[...])
+            my_predictions = response.predictions
+
+        Args:
+            instances (List):
```
ivanmkc, line 1784: Maybe make a note that the instance types must be JSON serializable.

Seems there is no clear guideline on how to typehint Serializable. Closest I see is https://www.oreilly.com/library/view/mastering-object-oriented-python/9781789531367/13d1b3d7-12ef-4d21-bd0b-1d45c3ac1ad8.xhtml.

I think including the requirement in the docstring is fine for now.

google/cloud/aiplatform/models.py
```diff
@@ -1326,15 +1364,508 @@ def explain(
...
+            parameters (Dict):
+                The parameters that govern the prediction. The schema of
+                the parameters may be specified via Endpoint's
+                DeployedModels' [Model's
+                ][google.cloud.aiplatform.v1beta1.DeployedModel.model]
+                [PredictSchemata's][google.cloud.aiplatform.v1beta1.Model.predict_schemata]
+                ``parameters_schema_uri``.
+
+        Returns:
+            prediction: Prediction with returned predictions and Model Id.
```
ivanmkc, line 1805: Seems like we have no consistent way to denote return parameters.

Prefer this format as it includes the type. Example:

```
        Returns:
            prediction_client (prediction_service_client.PredictionServiceClient):
                Initialized prediction client with optional overrides.
```

google/cloud/aiplatform/models.py
```diff
@@ -1326,15 +1364,508 @@ def explain(
...
+            location (str):
+                Optional. Location to construct Endpoint object from. If not set,
+                location set in aiplatform.init will be used.
+            credentials (auth_credentials.Credentials):
+                Optional. Custom credentials to use to construct Endpoint.
+                Overrides credentials set in aiplatform.init.
+
+        Returns:
+            PrivateEndpoint:
+                An initialized PrivateEndpoint resource.
```
sasha-gitg, line 1700: Raises section

google/cloud/aiplatform/models.py
```diff
@@ -1326,15 +1364,508 @@ def explain(
...
+                Endpoint, an instance is sent and a prediction response is expected.
+            headers (Dict[str, str]):
+                Optional. Header in the HTTP request.
+
+        Returns:
+            urllib3.response.HTTPResponse:
+                A HTTP Response container.
+
+        Raises:
+            RuntimeError: If a HTTP request could not be made.
```
sasha-gitg, line 1744: Add other exception types as well.

google/cloud/aiplatform/models.py
sasha-gitg: Why was this resolved?

google/cloud/aiplatform/models.py
```diff
@@ -1371,51 +1905,190 @@ def list(
...
+            explanation_parameters=explanation_parameters,
+            metadata=metadata,
+            sync=sync,
+        )
 
-        return self
+    def undeploy(
+        self,
+        deployed_model_id: str,
+        traffic_split: Optional[Dict[str, int]] = None,
```
sasha-gitg, line 2042: Why is this exposed if only one Model can be deployed?

google/cloud/aiplatform/models.py
```diff
@@ -2070,31 +2751,48 @@ def deploy(
...
                 Whether to execute this method synchronously. If False, this method
                 will be executed in concurrent Future and any downstream object will
                 be immediately returned and synced when the Future has completed.
             deploy_request_timeout (float):
                 Optional. The timeout for the deploy request in seconds.
+
         Returns:
-            endpoint ("Endpoint"):
+            endpoint (Union[Endpoint, PrivateEndpoint]):
                 Endpoint with the deployed model.
```
sasha-gitg, line 2773: Raises section

google/cloud/aiplatform/models.py
```diff
@@ -2212,30 +2914,49 @@ def _deploy(
...
+                    project=self.project,
+                    location=self.location,
+                    credentials=self.credentials,
+                    encryption_spec_key_name=encryption_spec_key_name,
+                )
 
         _LOGGER.log_action_start_against_resource("Deploying model to", "", endpoint)
 
-        Endpoint._deploy_call(
+        endpoint.__class__._deploy_call(
```
sasha-gitg, line 2959: I don't think you need `__class__`

sasha-gitg (changes requested):

ivanmkc (approved):
LGTM

google/cloud/aiplatform/models.py
```diff
@@ -2399,24 +3110,36 @@ def _deploy(
...
             display_name = self.display_name[:118] + "_endpoint"
-            endpoint = Endpoint.create(
-                display_name=display_name,
-                project=self.project,
-                location=self.location,
-                credentials=self.credentials,
-                encryption_spec_key_name=encryption_spec_key_name,
-            )
+
+            if not isinstance(endpoint, PrivateEndpoint):
```
sasha-gitg, line 3122: endpoint is always `None` based on the if statement above, shouldn't this be `if not network`? If so, preference to add unit test cases to catch this issue.

sasha-gitg (changes requested):

PR #1062, opened 2022-03-04
feat: Add done method for pipeline, training, and batch prediction jobs

sararob:
Added a done method via a `DoneMixin` class to check the status of long running jobs (returns True or False based on job state):
* Implemented by `PipelineJob`, `_Job`, and `_TrainingJob`
* Added system tests in `aiplatform/tests/system/aiplatform/test_e2e_tabular.py`
* Added pipeline job tests in `tests/unit/aiplatform/test_pipeline_jobs.py`
* Still need to add unit tests in `test_jobs` and `test_training_jobs`

Fixes b/215396514

### DIFF
--- a/google/cloud/aiplatform/base.py
+++ b/google/cloud/aiplatform/base.py
@@ -1220,3 +1220,36 @@ def get_annotation_class(annotation: type) -> type:
         return annotation.__args__[0]
     else:
         return annotation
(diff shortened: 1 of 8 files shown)


sasha-gitg (approved):
LGTM. Great work!

PR #1067, opened 2022-03-09
fix: Read project id from initializer.global_config and pass it to the container.

abcdefgs0324:
Fixed b/222604800.


google/cloud/aiplatform/prediction/local_endpoint.py
```diff
@@ -140,19 +145,42 @@ def __init__(
...
         try:
+            try:
+                project_id = initializer.global_config.project
+                _logger.info(
+                    f"Got the project id from the global config: {project_id}."
+                )
+            except (GoogleAuthError, ValueError):
+                project_id = None
+
+            envs = (
```
kthytang, line 165: Are you just trying to copy `self.serving_container_environment_variables` here?

Can you just do...

```
env = {}
if self.serving_container_environment_variables:
    env = dict(self.serving_container_environment_variables)
if project_id is not None:
    envs[_GCLOUD_PROJECT_ENV] = project_id
```

kthytang (changes requested):

google/cloud/aiplatform/prediction/local_endpoint.py
abcdefgs0324: Done.

kthytang: I think this is cleaner, but I'll leave up to you!
```
env = {}
if self.serving_container_environment_variables:
    env = dict(self.serving_container_environment_variables)
```

kthytang (approved):

PR #1068, opened 2022-03-09
fix: add resource manager utils to get project ID from project number

### LINKED ISSUE #852: The SDK does not pick up the correct user project when running on Vertex Pipelines or CustomJobs

morgandu:
- [x] added resource_manager_utils for converting project number to project id
- [x] updated project_id for featurestore df related method 
- [x] added/updated unit tests
- [x] added setup package for google-cloud-resource-manager

Fixes #852, #943
b/216497263


dizcology (approved):

Ark-kun:
Thank you for this fix.

However I'm not sure it fixes https://github.com/googleapis/python-aiplatform/issues/852 and  https://github.com/googleapis/python-aiplatform/pull/943

PR #1069, opened 2022-03-09
fix: Optimize CPR Dockerfile.

abcdefgs0324:
- Allow to use docker build cache.
- Reorder the operations in Dockerfile.
- Allow to install packages from extra_packages.

Fixed b/222587940.


google/cloud/aiplatform/docker_utils/build.py
```diff
@@ -262,35 +297,65 @@ def make_dockerfile(
...
         """.format(
-            base_image=base_image,
             workdir=shlex_quote(container_workdir),
             container_home=shlex_quote(container_home),
         )
     )
 
-    dockerfile += _prepare_package_entry(main_package)
+    # Installs extra requirements which do not involve user source code.
+    dockerfile += _prepare_dependency_entries(
```
kthytang, line 325: Why do the requirements_path, extra_requirements, etc, need to be separated into individual _prepare_dependency_entries calls?

kthytang (changes requested):

google/cloud/aiplatform/docker_utils/build.py
abcdefgs0324: Separating the calls allows us to control the order of the operations in Dockerfile instead of depending on the code in `_prepare_dependency_entries`. Therefore, we could install the things that are likely to be unchanged first to optimize the time building images with docker build cache.

kthytang: Ah, got it. Thanks.

kthytang (approved):

PR #1071, opened 2022-03-10
fix: Don't throw exception when getting representation of unrun GCA objects

samgoodman:
- [] Appropriate docs were updated (if necessary)

Fixes b/197264919 🦕


tests/unit/aiplatform/test_custom_job.py
```diff
@@ -661,3 +661,36 @@ def test_create_from_local_script_with_all_args(
         assert (
             job._gca_resource.state == gca_job_state_compat.JobState.JOB_STATE_SUCCEEDED
         )
+
+    def test_check_custom_job_availability(self, get_custom_job_mock, create_custom_job_mock):
```
sasha-gitg, line 665: We're moving towards using `@pytest.mark.usefixtures` when we don't consume the fixture in the test directly (asserting some state against it) in the test:

```suggestion
    @pytest.mark.usefixtures('get_custom_job_mock', 'create_custom_job_mock')
    def test_check_custom_job_availability(self):
```

sasha-gitg (commented):
LGTM! Minor comments.

google/cloud/aiplatform/base.py
```diff
@@ -655,6 +655,15 @@ def gca_resource(self) -> proto.Message:
         self._assert_gca_resource_is_available()
         return self._gca_resource
 
+    @property
+    def resource_is_available(self) -> bool:
```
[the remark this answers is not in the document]
samgoodman: Done

tests/unit/aiplatform/test_custom_job.py
samgoodman: Done

sasha-gitg (approved):

PR #1076, opened 2022-03-10
fix: Fix import error string showing wrong pip install path

taiseiak:
Fix import error showing `python-aiplatform` instead of `google-cloud-aiplatform`.

### DIFF
--- a/google/cloud/aiplatform/explain/lit.py
+++ b/google/cloud/aiplatform/explain/lit.py
@@ -29,7 +29,7 @@ try:
 except ImportError:
     raise ImportError(
         "LIT is not installed and is required to get Dataset as the return format. "
-        'Please install the SDK using "pip install python-aiplatform[lit]"'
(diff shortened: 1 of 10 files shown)


google/cloud/aiplatform/explain/lit.py
```diff
@@ -29,23 +29,23 @@
...
+        'Please install the SDK using "pip install google-cloud-aiplatform[lit]"'
     )
 
 try:
     import tensorflow as tf
 except ImportError:
     raise ImportError(
         "Tensorflow is not installed and is required to load saved model. "
-        'Please install the SDK using "pip install pip install python-aiplatform[lit]"'
+        'Please install the SDK using "pip install pip install google-cloud-aiplatform[lit]"'
```
parthea, line 40:
```suggestion
        'Please install the SDK using "pip install google-cloud-aiplatform[lit]"'
```

parthea (approved):
minor observation, otherwise LGTM

PR #1080, opened 2022-03-15
chore: add cdpe-cloudai team as default codeowner

parthea:


tswast (approved):
I recall this is usually templated, right?

parthea:
@tswast Correct, there is a template for `CODEOWNERS` in synthtool [here](https://github.com/googleapis/synthtool/blob/master/synthtool/gcp/templates/python_library/.github/CODEOWNERS) however this repo requires granular control which is not supported by the template.

PR #1081, opened 2022-03-15
Fix: Give aiplatform logging its own log namespace, let the user configure their own root logger

samgoodman:

Addresses https://github.com/googleapis/python-aiplatform/issues/967 and most of internal b/222347808


google/cloud/aiplatform/featurestore/entity_type.py
```diff
@@ -37,7 +37,7 @@
 
 from google.cloud import bigquery
 
-_LOGGER = base.Logger(__name__)
+_LOGGER = base.Logger()
```
Ark-kun, line 40: This changes the behavior. Previously every module had it's own logger (based on `__name__`).
Not sure what's the best way.

samgoodman: Nope, that's actually still the behavior! __name__ is the default name in the logger constructor.

This 1. removes boilerplate and 2. prevents bad patterns (__name__ should be used whenever possible, this makes it harder to make a mistake)

Ark-kun: >Nope, that's actually still the behavior! name is the default name in the logger constructor.

I'm not sure this is the case. Previously, every python module passed `__name__` to the `Logger` class constructor. So logger for every module was getting a different name.
Now, the `__name__` is only used in the `logger` module. So all logger names will be the same now.
The key piece of information is that `__name__` is bound to the module where it's used.

morgandu: +1, when `_LOGGER = base.Logger()` is being called in different modules, the default value for `__name__` is still the base, though, it does not seem we are enforcing any logging formatting to show the `__name__`. Prefer to be able to trace the logger since we can.

tests/unit/aiplatform/test_logging.py
```diff
@@ -0,0 +1,51 @@
+# -*- coding: utf-8 -*-
+
+# Copyright 2020 Google LLC
```
sasha-gitg, line 3:
```suggestion
# Copyright 2022 Google LLC
```

google/cloud/aiplatform/base.py
```diff
@@ -47,22 +47,26 @@
...
 _DEFAULT_RETRY = retry.Retry()
 
 
 class Logger:
     """Logging wrapper class with high level helper methods."""
 
-    def __init__(self, name: str = ""):
-        """Initializes logger with name.
+    def __init__(self, name: str):
+        """Initializes logger with optional name.
```
sasha-gitg, line 58: No longer optional.

sasha-gitg (approved):
LGTM with minor comments. Thanks Sam!

PR #1133, opened 2022-04-04
fix: Set serving_container_command to entrypoint instead of command to match our requirements

abcdefgs0324:
as https://cloud.google.com/vertex-ai/docs/predictions/custom-container-requirements#server.


kthytang (approved):
Thanks for fixing. How did this come up?

abcdefgs0324:
I found it while I was playing around.

PR #1137, opened 2022-04-04
chore: remove matching engine add copyright header

morgandu:


sasha-gitg (commented):
Can we also make the the matching_engine directory private? `_matching_engine`

sasha-gitg (approved):

andrewferlitsch (approved):

PR #1142, opened 2022-04-06
fix: Replace env var references in given env, command and arg.

abcdefgs0324:
See https://cloud.google.com/ai-platform/prediction/docs/reference/rest/v1/projects.models.versions#containerspec.

### DIFF
--- a/google/cloud/aiplatform/docker_utils/run.py
+++ b/google/cloud/aiplatform/docker_utils/run.py
@@ -18,6 +18,7 @@
 import logging
 import os
 from pathlib import Path
+import re
 from typing import Dict, Optional, Sequence
 
 try:
@@ -46,6 +47,32 @@ def _get_adc_environment_variable():
         The value of the environment variable or None if unset.
     """
     return os.environ.get(_ADC_ENVIRONMENT_VARIABLE)
+
+
+def _replace_env_var_reference(
+        target: str, env_vars: Dict[str, str],
+    ) -> str:
+    """Replaces the environment variable reference in the given string.
+
+    Variable references $(VAR_NAME) are expanded using the container's environment.
+    If a variable cannot be resolved, the reference in the input string will be unchanged.
+    The $(VAR_NAME) syntax can be escaped with a double $$, ie: $$(VAR_NAME). Escaped
+    references will never be expanded, regardless of whether the variable exists or not.
+    More info:
+    https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#running-a-command-in-a-shell
+
+    Args:
+        target (str):
+            Required. The string to be replaced with the environment variable reference.
+        env_vars (Dict[str, str]):
+            Required. The environment variables used for reference.
+
+    Returns:
+        The updated string.
+    """
+    for key, value in env_vars.items():
+        target = re.sub(fr"(?<!\$)\$\({key}\)", str(value), target)
+    return target
 
 
 def run_prediction_container(
@@ -122,11 +149,10 @@ def run_prediction_container(
     """
     client = docker.from_env()
 
-    envs = (
-        {key: value for key, value in serving_container_environment_variables.items()}
-        if serving_container_environment_variables is not None
-        else {}
-    )
+    envs = {}
+    if serving_container_environment_variables is not None:
+        for key, value in serving_container_environment_variables.items():
+            envs[key] = _replace_env_var_reference(value, envs)
 
     port = prediction_utils.get_prediction_aip_http_port(serving_container_ports)
     envs[prediction.AIP_HTTP_PORT] = port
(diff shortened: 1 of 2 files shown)


tests/unit/aiplatform/test_docker_utils.py
```diff
@@ -154,6 +154,112 @@ def test_run_prediction_container_with_all_parameters(
...
+        environment = {k: v for k, v in serving_container_environment_variables.items()}
+        environment[prediction.AIP_HTTP_PORT] = prediction.DEFAULT_AIP_HTTP_PORT
+        environment[prediction.AIP_HEALTH_ROUTE] = None
+        environment[prediction.AIP_PREDICT_ROUTE] = None
+        environment[prediction.AIP_STORAGE_URI] = ""
+        environment[
+            run._ADC_ENVIRONMENT_VARIABLE
+        ] = run._DEFAULT_CONTAINER_CRED_KEY_PATH
+        # Command references existing environment variables.
+        expected_entrypoint = ["foo", "$$(VAR_1)", "$(VAR_2)"]
```
kthytang, line 210: Can you check to make sure that if the user escapes the syntax, `"$$(VAR_1)"`, that the expected rendering is `"$$(VAR_1)"` and not `"$(VAR_1)"`

kthytang (changes requested):

tests/unit/aiplatform/test_docker_utils.py
abcdefgs0324: Should be `"$(VAR_1)` as https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.22/#envvar-v1-core. Thanks!

google/cloud/aiplatform/docker_utils/run.py
```diff
@@ -138,11 +168,6 @@ def run_prediction_container(
...
     envs[prediction.AIP_STORAGE_URI] = artifact_uri if artifact_uri is not None else ""
 
-    entrypoint = (
-        serving_container_command[:] if serving_container_command is not None else []
-    )
-    command = serving_container_args[:] if serving_container_args is not None else []
-
     credential_from_adc_env = credential_path is None
     credential_path = credential_path or _get_adc_environment_variable()
     if credential_path is not None:
```
kthytang, line 173:
```suggestion
    if credential_path:
```

google/cloud/aiplatform/docker_utils/run.py
```diff
@@ -162,6 +187,17 @@ def run_prediction_container(
     )
     envs[_ADC_ENVIRONMENT_VARIABLE] = credential_mount_path
 
+    entrypoint = (
+        [_replace_env_var_reference(i, envs) for i in serving_container_command]
```
kthytang, line 191: Should `serving_container_command` and `serving_container_args` really be of the type Sequence or can they be Lists? If they can be List, can we default them to [] to avoid this if not None/else [] logic?

Otherwise maybe it's clearer as:
`[_replace_env_var_reference(i, envs) for i in serving_container_command or []]`

or 
`serving_container_command = serving_container_command or []`
`[_replace_env_var_reference(i, envs) for i in serving_container_command]`

kthytang (changes requested):

google/cloud/aiplatform/docker_utils/run.py
```diff
@@ -48,6 +49,36 @@ def _get_adc_environment_variable():
...
+            Required. The environment variables used for reference.
+
+    Returns:
+        The updated string.
+    """
+    # Replace env var references with env vars.
+    for key, value in env_vars.items():
+        target = re.sub(rf"(?<!\$)\$\({key}\)", str(value), target)
+    # Replace $$ with $.
+    target = re.sub(r"\$\$", "$", target)
```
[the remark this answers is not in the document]
abcdefgs0324: I think all double `$$` should be replaced with `$` no matter whether it's followed by the env vars.

From https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.22/#envvar-v1-core,
`
Double $$ are reduced to a single $, which allows for escaping the $(VAR_NAME) syntax: i.e. "$$(VAR_NAME)" will produce the string literal "$(VAR_NAME)".
`

google/cloud/aiplatform/docker_utils/run.py
abcdefgs0324: I choose the second option.

`serving_container_command` and `serving_container_args` are the type Sequence across the SDK. I would like to keep it as is.

google/cloud/aiplatform/docker_utils/run.py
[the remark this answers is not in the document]
kthytang: Got it, SG.

kthytang (approved):

PR #1148, opened 2022-04-06
docs: Added two samples for PyTorch usage.

abcdefgs0324:
1. Create CPR images for PyTorch.
2. Use Triton inference server to serve PyTorch.

### DIFF
--- a/samples/notebooks/prediction/SDK_Pytorch_Custom_Predict.ipynb
+++ b/samples/notebooks/prediction/SDK_Pytorch_Custom_Predict.ipynb
@@ -31,7 +31,7 @@
    "source": [
     "<table align=\"left\">\n",
     "  <td>\n",
-    "    <a href=\"https://github.com/googleapis/python-aiplatform/blob/custom-prediction-routine/samples/notebooks/prediction/SDK_Pytorch_Custom_Container.ipynb\">\n",
+    "    <a href=\"https://github.com/googleapis/python-aiplatform/blob/custom-prediction-routine/samples/notebooks/prediction/SDK_Pytorch_Custom_Predict.ipynb\">\n",
     "      <img src=\"https://cloud.google.com/ml-engine/images/github-logo-32px.png\" alt=\"GitHub logo\">\n",
     "      View on GitHub\n",
     "    </a>\n",
--- a/google/cloud/aiplatform/prediction/README.md
+++ b/google/cloud/aiplatform/prediction/README.md
@@ -248,7 +248,7 @@ endpoint = model.deploy(machine_type="n1-standard-4")
 
 ## Notebook Samples
 
-The four samples below showcase the different ways you can deploy a sklearn iris model with custom pre/post-processing on Vertex AI Prediction.
+The samples below showcase the different ways you can deploy an iris model with custom pre/post-processing on Vertex AI Prediction.
 
 The samples start with the simplest user journey (custom pre/post processing for sklearn). Each subsequent sample follows user journeys that allows for more flexibility during prediction handling. 
 
@@ -263,7 +263,13 @@ Each sample will also cover the local debugging experience.
     - Custom implementation of Predictor and Handler.
     - Customizing the Handler allows the model server to handle csv inputs.
     - https://github.com/googleapis/python-aiplatform/blob/custom-prediction-routine/samples/notebooks/prediction/SDK_Custom_Predict_and_Handler_SDK_Integration.ipynb
+- Custom Predictor, build your own container with Vertex SDK for PyTorch.
+    - Custom implementation of the Predictor.
+    - https://github.com/googleapis/python-aiplatform/blob/custom-prediction-routine/samples/notebooks/prediction/SDK_Pytorch_Custom_Predict.ipynb
 - Custom Predictor, build your own container with docker commands.
     - Custom container building.
     - Users have full control over the docker build process (files included in the container, dependencies installed, entrypoint, etc.).
     - https://github.com/googleapis/python-aiplatform/blob/custom-prediction-routine/samples/notebooks/prediction/SDK_Custom_Predict.ipynb
+- Existing image, test prediction locally and deploy models with Vertex SDK.
+    - Use NVIDIA Triton inference server for PyTorch models.
+    - https://github.com/googleapis/python-aiplatform/blob/custom-prediction-routine/samples/notebooks/prediction/SDK_Triton_PyTorch_Local_Prediction.ipynb
--- a/samples/notebooks/prediction/SDK_Pytorch_Custom_Container.ipynb
+++ b/samples/notebooks/prediction/SDK_Pytorch_Custom_Container.ipynb
@@ -1,0 +1,1603 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "id": "ur8xi4C7S06n"
+   },
+   "outputs": [],
+   "source": [
+    "# Copyright 2022 Google LLC\n",
+    "#\n",
+    "# Licensed under the Apache License, Version 2.0 (the \"License\");\n",
+    "# you may not use this file except in compliance with the License.\n",
+    "# You may obtain a copy of the License at\n",
+    "#\n",
+    "#     https://www.apache.org/licenses/LICENSE-2.0\n",
+    "#\n",
+    "# Unless required by applicable law or agreed to in writing, software\n",
+    "# distributed under the License is distributed on an \"AS IS\" BASIS,\n",
+    "# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.\n",
+    "# See the License for the specific language governing permissions and\n",
+    "# limitations under the License."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "id": "JAPoU8Sm5E6e"
+   },
+   "source": [
+    "<table align=\"left\">\n",
+    "  <td>\n",
+    "    <a href=\"https://github.com/googleapis/python-aiplatform/blob/custom-prediction-routine/samples/notebooks/prediction/SDK_Pytorch_Custom_Container.ipynb\">\n",
+    "      <img src=\"https://cloud.google.com/ml-engine/images/github-logo-32px.png\" alt=\"GitHub logo\">\n",
+    "      View on GitHub\n",
+    "    </a>\n",
+    "  </td>\n",
+    "</table>"
+   ]
+  },
+  {
+   "cell_type": "markdown",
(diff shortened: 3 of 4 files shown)


samples/notebooks/prediction/SDK_Pytorch_Custom_Predict.ipynb
```diff
@@ -0,0 +1,1603 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "id": "8phJWBhe7u9u",
+    "tags": []
+   },
+   "source": [
+    "#### Set up credentials\n",
```
samthrasher, line 972: We can simplify these notebooks a lot by removing the requirement to sign in for local testing. It doesn't really make sense in the first place that users would have to be signed in for testing their container locally. How hard is it to have our system handle a model artifact that could either be on GCS or in a local directory?

samples/notebooks/prediction/SDK_Pytorch_Custom_Predict.ipynb
```diff
@@ -0,0 +1,1603 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "id": "62ed2d334d0f"
+   },
+   "outputs": [],
+   "source": [
+    "with local_model.deploy_to_local_endpoint(\n",
```
samthrasher, line 1224: Is there a non-janky way I can use deploy_to_local_endpoint interactively? (i.e. without the context manager?) I want the container to hang out for a while so that I can time how long a request takes, and I don't want to block the notebook kernel. Actuating the context manager hooks directly works, but is not a good interface.

e.g. right now I'm doing
```
<cell>
endpt = local_model.deploy_to_local_endpoint(
    artifact_uri=f"{BUCKET_NAME}/{MODEL_ARTIFACT_DIR}",
    credential_path=CREDENTIALS_FILE,
)

local_endpoint = endpt.__enter__()
</cell>

<cell>
%%timeit
predict_response = local_endpoint.predict(
    request_file=INPUT_FILE,
    headers={"Content-Type": "application/json"},
)
</cell>
</cell>
endpt.__exit(None, None, None)
</cell>

File: samples/notebooks/prediction/SDK_Pytorch_Custom_Predict.ipynb [hunk 3]
```diff
@@ -0,0 +1,1603 @@
...
+    "\n",
+    "import pandas as pd\n",
+    "import pickle\n",
+    "import torch\n",
+    "\n",
+    "from google.cloud import storage\n",
+    "from google.cloud.aiplatform.prediction.predictor import Predictor\n",
+    "\n",
+    "\n",
+    "class CprPredictor(Predictor):\n",
```
samthrasher, line 803: More specific name might be appropriate. CustomPyTorchPredictor?

samples/notebooks/prediction/SDK_Pytorch_Custom_Predict.ipynb
```diff
@@ -0,0 +1,1603 @@
...
+    "                f\"{gcs_artifacts_uri}/{model_file}\", model_f\n",
+    "            )\n",
+    "\n",
+    "        self._model = torch.load(model_file)\n",
+    "\n",
+    "    def predict(self, instances):\n",
+    "        \"\"\"Performs prediction.\"\"\"\n",
+    "        instances = instances[\"instances\"]\n",
+    "        data = pd.DataFrame(instances).values\n",
+    "        inputs = torch.Tensor(data)\n",
```
samthrasher, line 824: These should go in the preprocess method. Predict should take and return a torch.Tensor. Please add the type annotations as well.

samples/notebooks/prediction/SDK_Pytorch_Custom_Predict.ipynb
```diff
@@ -0,0 +1,1603 @@
...
+    "        outputs = self._model(inputs)\n",
+    "        _ , predicted = torch.max(outputs, 1)\n",
+    "        return {\"predictions\": [self._class_names[class_num] for class_num in predicted]}"
```
samthrasher, line 827: These should go in the postprocess method. Make sure it's clear what this is doing (find the index with maximum value from the class probabilities produced by the module, then return the class name)

samples/notebooks/prediction/SDK_Pytorch_Custom_Predict.ipynb
```diff
@@ -0,0 +1,1603 @@
...
+    "            )\n",
+    "\n",
+    "        self._model = torch.load(model_file)\n",
+    "\n",
+    "    def predict(self, instances):\n",
+    "        \"\"\"Performs prediction.\"\"\"\n",
+    "        instances = instances[\"instances\"]\n",
+    "        data = pd.DataFrame(instances).values\n",
+    "        inputs = torch.Tensor(data)\n",
+    "        outputs = self._model(inputs)\n",
```
samthrasher, line 825: Use torch.no_grad here
https://pytorch.org/docs/stable/generated/torch.no_grad.html

samthrasher (changes requested):
This is great! One other thing I think would be helpful here for PyTorch users is saving/loading using an `nn.Module` subclass from an imported Python module. PyTorch's saved models are pickle files, so this won't work unless the same class is imported with the same name. The right way to do it is not obvious, and this is a common pitfall for folks who train a model from such a class, export it, then want to serve it. Some demo for GPU functionality would also be useful here ofc, but we still need to work out the local testing story for that.

samples/notebooks/prediction/SDK_Pytorch_Custom_Predict.ipynb
samthrasher: or torch.inference_mode which does the same, but is named more explicitly https://pytorch.org/docs/stable/generated/torch.inference_mode.html

samples/notebooks/prediction/SDK_Pytorch_Custom_Predict.ipynb
```diff
@@ -0,0 +1,1603 @@
...
+   "outputs": [],
+   "source": [
+    "!gcloud auth login"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Grant roles to the service account. Since we will be using this service account to download the model artifacts from Google Cloud Storage, we grant the service account **Storage Object Admin** role. See [IAM roles for Cloud Storage](https://cloud.google.com/storage/docs/access-control/iam-roles) for more details."
```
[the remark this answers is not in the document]
abcdefgs0324: Viewer is not sufficient.

samples/notebooks/prediction/SDK_Pytorch_Custom_Predict.ipynb
abcdefgs0324: I think there are two parts for this.
1. Our local run allows users to pass a local directory.
2. Users write two ways to load models from GCS and local files.
1 is not hard but in my opinion it doesn't make much sense for users to write two ways to load models in order to test images locally. Any it is possible (at a lower chance though) that images work well with local directories but are broken with GCS path if the code in `load` function is wrong.

samples/notebooks/prediction/SDK_Pytorch_Custom_Predict.ipynb
abcdefgs0324: The above is the only way I'm aware of.

samples/notebooks/prediction/SDK_Pytorch_Custom_Predict.ipynb
abcdefgs0324: Done.

samples/notebooks/prediction/SDK_Pytorch_Custom_Predict.ipynb
abcdefgs0324: Done.

samples/notebooks/prediction/SDK_Pytorch_Custom_Predict.ipynb
abcdefgs0324: Done.

samples/notebooks/prediction/SDK_Pytorch_Custom_Predict.ipynb
abcdefgs0324: cool, used torch.inference_mode here.

samples/notebooks/prediction/SDK_Pytorch_Custom_Predict.ipynb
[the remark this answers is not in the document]
samthrasher: That seems odd. Are we doing anything other than downloading objects from preexisting buckets with the service account?

samples/notebooks/prediction/SDK_Pytorch_Custom_Predict.ipynb
samthrasher: OK, then please consider supporting this use case, either in this PR or a future one.

samples/notebooks/prediction/SDK_Pytorch_Custom_Predict.ipynb
samthrasher: OK, sounds like we should make it easier for the users, then. Let's provide an importable function that can load from either GCS or a local directory, like tf.io.gfile (except we don't want to include all of tensorflow by default)

samples/notebooks/prediction/SDK_Pytorch_Custom_Predict.ipynb
samthrasher: Easiest thing to do would be to make some friendly aliases for local_endpoint's context manager methods.

samples/notebooks/prediction/SDK_Pytorch_Custom_Predict.ipynb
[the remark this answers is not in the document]
abcdefgs0324: I found I was wrong. The Viewer propagation was much slower than Admin in the first project I tested. Viewer permission is sufficient. Updated all samples.

samples/notebooks/prediction/SDK_Pytorch_Custom_Predict.ipynb
abcdefgs0324: Ack.

samples/notebooks/prediction/SDK_Pytorch_Custom_Predict.ipynb
abcdefgs0324: Ack.

samthrasher (approved):

kthytang (approved):

PR #1152, opened 2022-04-07
chore: Pull the image if it doesn't exist before image copying.

abcdefgs0324:


kthytang (changes requested):
Thanks for adding!

Do we need to do the same to check if the image exists before running locally?

abcdefgs0324:
docker run defaults pulling the image if it was not found in the image cache, or use the cached image otherwise.

kthytang (approved):

PR #1154, opened 2022-04-08
docs: update model code snippet order in README

nayaknishant:
Changing order of model code snippets, so model is retrieved first before doing the calls.


yil532 (approved):
LGTM. Proper order change of the code sample.

PR #1162, opened 2022-04-13
feat: add update endpoint

morgandu:
- [x] add update endpoint
- [x] update unit tests
- [x]  make sure all unit tests pass
- [ ]  make sure all system tests pass locally

Fixes b/227509107

blocked on b/227467389


google/cloud/aiplatform/models.py
```diff
@@ -1227,6 +1198,200 @@ def _instantiate_prediction_client(
...
+        """
+        traffic_split_model_ids = list(traffic_split.keys())
+        endpoint_model_ids = set(
+            [] if not endpoint else [model.id for model in endpoint.list_models()]
+        )
+
+        if deploy_model:
+            endpoint_model_ids.add(_DEPLOYING_MODEL_TRAFFIC_SPLIT_KEY)
+
+        if any(
```
sasha-gitg, line 1237: Is the SDK introducing any additional functionality that requires we validate the deployed models locally? If not, then preference to push this validation to the service.

tests/unit/aiplatform/test_endpoints.py
```diff
@@ -25,6 +25,8 @@
 from google.api_core import operation as ga_operation
```
sasha-gitg, line 25: Preference to update system tests with an update call with traffic split. Can be added to the end to end test.

google/cloud/aiplatform/models.py
samgoodman: Agreed. I'll pick up this branch from Morgan and finish this PR this week.

samgoodman (approved):

PR #1167, opened 2022-04-13
feat: add ModelEvaluation support

sararob:
Added initial support for the ModelEvaluation gapic client:
* `ModelEvaluation` class defined in `model_evaluation/`
* Exposed `ModelEvaluation` class in `models.py`
* Added related tests

Fixes b/224573576 🦕

### DIFF
--- a/google/cloud/aiplatform/model_evaluation/model_evaluation.py
+++ b/google/cloud/aiplatform/model_evaluation/model_evaluation.py
@@ -48,8 +48,8 @@ class ModelEvaluation(base.VertexAiResourceNounWithFutureManager):
             pipeline output artifact. Returns None if the underlying
             PipelineJob has not yet completed.
         """
-        # print(self.api_client.get_model_evaluation())
-        # return self.api_client.get_model_evaluation()
+        if self._gca_resource.metrics:
+            return self.to_dict()['metrics']
 
     def __init__(
         self,
--- a/google/cloud/aiplatform/model_evaluation/model_evaluation.py
+++ b/google/cloud/aiplatform/model_evaluation/model_evaluation.py
@@ -39,54 +39,13 @@ class ModelEvaluation(base.VertexAiResourceNounWithFutureManager):
     _format_resource_name_method = "model_evaluation_path"
 
     @property
-    def evaluation_metrics(self) -> Optional[Dict[str, Any]]:
+    def metrics(self) -> Optional[Dict[str, Any]]:
         """Gets the evaluation metrics from the Model Evaluation.
         Returns:
             A dict with model metrics created from the Model Evaluation or
             None if the metrics for this evaluation are empty.
         """
-        if self._gca_resource.metrics:
-            return self.to_dict()["metrics"]
-
-    @property
-    def summary_metrics(self) -> Optional[Dict[str, Any]]:
-        """Gets the summary metrics for the Model Evaluation.
-
-        For classification models, this includes auPrc, auRoc, logLoss for the model overall,
-        and precision, recall, and F1 score at a confidence threshold of 0.5.
-
-        For regression models, the summary metrics are the same as `evaluation_metrics`.
-        This includes RMSE, MAE, MAPE, r^2, and RMSLE.
-
-        Returns:
-            A dict with model metrics created from the Model Evaluation or
-            None if the metrics for this evaluation are empty.
-        """
-
-        if self._gca_resource.metrics:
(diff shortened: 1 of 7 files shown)


google/cloud/aiplatform/models.py
```diff
@@ -3210,3 +3211,93 @@ def upload_tensorflow_saved_model(
...
+            ModelEvaluation resource.
+        """
+        if not evaluation_id:
+            evaluation_resource_name = self.list_model_evaluations(self.resource_name)[
+                0
+            ].resource_name
+        else:
+            # TODO: validate evaluation_id
+            evaluation_resource_name = (
+                f"{self.resource_name}/evaluations/{evaluation_id}"
```
sasha-gitg, line 3298: Preference to rely on Model resource names parsers and formatter. Something like:

```Python
parts = self._parse_resource_name(self.resource_name)
eval_resource_name = ModelEvaluation._format_resource_name(**parts, evalution=evaluation_id)
```

google/cloud/aiplatform/models.py
```diff
@@ -3210,3 +3211,93 @@ def upload_tensorflow_saved_model(
...
+
+        Args:
+            evaluation_id (str):
+                Optional. The ID of the model evaluation to retrieve.
+        Returns:
+            model_evaluation.ModelEvaluation: Instantiated representation of the
+            ModelEvaluation resource.
+        """
+        if not evaluation_id:
+            evaluation_resource_name = self.list_model_evaluations(self.resource_name)[
```
sasha-gitg, line 3292: Preference to warn if more than one model evaluation is present and log that we are returning this particular model eval resource name.

Additionally, this can just be returned directly.

sasha-gitg (changes requested):

google/cloud/aiplatform/model_evaluation/model_evaluation.py
```diff
@@ -0,0 +1,132 @@
...
+    Dict,
+    Optional,
+)
+
+
+class ModelEvaluation(base.VertexAiResourceNounWithFutureManager):
+
+    client_class = utils.ModelClientWithOverride
+    _resource_noun = "evaluations"
+    _delete_method = "delete_pipeline_job"
```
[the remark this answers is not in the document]
sararob: It doesn't look like `ModelEvaluation` has a delete method, just get and import: https://github.com/googleapis/python-aiplatform/blob/c7bfd8000eaace6e3dc01d17270b364e1db3f530/google/cloud/aiplatform_v1/services/model_service/client.py#L1227 Should I make this an empty string or use `delete_model` (I'm guessing this deletes the eval metrics)?

google/cloud/aiplatform/models.py
```diff
@@ -3210,3 +3211,80 @@ def upload_tensorflow_saved_model(
...
+            resource_noun=Model._resource_noun,
+            parse_resource_name_method=Model._parse_resource_name,
+            format_resource_name_method=Model._format_resource_name,
+        )
+
+        self.wait()
+
+        return model_evaluation.ModelEvaluation._list(
+            parent=parent,
+        )
```
sasha-gitg, line 3243: `self.credentials` should be forwarded here

google/cloud/aiplatform/models.py
```diff
@@ -3210,3 +3211,80 @@ def upload_tensorflow_saved_model(
...
+
+        Args:
+            evaluation_id (str):
+                Optional. The ID of the model evaluation to retrieve.
+        Returns:
+            model_evaluation.ModelEvaluation: Instantiated representation of the
+            ModelEvaluation resource.
+        """
+        if not evaluation_id:
+            evaluation_resource_name = self.list_model_evaluations()[0].resource_name
```
sasha-gitg, line 3274: This can throw an exception if there are no model evalutions. Also there is no need to do the additional GET call below:

```Python
evals = self.list_model_evaluations()

if len(evals) > 1:
  _LOGGER.warning(
                f"Your model has more than one model evaluation, this is returning only one evaluation resource: {evals[0].resource_name}"
            )
            
return evals[0] if evals else evals
```

sasha-gitg (changes requested):

google/cloud/aiplatform/models.py
sararob: Good point re: models w/ no evaluations. I think I've fixed this.

sasha-gitg (approved):

PR #1168, opened 2022-04-19
feat: Add endpoind_id arg to Endpoint#create

### LINKED ISSUE #1169: `endpoint_id` arg to `Endpoint#create` method

[`models.Endpoint#create` method](https://github.com/googleapis/python-aiplatform/blob/c1e899dba3f57e515b1f1958e962f355276460c4/google/cloud/aiplatform/models.py#L198) does not have `endpoint_id` argument. Hence, we can't create Vertex Endpoints with a specific Endpoint ID. Endpoint ID will always be generated randomly upon creation.

Add `endpoint_id` argument to the method.

 **Describe alternatives you've considered**

N/A

 **Additional context**

N/A

mai-nakagawa:

Fixes #1169 🦕

### DIFF
--- a/google/cloud/aiplatform/models.py
+++ b/google/cloud/aiplatform/models.py
@@ -198,7 +198,6 @@ class Endpoint(base.VertexAiResourceNounWithFutureManager):
     def create(
         cls,
         display_name: Optional[str] = None,
-        endpoint_id: Optional[str] = None,
         description: Optional[str] = None,
         labels: Optional[Dict[str, str]] = None,
         metadata: Optional[Sequence[Tuple[str, str]]] = (),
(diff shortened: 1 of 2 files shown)


google/cloud/aiplatform/models.py
```diff
@@ -198,6 +198,7 @@ def network(self) -> Optional[str]:
     def create(
         cls,
         display_name: Optional[str] = None,
+        endpoint_id: Optional[str] = None,
```
sasha-gitg, line 201: Please add this field to the end of the signature to avoid breaking changes on order of arguments(not necessary on private methods).

sasha-gitg (changes requested):
Thanks for this contribution, Mai!

google/cloud/aiplatform/models.py
mai-nakagawa: Thanks for your review! Moved the field to the end(also on the private method): https://github.com/googleapis/python-aiplatform/pull/1168/commits/14fa07104b3c90cf2b06cdc4b895e9166cf4ef7c

sasha-gitg (approved):

Issue #1036, opened 2022-02-24
copytree's dirs_exist_ok parameter is only available in Python 3.8+


ivanmkc:
https://github.com/googleapis/python-aiplatform/blob/3aec6a7b8f26ef2a5b378a6224d6402e3b42c917/google/cloud/aiplatform/utils/source_utils.py#L174

See https://docs.python.org/3/whatsnew/3.8.html#shutil

Need to replace with something else to support <3.8 or else an exception occurs.

# TODO:
- [ ] Replace shutil.copytree with something more compatible
- [ ] Add unit test to cover passing in a folder to `_TrainingScriptPythonPackager`

kweinmeister:
A potential workaround for <3.8: https://stackoverflow.com/a/31039095

ivanmkc:
Unfortunately, "distutils has been deprecated in Python 3.10 and is planned to be removed in 3.12, see [PEP 632](https://www.python.org/dev/peps/pep-0632/) for info."

Perhaps https://stackoverflow.com/a/12514470 would work.

kweinmeister:
Yup, and another option is to check `if sys.version_info < (3, 8)` to only apply the workaround in the post you mentioned on 3.6 and 3.7.

PR #1170, opened 2022-04-19
fix: Removed dirs_exist_ok parameter as it's not backwards compatible

ivanmkc:
Removed dirs_exist_ok parameter as it's not backwards compatible.

Decided to delete destination folder as it would keep the destination faithful to the source.

Fixes https://github.com/googleapis/python-aiplatform/issues/1036


ivanmkc:
I should add unit tests for this.

rosiezou (approved):
All tests passed. LGTM

PR #1174, opened 2022-04-20
chore: Check in service proto file

### LINKED ISSUE #1173: Check in match service definition
Check the service definition in

google/cloud/aiplatform/_matching_engine/match_service.proto

so it is easy to access

yan283:

Fixes #1173  🦕


ivanmkc:
The proto looks good but what was the rationale to not just add a match method for `IndexEndpointServiceClient`?

yan283:
Match is a method of MatchService, not IndexEndpointService.  I assume that means it should not be part of IndexEndpointServiceClient?

ivanmkc:
Sorry, I mean MatchService. Will there be a GAPIC version of MatchService?

yan283:
This is a short term solution to make it easy for clients to generate their own library.   We are looking into operations to publish GAPICs in the long run.

ivanmkc:
As discussed, please move to the _proto folder after https://github.com/googleapis/python-aiplatform/pull/1192 is merged.

ivanmkc:
Hi @yan283, the matching engine PR is merged so you can move the proto to the google/cloud/aiplatform/matching_engine/_protos folder.

ivanmkc (approved):
LGTM

PR #1183, opened 2022-04-26
chore: add tests for timeout arg not being explicitly set

sararob:
Added tests to ensure the default value of the timeout arg is set correctly. 

Fixes b/229967745 🦕


sasha-gitg (approved):
LGTM. I think all these files need their copyright years updated (won't block on it)

PR #1196, opened 2022-05-04
feat: add seq2seq forecasting training job

TheMichaelHu:

Fixes b/229909845 🦕

---

Adds a `SequenceToSequencePlusForecastingTrainingJob` to training jobs. This job has the exact same signature as `AutoMLForecastingTrainingJob`, but we are creating a separate job in case the two models diverge in the future.

The logic for `AutoMLForecastingTrainingJob` has been moved to a new abstract base class `_ForecastingTrainingJob`. The only things that differ between the seq2seq and automl training jobs that extend it are the `model_type` and `training_task_definition`.


TheMichaelHu:
TODO: split commit for better diff

google/cloud/aiplatform/training_jobs.py
```diff
@@ -4188,78 +4887,192 @@ def run(
...
+        """BigQuery location of exported evaluated examples from the Training Job
+        Returns:
+            str: BigQuery uri for the exported evaluated examples if the export
+                feature is enabled for training.
+            None: If the export feature was not enabled for training.
+        """
+        return super().evaluated_data_items_bigquery_uri
+
+
+class SequenceToSequenceForecastingTrainingJob(_ForecastingTrainingJob):
```
geraint0923, line 4956: SequenceToSequence**Plus**ForecastingTrainingJob

geraint0923:
Overall LG, just one minor comment.

geraint0923 (approved):
LGTM, thanks!

google/cloud/aiplatform/training_jobs.py
```diff
@@ -4453,110 +5265,53 @@ def _run(
...
-    @property
-    def _model_upload_fail_string(self) -> str:
-        """Helper property for model upload failure."""
-        return (
-            f"Training Pipeline {self.resource_name} is not configured to upload a "
-            "Model."
-        )
-
     @property
     def evaluated_data_items_bigquery_uri(self) -> Optional[str]:
```
ivanmkc, line 5316: You can just remove this if you call the super implementation.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -3886,7 +4597,9 @@ class column_data_types:
         REPEATED_TEXT = "repeated_text"
 
 
-class AutoMLForecastingTrainingJob(_TrainingJob):
+class AutoMLForecastingTrainingJob(_ForecastingTrainingJob):
+    _model_type = "AutoML"
+    _training_task_definition = schema.training_job.definition.automl_forecasting
     _supported_training_schemas = (schema.training_job.definition.automl_forecasting,)
 
     def __init__(
```
ivanmkc, line 4605: If you just call super, can't you just remove this __init__?

google/cloud/aiplatform/training_jobs.py
```diff
@@ -4000,15 +4714,6 @@ def __init__(
...
-        self._column_transformations = (
-            column_transformations_utils.validate_and_get_column_transformations(
-                column_specs, column_transformations
-            )
-        )
-
-        self._optimization_objective = optimization_objective
-        self._additional_experiments = []
-
     def run(
```
ivanmkc, line 4717: Can you just remove the `run` override and defer to super?

google/cloud/aiplatform/training_jobs.py
```diff
@@ -4188,78 +4893,192 @@ def run(
...
+            validation_options=validation_options,
+            model_display_name=model_display_name,
+            model_labels=model_labels,
+            additional_experiments=additional_experiments,
+            sync=sync,
+            create_request_timeout=create_request_timeout,
+        )
+
+    @property
+    def evaluated_data_items_bigquery_uri(self) -> Optional[str]:
```
ivanmkc, line 4952: Just remove this method and defer to super.

google/cloud/aiplatform/training_jobs.py
TheMichaelHu: Do you know if removing this would impact the documentation?

ivanmkc (changes requested):
It looks like you can simplify both AutoMLForecastingTrainingJob and SequenceToSequencePlusForecastingTrainingJob by removing all methods, since they are the exact same as the super's implementation.

We also need to add system tests for these. See tests/system/aiplatform/test_e2e_tabular.py for an example.

ivanmkc:
Regarding my last comment, you can simplify to:

```


class AutoMLForecastingTrainingJob(_ForecastingTrainingJob):
    _model_type = "AutoML"
    _training_task_definition = schema.training_job.definition.automl_forecasting
    _supported_training_schemas = (schema.training_job.definition.automl_forecasting,)

class SequenceToSequencePlusForecastingTrainingJob(_ForecastingTrainingJob):
    _model_type = "Seq2Seq"
    _training_task_definition = schema.training_job.definition.seq2seq_plus_forecasting
    _supported_training_schemas = (
        schema.training_job.definition.seq2seq_plus_forecasting,
    )
```

I ran the unit tests and it works fine. Otherwise, everything else looks pretty good.

google/cloud/aiplatform/training_jobs.py
ivanmkc: I think they can follow the superclass for docstrings.
Not sure if we use any automated tools that would be affected by this.

@sasha-gitg any thoughts? I would prefer less redundant code if possible.

TheMichaelHu:
Thanks for the review @ivanmkc! qq: how will simplifying the way you suggested impact the [documentation](https://googleapis.dev/python/aiplatform/latest/aiplatform.html?highlight=batchpredictionjob#google.cloud.aiplatform.AutoMLForecastingTrainingJob.run)? Is there a way to check that there will be no impact?

tests/unit/aiplatform/test_automl_forecasting_training_jobs.py
```diff
@@ -257,17 +257,25 @@ def teardown_method(self):
...
+        ],
+    )
     def test_run_call_pipeline_service_create(
         self,
         mock_pipeline_service_create,
         mock_pipeline_service_get,
         mock_dataset_time_series,
         mock_model_service_get,
         sync,
+        training_job,
```
ivanmkc, line 274: Move this above sync for consistency's sake.

TheMichaelHu: Can you elaborate why `training_job` should be above `sync`? I wanted to keep it aligned with the order `parameterize` was called, so should the order of the calls to `parameterize` be swapped too?

TheMichaelHu:
> We also need to add system tests for these. See tests/system/aiplatform/test_e2e_tabular.py for an example.

Should this occur in a separate PR since we never had e2e tests for Vertex Forecasting in the first place? If so, I can make a GitHub issue to track it and assign to myself.

google/cloud/aiplatform/training_jobs.py
TheMichaelHu: `nox -s docs` isn't creating documentation for `run()` as expected, since it's now part of the parent class. Looking into ways to force a child class to document a method from the parent class.

TheMichaelHu: Removed `evaluated_data_items_bigquery_uri`, moving discussion to the `run` method which I kept.

google/cloud/aiplatform/training_jobs.py
TheMichaelHu: If we remove this, Sphinx will not generate documentation for the `run()` function. It can pull the docstring from the parent class (tested with `nox -s docs`), but I couldn't find a way to make it document the function signature other than by writing it out manually.

tests/unit/aiplatform/test_automl_forecasting_training_jobs.py
```diff
@@ -489,16 +513,24 @@ def test_run_call_pipeline_if_no_model_display_name_nor_model_labels(
 
     @pytest.mark.usefixtures("mock_pipeline_service_get")
     @pytest.mark.parametrize("sync", [True, False])
+    @pytest.mark.parametrize(
```
ivanmkc, line 516: training_jobs.AutoMLForecastingTrainingJob and training_jobs.SequenceToSequencePlusForecastingTrainingJob will not necessarily share the same API forever. Prefer splitting into two separate tests.

tests/unit/aiplatform/test_automl_forecasting_training_jobs.py
```diff
@@ -420,16 +436,24 @@ def test_run_call_pipeline_service_create_with_timeout(
...
```
ivanmkc, line 439: Ditto. Split into two tests.

tests/unit/aiplatform/test_automl_forecasting_training_jobs.py
```diff
@@ -345,17 +353,25 @@ def test_run_call_pipeline_service_create(
...
         mock_pipeline_service_get,
         mock_dataset_time_series,
         mock_model_service_get,
         sync,
+        training_job,
     ):
         aiplatform.init(project=_TEST_PROJECT, staging_bucket=_TEST_BUCKET_NAME)
 
-        job = AutoMLForecastingTrainingJob(
+        job = training_job(
```
ivanmkc, line 374: Ditto.

tests/unit/aiplatform/test_automl_forecasting_training_jobs.py
ivanmkc: Okay I see, feel free to leave it as is.

tests/unit/aiplatform/test_automl_forecasting_training_jobs.py
ivanmkc: I'm torn. I'm also open to deferring this until it's needed. Feel free to decide.

TheMichaelHu: Let's leave it as-is for now. The intention is for Seq2Seq to share the same API as AutoML, but we're leaving the door open for implementation differences on the SDK side by splitting into two classes instead of just having a `model_type` argument in the `AutoMLForecastingTrainingJob`. At least that's my understanding.

ivanmkc: Sure, we can revisit later. Thanks.

google/cloud/aiplatform/training_jobs.py
ivanmkc: Hm, that's quite annoying. Let me ask the SDK team for more solutions.

ivanmkc (approved):
LGTM if system tests pass.

PR #1204, opened 2022-05-06
tests: Added VPC network constants to common file instead of continuous

ivanmkc:
After testing, it seems that common.cfg should work better.


dizcology (approved):

ivanmkc:
@sasha-gitg @sararob Should I merge or will this interfere with the release?

PR #1211, opened 2022-05-09
chore: Improved CPR local run auth flow.

abcdefgs0324:
- Provide utility methods to download artifacts from GCS.
- Change predictor interface from gcs_artifacts_uri to artifacts_uri.
- Allow users to provide local path for deploy_to_local_endpoint.

Will update all samples in another PR.


google/cloud/aiplatform/prediction/local_model.py
```diff
@@ -460,9 +461,24 @@ def deploy_to_local_endpoint(
 
         Args:
             artifact_uri (str):
-                Optional. The Cloud Storage path to the directory containing the Model artifact
-                and any of its supporting files. The AIP_STORAGE_URI environment variable will
-                be set to this uri if given; otherwise, an empty string.
+                Optional. The path to the directory containing the Model artifact and any of its
```
kthytang, line 464: General comment - 

Do we need the artifact_workdir param? Does the user need to specify this? (Can we just choose a default for the user?)

If artifact_uri is a GCS path - then it'll be passed directly to Predictor.load().
If artifact_uri is a local path - then we'll mount that path to a default path (chosen by us), and the mount path gets passed to Predictor.load().

Does that work?

kthytang (changes requested):

google/cloud/aiplatform/prediction/sklearn/predictor.py
```diff
@@ -28,18 +28,14 @@ class SklearnPredictor(Predictor):
...
+            artifacts_uri (str):
                 Required. The value of the environment variable AIP_STORAGE_URI.
         """
-        gcs_client = storage.Client()
-        with open("model.joblib", "wb") as model_f:
-            gcs_client.download_blob_to_file(
-                f"{gcs_artifacts_uri}/model.joblib", model_f
-            )
+        prediction_utils.download_model_artifacts(artifacts_uri)
         self._model = joblib.load("model.joblib")
```
kthytang, line 39: Does this need to be `joblib.load(os.path.join(artifacts_uri, "model.joblib"))`?

abcdefgs0324: This cannot be this if `artifacts_uri` is a gcs uri.

google/cloud/aiplatform/prediction/local_model.py
abcdefgs0324: I got rid of the artifact_workdir parameter. In download_model_artifacts, I copied the files to the current workdir too if it's a local path. Therefore users don't need to have two code paths in their `Predictor.load` (they can assume the files are in the current workdir). Let me know if you have any comments. Thanks!

google/cloud/aiplatform/prediction/sklearn/predictor.py
kthytang: Ah, nice, copying the local files to the current working directory works too.

google/cloud/aiplatform/docker_utils/utils.py
```diff
@@ -28,6 +28,9 @@
 
 Package = collections.namedtuple("Package", ["script", "package_path", "python_module"])
 Image = collections.namedtuple("Image", ["name", "default_home", "default_workdir"])
+DEFAULT_HOME = "/home"
+DEFAULT_WORKDIR = "/usr/app"
+DEFAULT_MOUNTED_MODEL_DIRECTORY = "/tmp_cpr_local_model"
```
kthytang, line 33: How about just "/model"

google/cloud/aiplatform/docker_utils/run.py
```diff
@@ -195,13 +216,13 @@ def run_prediction_container(
...
-    volumes = (
-        [f"{credential_path_on_host}:{credential_mount_path}"]
-        if credential_path
-        else []
-    )
-    envs[_ADC_ENVIRONMENT_VARIABLE] = credential_mount_path
+        credential_mount_path = _DEFAULT_CONTAINER_CRED_KEY_PATH
+        volumes = volumes + (
+            [f"{credential_path_on_host}:{credential_mount_path}"]
+            if credential_path
```
kthytang, line 222: Is this if needed? It's already in the `if credential_path:` block.

google/cloud/aiplatform/prediction/local_model.py
```diff
@@ -460,9 +460,14 @@ def deploy_to_local_endpoint(
...
+                supporting files. The path is either a GCS uri or the path to a local directory.
+                If this parameter is set to a GCS uri, you may need to specify `credential_path`.
+                If this parameter is set to a path to a local directory, the directory will be
+                mounted to the default temporary directory in the container.
```
kthytang, line 467:
```suggestion
                mounted to the default temporary model directory in the container.
```

google/cloud/aiplatform/prediction/local_model.py
```diff
@@ -460,9 +460,14 @@ def deploy_to_local_endpoint(
...
+                The AIP_STORAGE_URI environment variable will be set to this parameter if it's a
```
kthytang, line 468: I don't think we need to mention `AIP_STORAGE_URI` here. I don't think CPR users should need to know about AIP_STORAGE_URI. Maybe something like...

If this parameter is set to a GCS uri:
(1) `credential_path` must be specified for local prediction.
(2) The GCS uri will be passed directly to `Predictor.load`.
If this parameter is a local directory:
(1) The directory will be mounted to a default temporary model path.
(2) The mounted path will be passed to `Predictor.load`.
If this is a GCS path, it will be passed directly to Predictor.load. If it's a local path,

kthytang (changes requested):

google/cloud/aiplatform/prediction/local_model.py
kthytang: Cool, that works.

google/cloud/aiplatform/docker_utils/run.py
abcdefgs0324: you're right. Done.

google/cloud/aiplatform/docker_utils/utils.py
abcdefgs0324: I prefer using a non-general name in case users' images already have files there since we don't provide a way to let users customize it.

google/cloud/aiplatform/docker_utils/run.py
```diff
@@ -100,9 +101,16 @@ def run_prediction_container(
...
-            be set to this uri if given; otherwise, an empty string.
+            Optional. The path to the directory containing the Model artifact and any of its
+            supporting files. The path is either a GCS uri or the path to a local directory.
+            If this parameter is set to a GCS uri:
+            (1) `credential_path` must be specified for local prediction.
+            (2) The GCS uri will be passed directly to `Predictor.load`.
+            If this parameter is a local directory:
+            (1) The directory will be mounted to a default temporary model path.
+            (2) The mounted path will be passed to `Predictor.load`.
+            If this is a GCS path, it will be passed directly to `Predictor.load`. If it's a
```
kthytang, line 112: Oops sorry, my typo. Let's remove this line and the line below it.

kthytang (approved):

PR #1228, opened 2022-05-14
chore: Add a new way to deploy to local endpoint for CPR to interactively do local prediction.

abcdefgs0324:
Let's discuss whether or not to check in this..

### DIFF
--- a/google/cloud/aiplatform/prediction/local_endpoint.py
+++ b/google/cloud/aiplatform/prediction/local_endpoint.py
@@ -136,7 +136,7 @@ class LocalEndpoint:
             ValueError: If both of gpu_count and gpu_device_ids are set.
         """
         self.container = None
-        self.container_exited = False
+        self.container_exited = True
         self.log_start_index = 0
         self.serving_container_image_uri = serving_container_image_uri
         self.artifact_uri = artifact_uri
@@ -158,6 +158,9 @@ class LocalEndpoint:
 
         self.credential_path = credential_path
         self.host_port = host_port
+        # assigned_host_port will be updated according to the running container
+        # if host_port is None.
+        self.assigned_host_port = host_port
 
         self.gpu_count = gpu_count
         self.gpu_device_ids = gpu_device_ids
@@ -180,7 +183,33 @@ class LocalEndpoint:
         )
 
     def __enter__(self):
-        """Enters the runtime context related to this object.
+        """Enters the runtime context related to this object."""
+        try:
+            self.serve()
+        except Exception as exception:
+            _logger.error(f"Exception during entering a context: {exception}.")
+            raise
+        return self
+
+    def __exit__(self, exc_type, exc_value, exc_traceback):
+        """Exits the runtime context related to this object.
+
+        Args:
+            exc_type:
+                Optional. Class of the exception.
+            exc_value:
+                Optional. Type of the exception.
+            exc_traceback:
+                Optional. Traceback that has the information of the exception.
+        """
+        self.stop()
+
+    def __del__(self):
+        """Stops the container when the instance is about to be destroyed."""
+        self.stop()
+
+    def serve(self):
+        """Starts running the container and serves the traffic locally.
 
         An environment variable, GOOGLE_CLOUD_PROJECT, will be set to the project in the global config.
         This is required if the credentials file does not have project specified and used to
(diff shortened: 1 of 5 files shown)


google/cloud/aiplatform/prediction/local_endpoint.py
```diff
@@ -250,6 +251,109 @@ def __exit__(self, exc_type, exc_value, exc_traceback):
...
+            serving_container_environment_variables=serving_container_environment_variables,
+            serving_container_ports=serving_container_ports,
+            credential_path=credential_path,
+            host_port=host_port,
+            container_ready_timeout=container_ready_timeout,
+            container_ready_check_interval=container_ready_check_interval,
+        )
+        return local_endpoint.__enter__()
+
+    def undeploy(self):
```
samthrasher, line 377: I'd name this "stop()". Also, `__exit__` should call this, not the other way around.

google/cloud/aiplatform/prediction/local_model.py
```diff
@@ -438,6 +438,9 @@ def deploy_to_local_endpoint(
     ):
```
samthrasher, line 438: Since LocalEndpoint implements __enter__ and __exit__ methods anyway, we don't need the @contextmanager decorator on this function. (https://docs.python.org/3/library/contextlib.html#contextlib.contextmanager) 

It should just create and return the LocalEndpoint object instead of using the try/yield/finally pattern.

google/cloud/aiplatform/prediction/local_endpoint.py
```diff
@@ -250,6 +251,109 @@ def __exit__(self, exc_type, exc_value, exc_traceback):
...
         """
         self._stop_container_if_exists()
+        self.container_exited = True
+
+    def __del__(self):
+        """Stops the container when the instance is about to be destroyed."""
+        self.__exit__(None, None, None)
+
+    @classmethod
+    def deploy(
```
samthrasher, line 261: This should be an instance method. I'd also call it "serve" or "start" rather than "deploy".
Example usage:
```python
ep = local_model.deploy_to_local_endpoint()
ep.serve()
ep.predict()
ep.stop()
```

Also, the `serve` method should contain the logic currently in  `__enter__`, and `__enter__` should call this.

google/cloud/aiplatform/prediction/local_endpoint.py
```diff
@@ -136,6 +136,7 @@ def __init__(
             ValueError: If both of gpu_count and gpu_device_ids are set.
         """
         self.container = None
+        self.container_exited = False
```
samthrasher, line 139: Since we're managing the container state, maybe we want it to be an enum? What states can the container be in? What actions should be possible in each state?

google/cloud/aiplatform/prediction/local_model.py
```diff
@@ -522,6 +525,81 @@ def deploy_to_local_endpoint(
         finally:
             pass
 
+    def deploy_to_local_endpoint_lastingly(
```
samthrasher, line 528: This method shouldn't be needed. Users should call `local_model.deploy_to_local_endpoint()` to get an endpoint object, then `.serve()` on that object to start the container.

google/cloud/aiplatform/prediction/local_endpoint.py
```diff
@@ -250,6 +251,109 @@ def __exit__(self, exc_type, exc_value, exc_traceback):
                 Optional. Traceback that has the information of the exception.
         """
         self._stop_container_if_exists()
+        self.container_exited = True
+
+    def __del__(self):
+        """Stops the container when the instance is about to be destroyed."""
+        self.__exit__(None, None, None)
```
samthrasher, line 258: Should call `stop()` rather than `__exit__`.

samthrasher (changes requested):
Thanks for taking this on!

google/cloud/aiplatform/prediction/local_endpoint.py
abcdefgs0324: We are actually maintaining a running container or none. I think this is enough.

samthrasher (approved):

tests/unit/aiplatform/test_prediction.py
```diff
@@ -2167,38 +2052,41 @@ def test_init_with_gpu_capabilities_but_count_and_device_ids_unset(
...
+        wait_until_health_check_succeeds_mock,
     ):
-        _ = LocalEndpoint.deploy(_TEST_IMAGE_URI)
+        local_endpoint = LocalEndpoint(_TEST_IMAGE_URI)
 
-        local_endpoint_init_mock.assert_called_once_with(
-            serving_container_image_uri=_TEST_IMAGE_URI,
+        local_endpoint.serve()
+
+        run_prediction_container_mock.assert_called_once_with(
```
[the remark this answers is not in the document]
abcdefgs0324: Acknowledge.

Issue #1190, opened 2022-05-01
API redundancies and documentation


tsvisab:
I'm always frustrated when i have to pick an aiplatform job to run, this is because you have lots of API redundancies,
 For each job (e.g: `HyperparameterTuningJob`), you have at least three different versions:
 
[ Runnable job](https://github.com/googleapis/python-aiplatform/blob/c871e738a0fc3abe2d79c57b24b5b462a93cd2e9/google/cloud/aiplatform/__init__.py#L48)
and twice as Proto messages, (v1/v1_beta)
and these messages are being run with a service_client, right?
so, when and what should we use? which library are you supporting? 
what about this [lib](https://pypi.org/project/google-cloud-pipeline-components/)? 
i think you should have a single SDK that is well documented, it would help so much because for so many reasons,
some are:

- you'll have to work less
- examples will be more consistent and available

rosiezou:
Hello @tsvisab, thank you for reaching out. The reason behind API redundancies is that not everything in this repo is handwritten. Regarding documentation, adding more comprehensive documentation is on our to-do list and is something that we'll try to actively work on over the upcoming months. For the time being, here is how code generation and deployment works at a high-level:

- `google/cloud/aiplatform_v1` and `google/cloud/aiplatform_v1beta1` -- these two directories point to the GAPIC interface; GAPIC stands for Google API CodeGen, and we have automatic tools that generate code from our service protos. Please note that `v1` would contain GA features and `v1beta1` would contain GA + beta features.

- `google/cloud/aiplatform` -- this is where handwritten Vertex SDK APIs go. It builds on top of `aiplatform_v1` and integrates the `gapic` property that points to the corresponding GAPIC interface.

For your programmatic needs:

1. Look through `google/cloud/aiplatform` first -- Vertex SDK's APIs will almost always be easier to use and more concise comparing with GAPIC
2. If the feature that you are looking for cannot be found there, look through `aiplatform_v1` to see if it's available in GAPIC
3. If it is still in beta phase, it will be available in `platform_v1beta1`

If none of the above scenarios could help you find the right tools for your task, please feel free to open a github issue and send us a feature request.

PR #1236, opened 2022-05-17
chore: docs clarification

rosiezou:
A follow-up PR to address the question raised in #1190. I figured this might be a common question that more developers would have.


nayaknishant:
I'm not sure if the title should be "Overview". I think this paragraph is too detailed to be the overview of Vertex SDK and the description right under "Vertex AI SDK for Python" already gives a brief overview. Maybe a title like "Vertex AI SDK Features" or "Vertex AI SDK Features & Requests"? Just a thought. Also just some small grammatical fixes, "publicly" is more common than "publically" and from "GAPIC library code..." to "The GAPIC library code..."

nayaknishant (approved):

PR #1245, opened 2022-05-19
chore: add unit test mock for AnonymousCredentials

sararob:
I added a mock for `google.auth.credentials` so that the tests can fall back to that when running in an environment that doesn't have the `GOOGLE_APPLICATION_CREDENTIALS` env var set. This includes:
* The auth mock in a new module-scoped fixture in `tests/unit/aiplatform/conftest.py`
* Adding this mock to any test classes that require it

Fixes b/198162654 🦕

### DIFF
--- a/tests/unit/aiplatform/test_explain_saved_model_metadata_builder_tf2_test.py
+++ b/tests/unit/aiplatform/test_explain_saved_model_metadata_builder_tf2_test.py
@@ -26,7 +26,7 @@ from google.cloud.aiplatform.compat.types import explanation_metadata
 import test_models
 from test_models import upload_model_mock, get_model_mock  # noqa: F401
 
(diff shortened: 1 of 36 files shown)


tests/unit/aiplatform/test_explain_saved_model_metadata_builder_tf2_test.py
```diff
@@ -26,7 +26,7 @@
 import test_models
 from test_models import upload_model_mock, get_model_mock  # noqa: F401
 
-
+@pytest.mark.usefixtures("google_auth_mock")
```
sasha-gitg, line 29: Don't think this is needed here.

sasha-gitg (approved):

PR #1099, opened 2022-03-22
feat: add timeout arg across SDK

sararob:
Addresses b/225964107 🦕

I added a `timeout` arg and relevant tests to these methods:

* `Model.upload`
* `Model.deploy`
* `PipelineJob.submit`
*  `*TrainingJob.run`
*  `CustomJob.run`
*  `HyperparameterTuningJob.run`
* `BatchPredictionJob.create`
* `*Dataset.create`
* `*Dataset.import_data`
* `Tensorboard.create`
* `TensorboardExperiment.create`
* `TensorboardRun.create`
* `Featurestore.create`
* `Featurestore.create_entity_type`
* `Featurestore.update`
* `Featurestore.batch_serve_to_bq`
* `Featurestore.batch_serve_to_gcs`
* `Featurestore.batch_serve_to_df`
* `EntityType.create`
* `EntityType.update`
* `EntityType.read`
* `EntityType.create_feature`
* `EntityType.batch_create_features`
* `EntityType.ingest_from_bq`
* `EntityType.ingest_from_gcs`
* `EntityType.ingest_from_df`
* `Feature.create`
* `Feature.update`

For most of these methods, I gave the timeout arg a name specific to the method (i.e. `upload_request_timeout`) to differentiate the timeout of initiating the request (what this PR applies to) vs. the timeout of the underlying job.

PR #1248, opened 2022-05-21
feat: Add Vertex Forecasting E2E test.

TheMichaelHu:
Adds e2e test for AutoML Forecasting and unit test for `TimeSeriesDataset`. Also adds `create_request_timeout` to `TimeSeriesDataset`, which #1099 seems to have missed.

---


tests/system/aiplatform/test_dataset.py
```diff
@@ -306,6 +309,30 @@ def test_create_tabular_dataset_from_dataframe_with_provided_schema(
         finally:
             tabular_dataset.delete()
 
+    def test_create_time_series_dataset(self):
```
ivanmkc, line 312: Thanks for adding this

tests/system/aiplatform/test_e2e_forecasting.py
```diff
@@ -0,0 +1,116 @@
...
+    """End to end system test of the Vertex SDK with forecasting data."""
+
+    _temp_prefix = "temp-vertex-sdk-e2e-forecasting"
+
+    def test_end_to_end_forecasting(self, shared_state):
+        """Builds a dataset, trains models, and gets batch predictions."""
+
+        # Collection of resources generated by this test, to be deleted during
+        # teardown.
+        shared_state["resources"] = []
```
ivanmkc, line 43: Sorry, we're deprecating the use of `shared_state`. Can you put your test in a `try` block and do the cleanup in `finally`?

tests/system/aiplatform/test_e2e_forecasting.py
```diff
@@ -0,0 +1,116 @@
...
+
+        aiplatform.init(
+            project=e2e_base._PROJECT,
+            location=e2e_base._LOCATION,
+            staging_bucket=shared_state["staging_bucket_name"],
```
ivanmkc, line 44: You can keep this for now. We'll replace shared_state["staging_bucket_name"] later.

tests/system/aiplatform/test_dataset.py
```diff
@@ -51,6 +51,9 @@
 )
 _TEST_DATASET_DISPLAY_NAME = "permanent_50_flowers_dataset"
 _TEST_TABULAR_CLASSIFICATION_GCS_SOURCE = "gs://ucaip-sample-resources/iris_1000.csv"
+_TEST_FORECASTING_BQ_SOURCE = (
```
ivanmkc, line 54: I don't know if we want to add an external dependency to our system tests.

TheMichaelHu: Dataset doesn't validate anything about the data, so we could also reuse `_TEST_TABULAR_CLASSIFICATION_GCS_SOURCE`. It just might look like a bug.

ivanmkc: I copied the BQ data to `bq://ucaip-sample-tests:ucaip_test_us_central1`

I'm actually not sure if we should be using `gs://ucaip-sample-resources/iris_1000.csv` either to be honest.

tests/system/aiplatform/test_e2e_forecasting.py
```diff
@@ -0,0 +1,124 @@
...
+            assert (
+                automl_job.state
+                == pipeline_state.PipelineState.PIPELINE_STATE_SUCCEEDED
+            )
+            assert (
+                automl_batch_prediction_job.state
+                == job_state.JobState.JOB_STATE_SUCCEEDED
+            )
+        finally:
+            resources = [
```
ivanmkc, line 116: Prefer to avoid "locals" and just be explicit about the variables.
When we later integrate mypy, it might have trouble with the type inference.

ivanmkc: Just put the deletions on separate lines.

Also instantiate each variable to None before the `try` block.

tests/system/aiplatform/test_e2e_forecasting.py
```diff
@@ -0,0 +1,127 @@
...
+# limitations under the License.
+#
+
+from google.cloud import aiplatform
+from google.cloud.aiplatform.compat.types import job_state
+from google.cloud.aiplatform.compat.types import pipeline_state
+import pytest
+from tests.system.aiplatform import e2e_base
+
+_TRAINING_DATASET_BQ_PATH = (
```
ivanmkc, line 24: Can you update these URI's as well?

ivanmkc (approved):
LGTM

PR #1253, opened 2022-05-24
feat: add holiday regions for vertex forecasting

TheMichaelHu:
Add holiday regions for Vertex Forecasting.

---

Fixes b/232964228 🦕

### DIFF
--- a/google/cloud/aiplatform/training_jobs.py
+++ b/google/cloud/aiplatform/training_jobs.py
@@ -17,7 +17,7 @@
 
 import datetime
 import time
-from typing import Dict, List, Optional, Sequence, Tuple, Union
+from typing import Any, Dict, List, Optional, Sequence, Tuple, Union
 
 import abc
 
@@ -4607,6 +4607,7 @@ class AutoMLForecastingTrainingJob(_TrainingJob):
             "optimizationObjective": self._optimization_objective,
             "holidayRegions": holiday_regions,
         }
+        self._validate_training_task_inputs(training_task_inputs_dict)
 
         # TODO(TheMichaelHu): Remove the ifs once the API supports these inputs.
         if any(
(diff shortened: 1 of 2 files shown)


google/cloud/aiplatform/training_jobs.py
```diff
@@ -4549,6 +4564,97 @@ def _run(
...
+            training_task_inputs (Dict[str, Any]):
+                Required. The training task's input that corresponds to the
+                training_task_definition parameter.
+
+        Raises:
+            ValueError: If a training task input is invalid.
+        """
+        # TODO(TheMichaelHu): Validate all training task inputs.
+        if training_task_inputs.get("holidayRegions"):
+            valid_regions = {
```
TheMichaelHu, line 4581: Is there a better place to put this?

google/cloud/aiplatform/training_jobs.py
```diff
@@ -4201,9 +4202,14 @@ def run(
             create_request_timeout (float):
                 Optional. The timeout for the create request in seconds.
             sync (bool):
-                Whether to execute this method synchronously. If False, this method
+                Optional. Whether to execute this method synchronously. If False, this method
                 will be executed in concurrent Future and any downstream object will
                 be immediately returned and synced when the Future has completed.
+            holiday_regions (List[str]):
```
ivanmkc, line 4208: Can we add this above sync?

google/cloud/aiplatform/training_jobs.py
```diff
@@ -4503,7 +4516,9 @@ def _run(
             "quantiles": quantiles,
             "validationOptions": validation_options,
             "optimizationObjective": self._optimization_objective,
+            "holidayRegions": holiday_regions,
         }
+        self._validate_training_task_inputs(training_task_inputs_dict)
```
ivanmkc, line 4521: We typically leave validation to the service, so we don't have to maintain two sources of truth. Do you see an issue with this?

It may add a bit more wait time for the user but it vastly reduces complexity.

TheMichaelHu: I've seen validation for things like the splitting strategy: https://github.com/googleapis/python-aiplatform/blob/main/google/cloud/aiplatform/training_jobs.py#L521

I think it would be a better user experience to do some simple validation where we can, can't see how it would hurt. The wait time is fairly significant since input validation occurs in the training pipeline, not in the request handler, meaning we need to wait for resources to spin up and there is a resource cost associated with passing in poor inputs via the SDK.

TheMichaelHu: Also, do you know if there is a way to validate inputs against a given [training task definition](https://source.corp.google.com/piper///depot/google3/google/cloud/aiplatform/publicfiles/trainingjob/definition/automl_time_series_forecasting_1.0.0.yaml) on the SDK side? Seems like a fairly simple and reasonable thing to do.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -4200,10 +4201,23 @@ def run(
...
+                  Top level: GLOBAL
+                  Second level: continental regions
+                    NA: North America
+                    JAPAC: Japan and Asia Pacific
+                    EMEA: Europe, the Middle East and Africa
+                    LAC: Latin America and the Caribbean
+                  Third level: countries from ISO 3166-1 Country codes.
             sync (bool):
-                Whether to execute this method synchronously. If False, this method
+                Optional. Whether to execute this method synchronously. If False, this method
```
ivanmkc, line 4217: Good catch.

ivanmkc (approved):
LGTM

PR #1254, opened 2022-05-25
chore: refactor training job tests to configure temp test directory

sararob:
Refactored `test_training_jobs.py` to include a temp directory for the test Python script file. This will allow the unit tests to run on different environments.

### DIFF
--- a/tests/unit/aiplatform/test_training_jobs.py
+++ b/tests/unit/aiplatform/test_training_jobs.py
@@ -31,7 +31,6 @@ import tempfile
 import uuid
 from unittest import mock
 from unittest.mock import patch
-from absl.testing import absltest
 
 import test_training_jobs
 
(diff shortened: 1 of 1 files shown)


tests/unit/aiplatform/test_training_jobs.py
```diff
@@ -73,7 +74,8 @@
 _TEST_GCS_PATH = f"{_TEST_BUCKET_NAME}/{_TEST_GCS_PATH_WITHOUT_BUCKET}"
 _TEST_GCS_PATH_WITH_TRAILING_SLASH = f"{_TEST_GCS_PATH}/"
 _TEST_LOCAL_SCRIPT_FILE_NAME = "____test____script.py"
-_TEST_LOCAL_SCRIPT_FILE_PATH = f"path/to/{_TEST_LOCAL_SCRIPT_FILE_NAME}"
+_TEST_TEMPDIR = tempfile.mkdtemp()
+_TEST_LOCAL_SCRIPT_FILE_PATH = os.path.join(_TEST_TEMPDIR, _TEST_LOCAL_SCRIPT_FILE_NAME)
```
ucdmkt, line 78: Do you think you can afford dependency on `absl-py`? If, standardizing on tmp dir handling, with ability to inject via a public envvar, is benefitial.

https://github.com/abseil/abseil-py/blob/main/absl/testing/absltest.py#L158-L165


Alternatively, `pytest` also has a built in fixture to do so in a standard way, but I understand that `pytest` is used only for test driver, but not as a unittest utility framework as it is now, so it will be anyways additional dependency.


https://docs.pytest.org/en/6.2.x/tmpdir.html#base-temporary-directory

ucdmkt (approved):

jaycee-li (approved):
LGTM

PR #1255, opened 2022-05-25
feat: Add hierarchy and window configs to Vertex Forecasting training job

TheMichaelHu:
Adds support for hierarchical forecasting and window filtering.

---

Fixes b/229907889 b/228499154 🦕


google/cloud/aiplatform/training_jobs.py
```diff
@@ -4157,7 +4164,7 @@ def run(
                 Applies only if [export_evaluated_data_items] is True and
                 [export_evaluated_data_items_bigquery_destination_uri] is specified.
             quantiles (List[float]):
-                Quantiles to use for the `minimize-quantile-loss`
+                Quantiles to use for the ``minimize-quantile-loss``
```
ivanmkc, line 4167: Are the double back-ticks intentional?

google/cloud/aiplatform/training_jobs.py
```diff
@@ -4582,6 +4678,29 @@ def _add_additional_experiments(self, additional_experiments: List[str]):
...
+        configs = {
+            "column": column,
+            "stride_length": stride_length,
+            "max_count": max_count,
+        }
+        present_configs = {k: v for k, v in configs.items() if v is not None}
+        if not present_configs:
+            return None
+        if len(present_configs) > 1:
+            raise ValueError(
```
ivanmkc, line 4697: Again, we can keep this here if you think there's a really good reason to do validation on the client-side, but we prefer to defer to the service for validation.

ivanmkc (approved):
LGTM

google/cloud/aiplatform/training_jobs.py
TheMichaelHu: Yes, looking at the generated documentation, the double backtick formats a string as code instead of as a string surrounded by backticks.

google/cloud/aiplatform/training_jobs.py
TheMichaelHu: This is no longer validation of inputs, this is validation that we are passing the input to the API correctly.

TheMichaelHu: This is similar to the error the sdk prints when multiple split strategies are specified.

joannejqi (approved):

PR #1409, opened 2022-06-06
fix: change endpoint update method to return resource

sararob:
Fixes endpoint update method so it returns the resource and not an LRO.

Fixes b/235084784 🦕

### DIFF
--- a/google/cloud/aiplatform/models.py
+++ b/google/cloud/aiplatform/models.py
@@ -1377,7 +1377,7 @@ class Endpoint(base.VertexAiResourceNounWithFutureManager):
             self,
         )
 
-        self.api_client.update_endpoint(
+        self._gca_resource = self.api_client.update_endpoint(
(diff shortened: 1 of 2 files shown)


google/cloud/aiplatform/models.py
```diff
@@ -1377,21 +1377,17 @@ def update(
...
 
-        _LOGGER.log_action_started_against_resource_with_lro(
-            "Update", "endpoint", self.__class__, update_endpoint_lro
-        )
-
-        update_endpoint_lro.result()
-
         _LOGGER.log_action_completed_against_resource("endpoint", "updated", self)
 
+        self._sync_gca_resource()
```
sasha-gitg, line 1389: `self._gca_resource = self.api_client.update_endpoint(...` avoids the extra get call.

sasha-gitg (approved):

sasha-gitg:
Please update or add unit tests as well.

PR #1416, opened 2022-06-08
feat: support dataset update

### LINKED ISSUE #1178: Support dataset update
I believe dataset update is already on your roadmap given the skeleton here: https://github.com/googleapis/python-aiplatform/blob/075fa3bc16ef6d5b2fa158e2750ec0a31fd44e51/google/cloud/aiplatform/datasets/dataset.py#L600-L601

Where can the community track the timeline for that API? Thank you!

jaycee-li:

Fixes #1178  🦕

### DIFF
--- a/google/cloud/aiplatform/datasets/dataset.py
+++ b/google/cloud/aiplatform/datasets/dataset.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 
-# Copyright 2020 Google LLC
+# Copyright 2022 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -31,6 +31,7 @@ from google.cloud.aiplatform.compat.types import (
     io as gca_io,
 )
 from google.cloud.aiplatform.datasets import _datasources
+from google.protobuf import field_mask_pb2
 
 _LOGGER = base.Logger(__name__)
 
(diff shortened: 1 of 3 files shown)


google/cloud/aiplatform/datasets/dataset.py
```diff
@@ -597,8 +598,68 @@ def export_data(self, output_dir: str) -> Sequence[str]:
 
         return export_data_response.exported_files
 
-    def update(self):
-        raise NotImplementedError("Update dataset has not been implemented yet")
+    def update(
+        self,
```
sasha-gitg, line 602: Preference to make these keyword only arguments:

```Python
def update(
  self,
  *,
  display_name,
   ...)
```

google/cloud/aiplatform/datasets/dataset.py
```diff
@@ -597,8 +598,68 @@ def export_data(self, output_dir: str) -> Sequence[str]:
...
+            update_mask.paths.append("description")
+
+        update_dataset = gca_dataset.Dataset(
+            name=self.resource_name,
+            display_name=display_name,
+            description=description,
+            labels=labels,
+        )
+
+        self.api_client.update_dataset(
```
sasha-gitg, line 656: Preference to assign `_gca_resource` to skip the additional GET call below.

```Python
self._gca_resource = self.api_client.update_dataset(...)

return self
```

tests/system/aiplatform/test_dataset.py
```diff
@@ -350,3 +352,27 @@ def test_export_data(self, storage_client, staging_bucket):
         blob = bucket.get_blob(prefix)
 
         assert blob  # Verify the returned GCS export path exists
+
+    def test_update_dataset(self):
```
sasha-gitg, line 356: Preference for unit test as well.

sasha-gitg (changes requested):

tests/system/aiplatform/test_dataset.py
```diff
@@ -350,3 +352,27 @@ def test_export_data(self, storage_client, staging_bucket):
...
+                update_request_timeout=None,
+            )
+            labels.update(_TEST_DATASET_LABELS)
+
+            assert dataset.display_name == _TEST_DATASET_DISPLAY_NAME
+            assert dataset.labels == labels
+            assert dataset.gca_resource.description == _TEST_DATASET_DESCRIPTION
+
+        finally:
+            if dataset is not None:
```
sasha-gitg, line 377: This line is superfluous. `dataset` will only exist on assignment after `ImageDataset.create` is successful.

The test should start with `dataset=None`

sasha-gitg (approved):

tests/system/aiplatform/test_dataset.py
jaycee-li: If the `ImageDataset.create` is successful, then we want to delete the dataset finally. But if `create()` fails, dataset is None and we shouldn't call `dataset.delete()`. So I think it's necessary to add this line.

jaycee-li: If the `ImageDataset.create` is successful, then we want to delete the dataset finally. But if `create()` fails, dataset is None and we shouldn't call `dataset.delete()`. So I think it's necessary to add this line.

sasha-gitg: If create fails then `dataset`  is not defined and `if dataset is not None` will throw a `NameError`. `dataset` isn't created until after the right side of the statement, ` dataset = aiplatform.ImageDataset.create()` is successful. You can validate this behavior in Python:

```Python
def f():
    raise RuntimeError('')
    
try:
    dataset = f()
except:
    if dataset is None:
        pass
```

jaycee-li: Got it! 

But I see other system tests like `test_create_and_import_image_dataset` and `test_create_tabular_dataset` have the same pattern that checks `if dataset is not None`. Is there any reason for that?

sasha-gitg: No, those should be eventually fixed as well.

jaycee-li: Got it. Just removed this line. Thank you!

PR #1418, opened 2022-06-09
fix: Fixes error when neither column_specs or column_transformations are provided for AutoMLForecastTrainingJob

ivanmkc:
When neither column_specs or column_transformations are provided for AutoMLForecastTrainingJob,  calling `job.run` will fail with:
```
   2269 if self._column_transformations is None:
   2270     _LOGGER.info(
   2271         "No column transformations provided, so now retrieving columns from dataset in order to set default column transformations."
   2272     )
   2274     (
   2275         self._column_transformations,
   2276         column_names,
-> 2277     ) = dataset._get_default_column_transformations(target_column)
   2279     _LOGGER.info(
   2280         "The column transformation of type 'auto' was set for the following columns: %s."
   2281         % column_names
   2282     )
   2284 window_config = self._create_window_config(
   2285     column=window_column,
   2286     stride_length=window_stride_length,
   2287     max_count=window_max_count,
   2288 )

AttributeError: 'TimeSeriesDataset' object has no attribute '_get_default_column_transformations'
```

This PR fixes that and also removes the unneeded deprecation warning.


ivanmkc:
Unit tests should be added to cover this edge case for both AutoMLTabularTrainingJob and AutoMLForecastingTrainingJob.

Also, a test for supplying `column_specs` should be added.

These can be done in a separate PR.

PR #1420, opened 2022-06-11
chore: Generate CPR model servers at image building time and set the

abcdefgs0324:
default number of workers to the number of cores.


google/cloud/aiplatform/utils/prediction_utils.py
```diff
@@ -150,40 +149,173 @@ def populate_entrypoint_if_not_exists(
...
+
+    entrypoint_path = src_dir_path.joinpath(filename)
+    if entrypoint_path.exists():
+        _logger.info(
+            f'"{entrypoint_path.as_posix()}" already exists, skip '
+            f'generating "{filename}" in "{src_dir}".'
+        )
+        return
+
+    entrypoint_content = textwrap.dedent(
```
kthytang, line 290: If there's no templating with the entrypoint.py file, can we just include it in the SDK as its own file instead of generating it?

google/cloud/aiplatform/utils/prediction_utils.py
```diff
@@ -150,40 +149,173 @@ def populate_entrypoint_if_not_exists(
...
+                    error_message = "An exception {{}} occurred. Arguments: {{}}.".format(
+                        type(exception).__name__, exception.args
+                    )
+                    logging.info(
+                        "{{}}\\nTraceback: {{}}".format(error_message, traceback.format_exc())
+                    )
+
+                    # Converts all other exceptions to HTTPException.
+                    raise HTTPException(status_code=500, detail=error_message)
+        '''.format(
```
kthytang, line 247: I think post GA, we may want to convert this to use a templating language. I think it's harder to maintain like this since the file is long.

kthytang (changes requested):

google/cloud/aiplatform/prediction/local_model.py
```diff
@@ -199,8 +201,23 @@ def create_cpr_model(
...
             |   |-- ...
             |-- ...
+            |-- cpr_model_server.py
             |-- entrypoint.py
 
+        The created CPR images default the number of workers to the number of cores. If
+        you would like to change the number of workers, you could set the following environment
+        variables for the LocalModel instances:
+            WEB_CONCURRENCY:
+                The number of the workers. This will overwrite the calculated number from
```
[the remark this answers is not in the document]
abcdefgs0324: Updated.

google/cloud/aiplatform/utils/prediction_utils.py
abcdefgs0324: Do this for the following reasons.
1. If we include it in the SDK, we need to know the path to the model server which is in the user-provided directories at this moment so we need to know the directory names.
2. I think we should not check in the entrypoint which just serves as an entrypoint of created CPR images in the SDK. Users are not able to import it.

google/cloud/aiplatform/utils/prediction_utils.py
abcdefgs0324: Ack.

google/cloud/aiplatform/utils/prediction_utils.py
kthytang: Ah, I guess it'll be clunky to include it with the SDK and copy the source code into the container?

If so, can you file a bug to clean up the code for clearer templating for GA? https://b.corp.google.com/hotlists/4307152

google/cloud/aiplatform/prediction/local_model.py
```diff
@@ -163,8 +163,9 @@ def create_cpr_model(
         """Creates a local model from a custom predictor.
 
         This method creates a docker image to include user-provided predictor, and handler.
-        It populates the entrypoint, entrypoint.py, right under the specified directory, src_dir,
-        if it doesn't exist and generates a Dockerfile to build the image.
+        It populates the model server, cpr_model_server.py, and the entrypoint, entrypoint.py,
```
kthytang, line 166:
```suggestion
        It generates the model server, cpr_model_server.py, and the entrypoint, entrypoint.py,
        in the directory specified by, `src_dir`, if they don't already exist.
        
        This also generates a Dockerfile to build the image.
```

google/cloud/aiplatform/prediction/local_model.py
```diff
@@ -199,8 +201,23 @@ def create_cpr_model(
             |   |-- custom_package.tar.gz
             |   |-- ...
             |-- ...
+            |-- cpr_model_server.py
             |-- entrypoint.py
 
+        The created CPR images default the number of workers to the number of cores. If
```
kthytang, line 207:
```suggestion
        The created CPR images default the number of model server workers to the number of cores. Depending on the characteristics of your model, you may need to adjust the number of workers. You can set the number of workers with the following environment variables:
            WORKERS_PER_CORE:
                The number of the workers per core. The default is 1.
            MAX_WORKERS:
                The maximum number of workers can be used given the value of WORKERS_PER_CORE
                and the number of cores.
            WEB_CONCURRENCY:
                The number of the workers. This will overwrite the number of workers calculated by
                the other variables, min(WORKERS_PER_CORE * number_of_cores, MAX_WORKERS).
        See https://github.com/tiangolo/uvicorn-gunicorn-fastapi-docker/blob/master/README.md#workers_per_core.
```

kthytang (approved):

google/cloud/aiplatform/utils/prediction_utils.py
abcdefgs0324: Yes, if we include it with the SDK and copy the source code into the container, we will need to do something to find the files which may be different across users' environments. Have filed a bug to track this.

PR #1433, opened 2022-06-14
chore: performance improvements to unit tests

sararob:
So that our unit test suite runs faster, I've made global variables for job polling times in `jobs.py` and `training_jobs.py`. Then I've patched these in our job tests.


sasha-gitg (approved):
This seemed to reduce test time by 20 minutes. Nice work!

One option is to capture the testing wait times as a constant that can be tuned in a single location. `time.sleep` can take floats so we could also try a value like `0.1`

PR #1438, opened 2022-06-15
feat: Support for Model Versioning

samgoodman:


google/cloud/aiplatform/training_jobs.py
```diff
@@ -743,12 +778,27 @@ def _run_job(
             bigquery_destination=bigquery_destination,
         )
 
+        parent_model = models.ModelRegistry._get_true_version_parent(
+            parent_model=parent_model,
+            project=self.project,
+            location=self.location,
+        )
+
+        if model:
```
sasha-gitg, line 787: It seems like this should go into `_prepare_and_validate_run` because it's coupled with the creation of the model proto.

tests/system/aiplatform/test_model_version_management.py
```diff
@@ -0,0 +1,138 @@
+# -*- coding: utf-8 -*-
+
+# Copyright 2020 Google LLC
```
sasha-gitg, line 3: 2022

google/cloud/aiplatform/training_jobs.py
```diff
@@ -575,6 +575,11 @@ def _run_job(
         timestamp_split_column_name: Optional[str] = None,
```
sasha-gitg, line 575: update copyright to 2022

tests/system/aiplatform/test_model_version_management.py
```diff
@@ -0,0 +1,138 @@
...
+
+from google.cloud import aiplatform
+from google.cloud import storage
+from google.cloud.aiplatform.models import ModelRegistry
+
+from tests.system.aiplatform import e2e_base
+from tests.system.aiplatform import test_model_upload
+
+
+@pytest.mark.usefixtures("delete_staging_bucket")
```
sasha-gitg, line 31: "tear_down_resources" should also be included

sasha-gitg (changes requested):

google/cloud/aiplatform/models.py
```diff
@@ -1737,12 +1777,95 @@ def container_spec(self) -> Optional[aiplatform.gapic.ModelContainerSpec]:
...
+    @property
+    def version_aliases(self) -> Sequence[str]:
+        """User provided version aliases so that a model version can be referenced via
+        alias (i.e. projects/{project}/locations/{location}/models/{model_id}@{version_alias}
+        instead of auto-generated version id (i.e.
+        projects/{project}/locations/{location}/models/{model_id}@{version_id}).
+        The format is [a-z][a-zA-Z0-9-]{0,126}[a-z0-9] to distinguish from
+        version_id. A default version alias will be created for the first version
+        of the model, and there must be exactly one default version alias for a model."""
+        self._assert_gca_resource_is_available()
```
rosiezou, line 1798: Do these assertions raise a runtime error for the user? If not, I think it'll be a good idea to do so, along with a clear error message explaining that the model resource isn't available.

google/cloud/aiplatform/training_jobs.py
```diff
@@ -3073,6 +3171,36 @@ def _run(
                 Worker pools pecs required to run job.
             managed_model (gca_model.Model):
                 Model proto if this script produces a Managed Model.
+            model_id (str):
```
rosiezou, line 3174: I noticed duplicate doc strings throughout the files. For instance, the doc string for `is_default_version` parameter was duplicated 19 times. Would it be possible to simplify the doc strings & only have it show up at the top-level functions? eg. we can remove these doc strings from the `_run` method and only keep it in `run`.

google/cloud/aiplatform/models.py
samgoodman: Yes, its in base.py

google/cloud/aiplatform/models.py
```diff
@@ -3505,3 +3864,271 @@ def get_model_evaluation(
...
+        self.client = Model._instantiate_client(location, credentials)
+
+    def get_model(
+        self,
+        version: Optional[str] = None,
+    ) -> Model:
+        """Gets a registered model with optional version.
+
+        Args:
+            version (Optional[str], optional):
```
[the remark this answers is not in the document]
samgoodman: My docstring generator hates this, but will do

google/cloud/aiplatform/training_jobs.py
samgoodman: Preference is to leave it as is because _prepare_and_validate_run doesn't run for the AutoML training jobs and this keeps the versioning logic in one spot.

Open to pushback, however.

google/cloud/aiplatform/training_jobs.py
samgoodman: I'm not sure we can avoid this, sadly. All the run() and _run() methods in this class accept different args, so you can't easily do docstring inheritance or borrow docstrings from another method. I tried earlier.

Open to clever solutions if you have one.

rosiezou: In that case I think we should leave it as is for now. I have some concerns regarding the current way that the code base is organized as a whole, given that we have files that are ~2k-3k lines long and a large percentage of those lines of code are doc strings. But this has more to do with the continuous and ever-evolving discussion regarding best practice in programming, which is out of scope for this pull request.

google/cloud/aiplatform/models.py
```diff
@@ -3503,3 +3848,309 @@ def get_model_evaluation(
...
+        """Creates a versioned form of a model resource name.
+
+        Args:
+            resource_name (str): Required. A fully-qualified resource name or resource ID.
+            version (str): Required. The version or alias of the resource.
+
+        Returns:
+            versioned_name (str): The versioned resource name in revisioned format.
+        """
+        if version:
```
sasha-gitg, line 4072: This is required in the method signature.

google/cloud/aiplatform/training_jobs.py
sasha-gitg: Confirming this was resolved.

google/cloud/aiplatform/training_jobs.py
sasha-gitg: SGTM

google/cloud/aiplatform/models.py
```diff
@@ -1760,15 +1871,41 @@ def __init__(
...
         super().__init__(
             project=project,
             location=location,
             credentials=credentials,
             resource_name=model_name,
         )
-        self._gca_resource = self._get_gca_resource(resource_name=model_name)
+
+        # Model versions can include @{version} in the resource name.
+        self._resource_id_validator = super()._revisioned_resource_id_validator
```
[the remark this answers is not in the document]
samgoodman: The "revisioned resource" format isn't specific to versioned models, it's a format available for all APIs. It just so happens that Models are the first resources in the SDK to make use of this format.

See https://google.aip.dev/162

My preference is to keep the revisioned resource ID validator in base for the next resource type that uses the revision format, lest it need refactoring to decouple it from Model resources.

google/cloud/aiplatform/models.py
```diff
@@ -3503,3 +3848,309 @@ def get_model_evaluation(
...
+                format_resource_name_method=Model._format_resource_name,
+                project=project,
+                location=location,
+                resource_id_validator=base.VertexAiResourceNoun._revisioned_resource_id_validator,
+            )
+
+        self.credentials = credentials or (
+            model.credentials if isinstance(model, Model) else None
+        )
+        self.client = Model._instantiate_client(location, self.credentials)
```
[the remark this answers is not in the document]
samgoodman: Resolved. Instead of self.credentials being None, self.credentials will be set to the default credentials. Models retrieved from the registry will use the saved self.credentials, even if the default creds change after MR creation.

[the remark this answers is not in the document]
sasha-gitg: LGTM

sasha-gitg (approved):

PR #1439, opened 2022-06-16
samples: Updated Prediction CPR notebooks.

abcdefgs0324:
- Use the specific version of uvicorn.
- Provide sample usages for local run with local paths and use
  download_model_artifacts function.
- Set up logging level.
- Change description to that model server is generated by SDK.
- Remove SDK_Custom_Predict notebook.


samples/notebooks/prediction/SDK_Custom_Predict_SDK_Integration.ipynb
```diff
@@ -1078,11 +1123,100 @@
...
-    "Note: You need to have the credentials set up in the previous step and pass the path to the credentials while running the container. The service account should have the **Storage Object Viewer** permission.\n",
-    "\n",
     "In this example, the container executes a prediction request and a health check."
    ]
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "**Option 1. Pass local paths**\n",
```
kthytang, line 1133: Since printing the responses and viewing logs are the same for both options, can we structure this section to be...

```
Deploy your CPR container locally and send some requests.

Option 1. Deploy locally using a local path for `artifacts_uri`
[code]
Option 2. Deploy locally using a GCS path for `artifacts_uri`
[code]

Print out the health check response and its content...

Print out the predict response and its content...

Also print out all the container logs...
```

Also, can you change the code snippet to send a health check first, before the predict request?

kthytang (changes requested):
- Also, can you change line 738 to add the word "typically":
  - To build a custom container, we *typically* also need to write an entrypoint of the image that starts the model server. 
- Change create_cpr_model to build_cpr_model

abcdefgs0324:
> Oh I see that the build_cpr_model changes are in PR #1449 so it's ok to remove them from here.

I remove them in order to check in this PR first. Thanks!

PR #1444, opened 2022-06-19
chore: Support bigquery 3.x

kweinmeister:
`google-cloud-bigquery` is now on v3.2.0. Opening up the range of allowed dependencies.


kweinmeister:
With this change, a couple system tests related to Feature Store are failing, due to expecting a timestamp rather than datetime. FYI @sasha-gitg @rosiezou

rosiezou (approved):
manual-triggered system test passed. approved changes

PR #1449, opened 2022-06-21
chore: Merged LocalModel.create into init and removed upload method from LocalModel class.

abcdefgs0324:


google/cloud/aiplatform/models.py
```diff
@@ -65,6 +65,181 @@
 ]
 
 
+class ModelMixin:
+
```
samgoodman, line 69: Please add a docstring explaining the purpose of the mixin

google/cloud/aiplatform/models.py
```diff
@@ -65,6 +65,181 @@
...
+            staging_bucket (str):
+                Optional. Bucket to stage local model artifacts. Overrides
+                staging_bucket set in aiplatform.init.
+            appended_user_agent (List[str]):
+                Optional. User agent appended in the client info. If more than one, it will be
+                separated by spaces.
+            upload_request_timeout (float):
+                Optional. The timeout for the upload request in seconds.
+    """
+        pass
```
samgoodman, line 240: Throw instead of pass, make it clear that this method must be overridden

google/cloud/aiplatform/prediction/local_model.py
```diff
@@ -118,35 +108,43 @@ def create(
...
-                for key, value in serving_container_environment_variables.items()
-            ]
-        if serving_container_ports:
-            ports = [
-                gca_model_compat.Port(container_port=port)
-                for port in serving_container_ports
-            ]
+        if serving_container_spec:
+            if not serving_container_spec.image_uri:
+                raise ValueError(
```
samgoodman, line 114: Add a "Raises" docstring section for this error

google/cloud/aiplatform/prediction/local_model.py
```diff
@@ -118,35 +108,43 @@ def create(
...
-            args=serving_container_args,
-            env=env,
-            ports=ports,
-            predict_route=serving_container_predict_route,
-            health_route=serving_container_health_route,
-        )
+            self.serving_container_spec = serving_container_spec
+        else:
+            if not serving_container_image_uri:
+                raise ValueError(
```
samgoodman, line 121: Add "Raises" docstring for this error

google/cloud/aiplatform/prediction/local_model.py
```diff
@@ -118,35 +108,43 @@ def create(
...
+                command=serving_container_command,
+                args=serving_container_args,
+                env=env,
+                ports=ports,
+                predict_route=serving_container_predict_route,
+                health_route=serving_container_health_route,
+            )
 
     @classmethod
     def create_cpr_model(
```
samgoodman, line 150: This can be moved into __init__ as well, "create" is a reserved verb

google/cloud/aiplatform/prediction/local_model.py
```diff
@@ -261,7 +259,7 @@ def create_cpr_model(
             health_route=DEFAULT_HEALTH_ROUTE,
         )
 
-        return cls(container_spec)
+        return cls(serving_container_spec=container_spec)
 
     def upload(
```
samgoodman, line 264: This doesn't use the mixin, the problem with two competing `upload` methods still exists.

If you're going the route of giving the mixin's `upload` method no implementation at all, then make the mixin an abstract class to enforce that it's method signature gets used. `LocalModel.upload` can call `LocalModel._upload` to make use of instance properties if necessary.

samgoodman (changes requested):

google/cloud/aiplatform/prediction/local_model.py
abcdefgs0324: I think moving this method into `init` is confusing because `init` will become pretty large, e.g. which parameters go first, which group of parameters should be set together, etc. Could we just rename this method but keep it as a separate method if "create" is a reserved verb?

samgoodman: That works

google/cloud/aiplatform/prediction/local_model.py
abcdefgs0324: Done.

google/cloud/aiplatform/prediction/local_model.py
abcdefgs0324: Done.

google/cloud/aiplatform/prediction/local_model.py
abcdefgs0324: Done.

google/cloud/aiplatform/prediction/local_model.py
abcdefgs0324: Removed upload method from LocalModel.

google/cloud/aiplatform/models.py
```diff
@@ -1861,6 +1875,7 @@ def upload(
         serving_container_args: Optional[Sequence[str]] = None,
         serving_container_environment_variables: Optional[Dict[str, str]] = None,
         serving_container_ports: Optional[Sequence[int]] = None,
+        local_model: Optional["LocalModel"] = None,
```
kthytang, line 1878: Should this be `Optional[LocalModel]` instead?

google/cloud/aiplatform/models.py
```diff
@@ -2032,6 +2048,10 @@ def upload(
             ValueError: If only `explanation_metadata` or `explanation_parameters`
                 is specified.
                 Also if model directory does not contain a supported model file.
+                If `local_model` is specified but `serving_container_spec.image_uri`
+                in the `local_model` is None.
+                If `local_model` is not specified but `serving_container_image_uri`
```
kthytang, line 2053:
```suggestion
                If `local_model` is not specified and `serving_container_image_uri`
```

google/cloud/aiplatform/models.py
```diff
@@ -2044,32 +2064,44 @@ def upload(
...
+            if not container_spec.image_uri:
+                raise ValueError(
+                    "If `local_model` is specified, `serving_container_spec.image_uri` in the `local_model` "
+                    "is required."
+                )
+            appended_user_agent = [prediction_constants.CUSTOM_PREDICTION_ROUTINES]
+        else:
+            if not serving_container_image_uri:
+                raise ValueError(
+                    "The parameter `serving_container_image_uri` is required."
```
kthytang, line 2079: Did this check not exist before?

kthytang (approved):

google/cloud/aiplatform/models.py
```diff
@@ -1934,6 +1949,10 @@ def upload(
                 no impact on whether the port is actually exposed, any port listening on
                 the default "0.0.0.0" address inside a container will be accessible from
                 the network.
+            local_model (Optional[LocalModel]):
+                Optional. A LocalModel instance which includes serving_container_spec.
```
samgoodman, line 1953:
```suggestion
                Optional. A LocalModel instance that includes a `serving_container_spec`.
```

google/cloud/aiplatform/models.py
```diff
@@ -1934,6 +1949,10 @@ def upload(
...
+                If it is specified, the serving container spec of the LocalModel instance
```
samgoodman, line 1954:
```suggestion
                If provided, the `serving_container_spec` of the LocalModel instance
```

google/cloud/aiplatform/models.py
samgoodman, line 2079:
```suggestion
                    "The parameter `serving_container_image_uri` is required if no `local_model` is provided."
```

google/cloud/aiplatform/models.py
```diff
@@ -2044,32 +2064,44 @@ def upload(
...
-            ]
-        if serving_container_ports:
-            ports = [
-                gca_model_compat.Port(container_port=port)
-                for port in serving_container_ports
-            ]
+        appended_user_agent = None
+        if local_model:
+            container_spec = local_model.get_serving_container_spec()
+            if not container_spec.image_uri:
```
samgoodman, line 2070: container_spec.image_uri doesn't get used in this code. Let the code that does use container_spec.image_uri throw if it isn't present. Remove this check.

samgoodman (changes requested):

google/cloud/aiplatform/models.py
```diff
@@ -48,8 +57,13 @@
     env_var as gca_env_var_compat,
 )
 
+from google.cloud.aiplatform.constants import prediction as prediction_constants
+
 from google.protobuf import field_mask_pb2, json_format
 
+if TYPE_CHECKING:
```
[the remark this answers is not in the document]
abcdefgs0324: The modules imported by LocalModel import CPR extra packages such as fastapi, uvicorn, and docker which require users to install SDK with prediction extra packages.
Without this, any users who import any modules in the SDK need to install CPR extra packages since Model is in google/cloud/aiplatform/__init__.py. I want to avoid CPR extra packages becoming hard dependencies.

google/cloud/aiplatform/models.py
abcdefgs0324: Done.

google/cloud/aiplatform/models.py
abcdefgs0324: Done.

samgoodman (approved):

PR #1452, opened 2022-06-22
feat: add support for failure_policy in PipelineJob

chongyouquan:


tests/unit/aiplatform/test_pipeline_jobs.py
```diff
@@ -621,6 +622,99 @@ def test_run_call_pipeline_service_create_with_timeout_not_explicitly_set(
...
+                pipeline_failure_policy.PipelineFailurePolicy.PIPELINE_FAILURE_POLICY_FAIL_SLOW,
+            ),
+            (
+                "fast",
+                pipeline_failure_policy.PipelineFailurePolicy.PIPELINE_FAILURE_POLICY_FAIL_FAST,
+            ),
+        ],
+    )
+    @pytest.mark.parametrize("sync", [True, False])
+    def test_run_call_pipeline_service_create_with_failure_policy(
```
sararob, line 643: Is there a way to test the result of setting `failure_policy` to fast vs. slow in a completed PipelineJob, or do you think we should leave this type of test to the service team?

google/cloud/aiplatform/pipeline_jobs.py
```diff
@@ -173,6 +174,15 @@ def __init__(
...
                 location set in aiplatform.init will be used.
+            failure_policy (str):
+                Optional. The failure policy - "slow" or "fast".
+                Currently, the default of a pipeline is that the pipeline will continue to
+                run until no more tasks can be executed, also known as
+                PIPELINE_FAILURE_POLICY_FAIL_SLOW (corresponds to "slow").
+                However, if a pipeline is set to
+                PIPELINE_FAILURE_POLICY_FAIL_FAST (corresponds to "fast"),
+                it will stop scheduling any new tasks when a task has failed. Any
+                scheduled tasks will continue to completion.
```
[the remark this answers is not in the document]
chongyouquan: If `failure_policy` is set to "fast", then, **if a task fails**, components that have started running will continue to run, but any components that haven't started executing won't be run.

tests/unit/aiplatform/test_pipeline_jobs.py
chongyouquan: I think we should leave the test to the service team.. They have tests on the backend that would make sure that it is working as intended.

sararob (changes requested):
One minor comment, once that's addressed this lgtm.

PR #1455, opened 2022-06-23
chore: update featurestore test to fail gracefully on import error

sararob:
Skips `test_batch_serve_to_df` if the BQ storage library is not installed where the test is running.

### DIFF
--- a/tests/unit/aiplatform/test_featurestores.py
+++ b/tests/unit/aiplatform/test_featurestores.py
@@ -52,10 +52,16 @@ from google.cloud.aiplatform.compat.types import (
 )
 
 from google.cloud import bigquery
-from google.cloud import bigquery_storage
-from google.cloud import resourcemanager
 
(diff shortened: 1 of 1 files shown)


tests/unit/aiplatform/test_featurestores.py
```diff
@@ -1603,6 +1609,9 @@ def test_batch_serve_to_gcs_with_invalid_gcs_destination_type(self):
                 read_instances_uri=_TEST_GCS_CSV_SOURCE_URI,
             )
 
+    @pytest.mark.skipif(
+        _USE_BQ_STORAGE is False, reason="batch_serve_to_df requires bigquery_storage"
```
ucdmkt, line 1613: (optional)

Another way to achieve this is `"bigquery_storage" not in  sys.modules", ...`

Up to your preference, since `sys` was not imported before.

PR #1465, opened 2022-06-28
feat: Add support for start_execution in MLMD SDK. 

SinaChavoshi:


google/cloud/aiplatform/metadata/schema/base_execution.py
```diff
@@ -112,3 +113,67 @@ def create(
...
+        ```
+        Args:
+            metadata_store_id (str):
+                Optional. The <metadata_store_id> portion of the resource name with
+                the format:
+                projects/123/locations/us-central1/metadataStores/<metadata_store_id>/executions/<resource_id>
+                If not provided, the MetadataStore's ID will be set to "default". Currently only the 'default'
+                MetadataStore ID is supported.
+            resume (bool):
+                Resume an existing execution. If resume set to `True`, resource_id must be provided.
```
jialuzh, line 147: Looks like we do not need user to provide resource_id in logic below, do we need to update the comment

jialuzh (approved):

google/cloud/aiplatform/metadata/schema/base_execution.py
SinaChavoshi: Done.

PR #1476, opened 2022-07-06
feat: add get_associated_experiment method to pipeline_jobs

sararob:
Added a `get_associated_experiment` method to `PipelineJob` to return the Experiment associated with the provided pipeline job.

### DIFF
--- a/google/cloud/aiplatform/pipeline_jobs.py
+++ b/google/cloud/aiplatform/pipeline_jobs.py
@@ -786,11 +786,17 @@ class PipelineJob(
             self._gca_resource.job_detail.pipeline_run_context.parent_contexts
         )
 
-        pipeline_experiment_resource_names = [
-            c
+        pipeline_experiment_resources = [
+            context._Context(resource_name=c)._gca_resource
             for c in pipeline_parent_contexts
             if c != self._gca_resource.job_detail.pipeline_context.name
         ]
+
+        pipeline_experiment_resource_names = []
+
+        for c in pipeline_experiment_resources:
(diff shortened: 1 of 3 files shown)


google/cloud/aiplatform/pipeline_jobs.py
```diff
@@ -776,3 +776,31 @@ def clone(
         )
 
         return cloned
+
+    def get_associated_experiment(self) -> Optional["experiment_resources.Experiment"]:
```
sasha-gitg, line 780:
```suggestion
    def get_associated_experiment(self) -> Optional["aiplatform.Experiment"]:
```

tests/unit/aiplatform/test_pipeline_jobs.py
```diff
@@ -29,6 +29,10 @@
 from google.cloud import aiplatform
```
sasha-gitg, line 29: Please also update the experiment system tests to test this functinality. Add a test right after this test: https://github.com/googleapis/python-aiplatform/blob/main/tests/system/aiplatform/test_experiments.py#L263

It should be able to get the `pipeline_job` using `job = aiplatform.PipelineJob(self._pipeline_job_id` and then verify the associated experiment is `self._experiment_name`.

sasha-gitg (changes requested):

tests/unit/aiplatform/test_pipeline_jobs.py
sararob: I added a call to `get_associated_experiment` to `test_add_pipeline_job_to_experiment`. Just kicked off our system tests from my branch.

PR #1477, opened 2022-07-06
chore: Fixed CPR GCS bug, allowed local util logging to set encoding, and removed limitation on predictor init method.

abcdefgs0324:


google/cloud/aiplatform/docker_utils/local_util.py
```diff
@@ -23,7 +23,9 @@
 _logger = logging.getLogger(__name__)
 
 
-def execute_command(cmd: List[str], input_str: Optional[str] = None) -> int:
+def execute_command(
+    cmd: List[str], input_str: Optional[str] = None, encoding="utf-8", errors=None
```
kthytang, line 27: How's the newly introduced encoding param used? (Don't see the caller of this updated)

kthytang, line 27: Maybe rename to output_encoding to avoid confusion? Since there's input encoding too.

kthytang (changes requested):

tests/unit/aiplatform/test_docker_utils.py
```diff
@@ -676,6 +676,19 @@ def test_build_image(self, make_dockerfile_mock, execute_command_mock):
             pip_command=self.PIP,
             python_command=self.PYTHON,
         )
+        execute_command_mock.assert_called_once_with(
```
kthytang, line 679: Is this change to the unit tests still useful?

kthytang (approved):

tests/unit/aiplatform/test_docker_utils.py
abcdefgs0324: yes, so we make sure the local_utils.execute_command is called as expected. I should have added this in the first place.

PR #1483, opened 2022-07-08
feat: Support custom containers in CustomJob.from_local_script

jaycee-li:


google/cloud/aiplatform/jobs.py
```diff
@@ -1403,6 +1406,29 @@ def from_local_script(
...
+                    + f"\npip3 install -q --user {package_gcs_uri[len(staging_bucket)+1:]}"
+                    + f"\npython3 -m {python_packager.module_name}",
+                ]
+
+                spec["container_spec"] = {
+                    "image_uri": container_uri,
+                    "command": command,
+                }
+
+                if args:
```
sasha-gitg, line 1424: Reuse `args` and `environment_variables` setting logic shared with the pre-built container condition.

google/cloud/aiplatform/jobs.py
```diff
@@ -1403,6 +1406,29 @@ def from_local_script(
                         {"name": key, "value": value}
                         for key, value in environment_variables.items()
                     ]
+            else:
```
sasha-gitg, line 1409: Docstrings need to be updated explaining this behavior and the customer container requirements.

tests/unit/aiplatform/test_custom_job.py
```diff
@@ -50,6 +50,7 @@
 _TEST_ENABLE_WEB_ACCESS = True
 _TEST_WEB_ACCESS_URIS = {"workerpool0-0": "uri"}
 _TEST_TRAINING_CONTAINER_IMAGE = "gcr.io/test-training/container:image"
+_TEST_PREBUILT_CONTAINER_IMAGE = "gcr.io/cloud-aiplatform/container:image"
```
sasha-gitg, line 53: Please add system test as well.

sasha-gitg (changes requested):

google/cloud/aiplatform/jobs.py
jaycee-li: They have different spec names (`spec['container_spec']` vs `spec['python_package_spec']`). I am not sure how can I share the setting logic?

google/cloud/aiplatform/jobs.py
```diff
@@ -1265,7 +1265,12 @@ def from_local_script(
             script_path (str):
                 Required. Local path to training script.
             container_uri (str):
-                Required: Uri of the training container image to use for custom job.
+                Required. Uri of the training container image to use for custom job.
+                Support images in Artifact Registry, Container Registry, or Docker Hub.
+                Vertex AI provides a wide range of executor images with pre-installed
+                packages to meet users' various use cases. See the list of `pre-built containers
+                for training <https://cloud.google.com/vertex-ai/docs/training/pre-built-containers>`.
+                If not using image from this list, please make sure python3 and pip3 are installed in your container.
```
sararob, line 1273: If python3 and pip3 aren't installed on a custom container, will the training service or the SDK throw an error? If it's the SDK, could you add some error handling for this case? Otherwise you can leave as is.

sararob (changes requested):

google/cloud/aiplatform/jobs.py
jaycee-li: Currently neither of them will throw the error if python3 and pip3 aren't installed. But `job.run()` will fail and there will be a link to the log, where users can see the reason of failure is `pip not found`. 

If we want to check this error in SDK, we will need to pull and run the docker image inside `from_local_script` method, which will take several minutes or more to finish. And that means the time to initialize a customJob goes from seconds to minutes.

Given that this error is unlikely to happen(customers use SDK means they are python users, so normally the container image they choose to use will have python installed), I think it's not necessary to add this error handling. What do you think?

sararob: Yes agreed, in that case let's not raise an error in the SDK.

sararob (approved):

SinaChavoshi (approved):

PR #1487, opened 2022-07-11
feat: Allow listing of model versions via Model.list when a resource name is provided.

samgoodman:

Fixes b/237795227


google/cloud/aiplatform/models.py
```diff
@@ -3591,6 +3591,7 @@ def list(
         order_by: Optional[str] = None,
         project: Optional[str] = None,
         location: Optional[str] = None,
+        model_name: Optional[str] = None,
```
sasha-gitg, line 3594: This implementation doesn't support the filter or order_by arguments. 

@andrewferlitsch To confirm this is the functionality you were expecting. My interpretation of your request, based on the usage in your comment, is that `Model.list()` should be able to provide all model versions, not necessarily scoped to the model resource name.

samgoodman: I'm open to reinterpretation of the objective for your second point.

As to the first point, the MR backend doesn't currently support filter or oder args for version lists. We can do this as a post-processing step, but that may stretch beyond the scope of the SDK

sasha-gitg: @andrewferlitsch pinging on this comment, thanks!

PR #1514, opened 2022-07-18
feat: Add metadata SDK samples for list artifact and list execution

SinaChavoshi:


samples/model-builder/conftest.py
```diff
@@ -583,6 +583,13 @@ def mock_create_execution(mock_execution):
         yield mock_create_execution
 
 
+@pytest.fixture
+def mock_list_execution(mock_execution):
+    with patch.object(aiplatform.Execution, "list") as mock_list_execution:
+        mock_list_execution.return_value = [mock_execution]
```
jialuzh, line 589: Do we need to return more than one executions to better mimic real world data?

SinaChavoshi: Done.

nayaknishant (approved):

PR #1524, opened 2022-07-20
feat: Added the PipelineJob.from_component_url method

Ark-kun:
Example: 

```python
PipelineJob.from_component_url(
    url="https://..../Train_TF_model/component.yaml",
    arguments=dict(
        training_data_uri="gs://.../data.csv",
        output_model_uri="gs://.../new_model",
    ),
).submit()
```

I wonder what's the best way to support components from URLs and local files.

Option 1: Separate methods:
```python
PipelineJob.from_component_url(url=...)
PipelineJob.from_component_file(path=...)
```

Option 2: Single method:
```python
PipelineJob.from_component(url=...)
PipelineJob.from_component(path=...)
```

I slightly prefer option 1 since I find it confusing where the function has optional parameters that are required (and also cannot be set at the same time): (e.g. both url and path are optional, but exactly one of them must be set).


nayaknishant:
hey @Ark-kun will these methods require unit tests?

jaycee-li (commented):
What about `PipelineJob.from_component(component: str, ...)` where `component` can be either url or file_path, and inside the method we check whether the input is url or path?

Ark-kun:
> hey @Ark-kun will these methods require unit tests?

I think that they can only be tested by integration tests.
Can integration tests access external URLs?

sasha-gitg:
> I think that they can only be tested by integration tests.
> Can integration tests access external URLs?

Yes

chensun:
The proposed feature could be covered by `template_path` from the constructor if it's not covered already:

https://github.com/googleapis/python-aiplatform/blob/944b03fc9bd7f1d69f6852c65286b72f49f425df/google/cloud/aiplatform/pipeline_jobs.py#L128-L132

I would suggest we don't create "multiple ways" to do the same or similar task. 
That being said, for non-trivial user interface changes like such, we should always start with a design review first. 

/hold

Ark-kun:
> The proposed feature could be covered by `template_path` from the constructor if it's not covered already:

Thank you for the review, Chen.
I think there is a misunderstanding.
You're correct that running *pipelines* is already supported by `PipelineJob(template_path=)`.
However this PR is about running **components**, not **pipelines**.

See an example:

```python
PipelineJob.from_component_url(
    url="https://..../Train_TF_model/component.yaml",
    arguments=dict(
        training_data_uri="gs://.../data.csv",
        output_model_uri="gs://.../new_model",
    ),
).submit()
```

What do you think?

chensun:
> > The proposed feature could be covered by `template_path` from the constructor if it's not covered already:
> 
> Thank you for the review, Chen. I think there is a misunderstanding. You're correct that running _pipelines_ is already supported by `PipelineJob(template_path=)`. However this PR is about running **components**, not **pipelines**.
> 
> See an example:
> 
> ```python
> PipelineJob.from_component_url(
>     url="https://..../Train_TF_model/component.yaml",
>     arguments=dict(
>         training_data_uri="gs://.../data.csv",
>         output_model_uri="gs://.../new_model",
>     ),
> ).submit()
> ```
> 
> What do you think?

In KFP SDK 2.0 (KFP repo master branch), components and pipelines are the same from users' perspective (not from authors' perspective), both can be submitted to run directly as they are in IR PipelineSpec format.

Ark-kun:
> components and pipelines are the same from users' perspective

AFAIK, this was also the case with the original KFP container components and graph components (that could encapsulate a pipelines).

>both can be submitted to run directly

Hmm. I've tested a few components that are in active use by our paying customers and partners and I was not able to make `PipelineJob(template_path=)` work with them.

For example, can you show how to run this randomly chosen component from our partners using the existing code? 
```python
PipelineJob(template_path="https://raw.githubusercontent.com/kubeflow/pipelines/a8e9f4b5ba84064412577b1050309abcd43a8347/components/PyTorch/pytorch-kfp-components/templates/ax_generate_trials_component.yaml")
```

With the proposed method it's trivial:
```python
PipelineJob.from_component_url("https://raw.githubusercontent.com/kubeflow/pipelines/a8e9f4b5ba84064412577b1050309abcd43a8347/components/PyTorch/pytorch-kfp-components/templates/ax_generate_trials_component.yaml")
```

The KFP component ecosystem consists of hundreds or thousands of components (both public and private).
Have the Pipelines team conducted formal tests to verify that they will work smoothly? I think we might want to minimize the customer disruption.

IronPan:
component.yaml in KFP V1 is on the way to be deprecated. We don't have plan to support such Vertex SDK functionality. 

Thanks for your contrib @Ark-kun. Please start such effort with a design review.

sararob:

PR #1527, opened 2022-07-20
chore: change dataset url in dataset import system test

sararob:
Update the dataset url for `test_get_new_dataset_and_import` to use a bucket in the same project as the test.

Fixes https://github.com/googleapis/python-aiplatform/issues/1484 🦕


SinaChavoshi (approved):
Thank you for the PR!

Issue #1533, opened 2022-07-21
Remove the restriction of specifying both explanation_metadata and explanation_parameters


siping-hu:
Hi, XAI team is now trying to remove the need of specifying metadata if users want to get feature attributions. Some code in the Vertex SDK is in conflict with this goal. For more details, now the code requires `Both `explanation_metadata` and `explanation_parameters` must be passed together when used`. Please remove this restriction.

PR #1534, opened 2022-07-21
Remove the restriction of specifying both explanation_metadata and explanation_parameters

siping-hu:

#1533


sararob (changes requested):
Thanks for this PR! Could you update the relevant unit tests in these files: 

* https://github.com/googleapis/python-aiplatform/blob/main/tests/unit/aiplatform/test_explain_saved_model_metadata_builder_tf1_test.py
* https://github.com/googleapis/python-aiplatform/blob/main/tests/unit/aiplatform/test_explain_saved_model_metadata_builder_tf2_test.py

You can add tests to verify that `ValueError` is raised correctly, and a that `upload` works when only `explanation_parameters` is provided.

siping-hu:
It seems better to add unit tests in test_models.py, because the change of code is not related to the metadata builder.
I've added tests to verify that ValueError is raised correctly, and a upload works when only explanation_parameters is provided, please check.
Thank you!

sararob (changes requested):
Thanks for adding the test!

Could you amend your commit message to add a `chore: ` prefix so that the Conventional Commits check passes? Details [here](https://www.conventionalcommits.org/en/v1.0.0/#summary). 

Then once you resolve the merge conflict in `models.py` and unit tests pass this should be ready to merge.

sararob (approved):

SinaChavoshi (approved):
Thank you for the PR!

google/cloud/aiplatform/models.py
```diff
@@ -687,10 +687,10 @@ def _validate_deploy_args(
                 raise ValueError(
                     "Sum of all traffic within traffic split needs to be 100."
                 )
-
-        if bool(explanation_metadata) != bool(explanation_parameters):
+        
+        if bool(explanation_metadata) == True and bool(explanation_parameters) == False:
```
SinaChavoshi, line 691: nit: it is better to use implicit if true/false if possible. see go/pystyle#truefalse-evaluations 

same for instances below.

SinaChavoshi (approved):

PR #1536, opened 2022-07-22
Feat: Refactor schema classes to subclass from _Resource

SinaChavoshi:
BREAKING CHANGE: The methods `Artifact.create_from_base_artifact_schema` and `Execution.create_from_base_execution_schema` are removed as the schema_base classes are now directly inheriting from the corresponding Artifact and Execution class. 


jialuzh (approved):
Thanks for the PR.

google/cloud/aiplatform/metadata/execution.py
```diff
@@ -167,57 +166,6 @@ def create(
 
         return self
 
-    @classmethod
-    def create_from_base_execution_schema(
```
Ark-kun, line 171: Is this a breaking change?
This method only existed for a couple of releases, but we might want to document it's removal.

SinaChavoshi: I generally agree with the sentiment however this method can not be used directly ( It should have been created as an internal method for schema subclass use only) Since it was not promoted in documentation or samples. I think it is safe to assume no one is using yet. 

let me know if you think it is still necessary, we can add a note in release documentation.

SinaChavoshi: Added a note with `BREAKING CHANGE:` tag in PR description to include in the release notes. Thank you for calling this out!

google/cloud/aiplatform/metadata/artifact.py
```diff
@@ -327,60 +326,15 @@ def create(
             credentials=credentials,
         )
 
-    @classmethod
```
dandhlee, line 330: Should the title be `feat!` to do a major bump rather than only mentioning that it's a breaking change?

dandhlee (approved):
Approving for samples.

google/cloud/aiplatform/metadata/artifact.py
SinaChavoshi: done. PR title is set to Feat: ...

dandhlee: `feat` will do a minor version bump (1.15.1 -> 1.16.0), `feat!` with the exclamation mark will do a major version bump (1.15.1 -> 2.0.0).

dandhlee: Since it's already been merged it won't have any impact on `release-please` bot. Could you let the client library maintainers for this repo know about this, and see if they'd like to do a minor version bump vs major version bump?

google/cloud/aiplatform/metadata/execution.py
SinaChavoshi: updating based on discussion / feedback from shiproom meeting. The breaking change tag forces the release process to bump the SDK version to 2.0. Given that we decided to revert the change "BREAKING CHAGE" change.  Given that this feature is supported via schema class create method this is inline with "out of scope" section of  Google Cloud Feature change and removal policy.

PR #1537, opened 2022-07-22
fix: Pass the PipelineJob credentials to `create_gcs_bucket_for_pipeline_artifacts_if_it_does_not_exist`

Ark-kun:

### DIFF
--- a/google/cloud/aiplatform/pipeline_jobs.py
+++ b/google/cloud/aiplatform/pipeline_jobs.py
@@ -348,6 +348,7 @@ class PipelineJob(
             service_account=self._gca_resource.service_account,
             project=self.project,
             location=self.location,
+            credentials=self.credentials,
         )
 
         # Prevents logs from being supressed on TFX pipelines


google/cloud/aiplatform/pipeline_jobs.py
```diff
@@ -348,6 +348,7 @@ def submit(
             service_account=self._gca_resource.service_account,
             project=self.project,
             location=self.location,
+            credentials=self.credentials,
```
sararob, line 351: Should `credentials` also be passed to `gcs_utils.create_gcs_bucket_for_pipeline_artifacts_if_it_does_not_exist` below?

sararob (changes requested):
Thanks for this! Could you also add a unit test for cases when `aiplatform.init` is not called before calling `PipelineJob.submit()`?

google/cloud/aiplatform/pipeline_jobs.py
Ark-kun: I'm not sure which code location are you talking about. Could you link to that code line?

Ark-kun:
> Thanks for this! Could you also add a unit test for cases when `aiplatform.init` is not called before calling `PipelineJob.submit()`?

What is the behavior that such test would needs to check?

Note that unit tests are executed without accessing any real infrastructure (GCS or PipelineService).

google/cloud/aiplatform/pipeline_jobs.py
sararob: Oops, I wasn't looking at the whole file. This is the right place to pass it in 👍

sararob (approved):

PR #1540, opened 2022-07-26
chore: Append users' src dir name to sys.path while model server starts for CPR.

abcdefgs0324:

### DIFF
--- a/google/cloud/aiplatform/prediction/local_model.py
+++ b/google/cloud/aiplatform/prediction/local_model.py
@@ -16,6 +16,7 @@
 #
 
 from copy import copy
+from pathlib import Path
 from typing import Dict, List, Optional, Sequence, Type
 
 from google.cloud import aiplatform
@@ -251,6 +252,7 @@ class LocalModel:
                 handler, src_dir
             )
         environment_variables = {
+            "CPR_USER_DIR_NAME": Path(src_dir).name,
             "HANDLER_MODULE": handler_module,
             "HANDLER_CLASS": handler_class,
         }
--- a/google/cloud/aiplatform/prediction/model_server.py
+++ b/google/cloud/aiplatform/prediction/model_server.py
@@ -19,6 +19,7 @@ import importlib
 import logging
 import multiprocessing
 import os
+import sys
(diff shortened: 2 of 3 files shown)


google/cloud/aiplatform/prediction/local_model.py
```diff
@@ -251,6 +252,7 @@ def build_cpr_model(
                 handler, src_dir
             )
         environment_variables = {
+            "CPR_USER_DIR_NAME": Path(src_dir).name,
```
kthytang, line 255: Why do we need to do Path().name here?

google/cloud/aiplatform/prediction/model_server.py
```diff
@@ -186,6 +187,10 @@ def set_number_of_workers_from_env() -> None:
 
 
 if __name__ == "__main__":
+    cpr_user_dir_name = os.getenv("CPR_USER_DIR_NAME")
+    if cpr_user_dir_name:
+        sys.path.append(os.path.join(os.getcwd(), cpr_user_dir_name))
```
kthytang, line 192: Can we insert at the beginning of sys.path? So it searches the user's modules first.

kthytang (changes requested):

google/cloud/aiplatform/prediction/local_model.py
abcdefgs0324: If users provide "tmp/src_dir", we check all the necessary code (predictor, handler, and requirements.txt) should be in "tmp/src_dir" and copy "src_dir" to the workdir in the images.

google/cloud/aiplatform/prediction/model_server.py
abcdefgs0324: Done.

google/cloud/aiplatform/prediction/local_model.py
kthytang: Ah, understood, thanks.

kthytang (approved):

SinaChavoshi (approved):
lgtm from metadata side.

PR #1541, opened 2022-07-26
feat: add delete method to PipelineJob

sararob:
Added `PipelineJob.delete()` method and relevant tests. This was already available in the gapic client but hadn't yet been exposed in the SDK. 

Fixes b/240203087 🦕


google/cloud/aiplatform/pipeline_jobs.py
```diff
@@ -959,3 +959,17 @@ def get_associated_experiment(self) -> Optional["aiplatform.Experiment"]:
                 location=self.location,
                 credentials=self.credentials,
             )
+
```
jaycee-li, line 962: Could you please change copyright year to 2022

google/cloud/aiplatform/pipeline_jobs.py
```diff
@@ -959,3 +959,17 @@ def get_associated_experiment(self) -> Optional["aiplatform.Experiment"]:
...
+    def delete(self, sync: bool = True) -> None:
+        """Deletes this PipelineJob resource. Does not delete Artifacts created
+        from this Pipeline Job.
+
+        WARNING: This deletion is permanent.
+
+        Args:
+            sync (bool):
+                Whether to execute this method synchronously. If False, this method
```
jaycee-li, line 971: Please add 'Optional'

google/cloud/aiplatform/pipeline_jobs.py
```diff
@@ -959,3 +959,17 @@ def get_associated_experiment(self) -> Optional["aiplatform.Experiment"]:
                 location=self.location,
                 credentials=self.credentials,
             )
+
+    def delete(self, sync: bool = True) -> None:
```
jaycee-li, line 963: I wonder if we can add optional kwarg(s) for this method, for example, `delete_artifacts: bool = False`. Otherwise it has no difference with the higher level delete method.

jaycee-li (changes requested):

google/cloud/aiplatform/pipeline_jobs.py
sararob: Good point. I think having a `delete_artifacts` option would be useful, however it doesn't look like the `Artifact` class has implemented delete yet: https://github.com/googleapis/python-aiplatform/blob/3d3e0aa9f0283dafe3056ed7d3067310535db48c/google/cloud/aiplatform/metadata/artifact.py#L40

I can chat with the metadata team about adding this and will follow up here.

For future reference, we can access the `ArtifactList` from a pipeline via `PipelineJob._gca_resource.job_detail.task_details`.

jaycee-li: Currently `Artifact` uses the high level delete method from [base.py](https://github.com/googleapis/python-aiplatform/blob/34bbd0abbaf29e644ed7703b3251f1de65bf5a86/google/cloud/aiplatform/base.py#L1200).

Good to know that we have an `ArtifactList` to track all the children artifacts! Then I think we can simply implement like this:
```
if delete_artifacts:
  for artifact in pipeline_job._gca_resource.job_detail.task_details.artifact_list:
     artifact.delete()
```

PR #1559, opened 2022-08-01
feat: Added `cloud_console_url` property to Vertex resources

Ark-kun:
The property returns a URL that leads to Google Cloud console page showing the resource.

The URL works for most resources, but not for all.
For example the correct URL for PipelineJob is
https://console.cloud.google.com/vertex-ai/locations/us-central1/pipelines/runs/<id>?project=<project>
instead of
https://console.cloud.google.com/vertex-ai/locations/us-central1/pipelineJobs/<id>?project=<project>

### DIFF
--- a/google/cloud/aiplatform/base.py
+++ b/google/cloud/aiplatform/base.py
@@ -736,6 +736,23 @@ class VertexAiResourceNoun(metaclass=abc.ABCMeta):
             prefix = cls.__name__
         return prefix + " " + datetime.datetime.now().isoformat(sep=" ")
 
+    @property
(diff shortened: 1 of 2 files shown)


sasha-gitg (changes requested):
This doesn't seem to work for literal resource names with multiple components because they are hyphenated in the url. ie: `metadataStores` -> `metadata-stores`

Also, preference to remove the dashboard_uri usage across the SDK and replace with the new implementation to:
1. Avoid keeping around unused code
2. Maintain a single implementation

sararob:

PR #1564, opened 2022-08-01
fix: Append users' src dir name to sys.path instead of inserting to the

abcdefgs0324:
first.

### DIFF
--- a/google/cloud/aiplatform/prediction/model_server.py
+++ b/google/cloud/aiplatform/prediction/model_server.py
@@ -189,7 +189,7 @@ def set_number_of_workers_from_env() -> None:
 if __name__ == "__main__":
     cpr_user_dir_name = os.getenv("CPR_USER_DIR_NAME")
     if cpr_user_dir_name:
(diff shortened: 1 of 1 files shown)


google/cloud/aiplatform/prediction/model_server.py
```diff
@@ -189,7 +189,7 @@ def set_number_of_workers_from_env() -> None:
 if __name__ == "__main__":
     cpr_user_dir_name = os.getenv("CPR_USER_DIR_NAME")
     if cpr_user_dir_name:
-        sys.path.insert(0, os.path.join(os.getcwd(), cpr_user_dir_name))
+        sys.path.append(os.path.join(os.getcwd(), cpr_user_dir_name))
```
kthytang, line 192: Can we insert in the second position after /usr/app? I think we should search through user provided src code first before searching through other installed dependencies.

abcdefgs0324: Done.

PR #1565, opened 2022-08-01
chore: fix resource deletion issues with model monitoring system tests

rosiezou:

Fixes #<240633310, 240929827> 🦕

### DIFF
--- a/tests/system/aiplatform/test_model_monitoring.py
+++ b/tests/system/aiplatform/test_model_monitoring.py
@@ -168,7 +168,9 @@ class TestModelDeploymentMonitoring(e2e_base.TestEndToEnd):
             serving_container_image_uri=IMAGE,
         )
         shared_state["resources"] = [model1, model2]
-        endpoint = aiplatform.Endpoint.create(display_name=self._make_display_name(key=MODEL_NAME))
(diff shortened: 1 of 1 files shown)


tests/system/aiplatform/test_model_monitoring.py
```diff
@@ -139,7 +139,7 @@ def temp_endpoint(self, shared_state):
         )
 
         model = aiplatform.Model.upload(
-            display_name=MODEL_NAME,
+            display_name=self._make_display_name(key=MODEL_NAME),
             artifact_uri=CHURN_MODEL_PATH,
             serving_container_image_uri=IMAGE,
         )
```
nayaknishant, line 145: When the model is being deployed, on the line `endpoint = model.deploy(machine_type="n1-standard-2")`, I think endpoint is (if unspecified) set to model display name+'_endpoint'. Want to make sure with you that this is the behavior you're looking for. Works for me

rosiezou: yep that's the intended behavior; no need to specify the endpoint name here because we're passing the endpoint object itself to the MDM job creation call

nayaknishant (approved):

PR #1570, opened 2022-08-04
feat: support model monitoring for batch prediction in Vertex SDK

rosiezou:


google/cloud/aiplatform/jobs.py
```diff
@@ -551,6 +558,23 @@ def create(
                 but too high value will result in a whole batch not fitting in a machine's memory,
                 and the whole operation will fail.
                 The default value is 64.
+            model_monitoring_objective_config (aiplatform.model_monitoring.ObjectiveConfig):
+                Optional. The objective config for model monitoring. Passing this parameter enables
+                monitoring on the model associated with this batch prediction job.
+            model_monitoring_alert_config (aiplatform.model_monitoring.EmailAlertConfig):
+                Optional. Configures how model monitoring alerts are sent to the user. Right now
```
dizcology, line 565: Are these docstrings copied from the source at https://github.com/googleapis/googleapis/tree/master/google/cloud/aiplatform?  Will we be able to remember to update this when/if alerts other than email alert become supported?

google/cloud/aiplatform/jobs.py
```diff
@@ -702,6 +759,11 @@ def create(
             sync=sync,
             create_request_timeout=create_request_timeout,
         )
+        # TODO: remove temporary re-import statements once model monitoring for batch prediction is GA
```
dizcology, line 762: Since there are several of these throughout the PR, perhaps add a tracking bug number so the TODO comments look like `# TODO(b/.....): remove...` with a common bug number.

google/cloud/aiplatform/model_monitoring/alert.py
```diff
@@ -17,9 +17,16 @@
 
 from typing import Optional, List
 from google.cloud.aiplatform_v1.types import (
-    model_monitoring as gca_model_monitoring,
+    model_monitoring as gca_model_monitoring_v1,
```
dizcology, line 20: Should we use `compat` for these imports?

google/cloud/aiplatform/model_monitoring/alert.py
```diff
@@ -40,8 +47,19 @@ def __init__(
         self.enable_logging = enable_logging
         self.user_emails = user_emails
 
-    def as_proto(self):
-        """Returns EmailAlertConfig as a proto message."""
+    # TODO: remove config_for_bp parameter when model monitoring for batch prediction is GA
+    def as_proto(self, config_for_bp: Optional[bool] = False):
```
dizcology, line 51: This typing annotation should be just `config_for_bp: bool = False`.

google/cloud/aiplatform/model_monitoring/objective.py
```diff
@@ -186,11 +199,49 @@ def __init__(
         self.drift_detection_config = drift_detection_config
         self.explanation_config = explanation_config
 
-    def as_proto(self):
-        """Returns _ObjectiveConfig as a proto message."""
+    # TODO: remove config_for_bp parameter when model monitoring for batch prediction is feature complete and in GA
+    def as_proto(self, config_for_bp: Optional[bool] = False):
```
dizcology, line 203: Same comment as above about typing annotation here.

tests/system/aiplatform/test_model_monitoring.py
```diff
@@ -24,6 +24,11 @@
...
 from tests.system.aiplatform import e2e_base
 
+from google.cloud.aiplatform_v1.types import (
+    io as gca_io,
+    model_monitoring as gca_model_monitoring,
+)
+
 # constants used for testing
 USER_EMAIL = ""
 MODEL_NAME = "churn"
```
dizcology, line 34: Is this used as the model's `display_name`?  if so maybe better call it `MODEL_DISPLAY_NAME`, since "model name" could be understood as its resource full name.

If this is indeed intended as a display name of a resource used only for testing, please prefix it with "temp".

dizcology (changes requested):

google/cloud/aiplatform/model_monitoring/alert.py
rosiezou: `compat` by default imports from GA version of GAPIC, unless if `DEFAULT_VERSION` is set to `v1beta1`. I tried to see if the import aliases will index into the correct version by simply setting `DEFAULT_VERSION = 'v1beta1'` and then switching it back to `v1` on an ad-hoc basis, but it doesn't dynamically index in the way I was hoping for. I think it's because the symbol table isn't automatically re-written unless if we explicitly re-import. So that's why I imported both v1 and v1beta1 versions explicitly.

google/cloud/aiplatform/jobs.py
rosiezou: They're not directly copied from GAPIC. But Jing's team also confirmed that there's no plans for additional alert configs.

tests/system/aiplatform/test_model_monitoring.py
rosiezou: I updated the variable names to make it more specific. But the extra prefixing is unnecessary because the actual display names are being created by `_make_display_name` implemented in `e2e_base.py`, which does append a prefix set by the class (in this case the prefix is `'temp_e2e_model_monitoring_test_'`)

google/cloud/aiplatform/model_monitoring/alert.py
rosiezou: updated

google/cloud/aiplatform/jobs.py
rosiezou: buganizer component created & assigned

google/cloud/aiplatform/model_monitoring/objective.py
rosiezou: updated

tests/system/aiplatform/test_model_monitoring.py
dizcology: Thanks. In that case can we rename this further and call it something like MODEL_DISPLAYNAME_KEY to prevent the next engineering reading this to misinterpret it as the actual model displayname?  More generally, what’s the meaning of “churn” in the context of testing?

dizcology (approved):

tests/system/aiplatform/test_model_monitoring.py
rosiezou: I've updated the variable name. Regarding the actual value string, "churn" is just a shorthand used to reference the BQ dataset. The same dataset and pre-trained model was used in the [example notebook](https://github.com/GoogleCloudPlatform/vertex-ai-samples/blob/main/notebooks/official/model_monitoring/model_monitoring.ipynb) on the Cloud SDK documentation page and a [separate example for BQML](https://cloud.google.com/blog/topics/developers-practitioners/churn-prediction-game-developers-using-google-analytics-4-ga4-and-bigquery-ml).

SinaChavoshi (approved):
Approving from MLMD, as this is a no-op on MLMD side.

PR #1574, opened 2022-08-05
chore: add model monitoring SDK code samples to model-builder

rosiezou:

### DIFF
--- a/samples/model-builder/create_model_monitoring_job_example.py
+++ b/samples/model-builder/create_model_monitoring_job_example.py
@@ -1,0 +1,67 @@
+# Copyright 2022 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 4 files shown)


yil532 (approved):
LGTM

yil532 (approved):
LGTM

samples/model-builder/create_model_monitoring_job_example.py
```diff
@@ -0,0 +1,69 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from google.cloud import aiplatform
+from google.cloud.aiplatform import model_monitoring
+from typing import Optional, Dict, Union, List
+
+
+#  [START aiplatform_sdk_create_model_monitoring_job_example]
+def create_model_monitoring_job_example(
```
dizcology, line 21: Why are we exposing all these arguments in the sample function?  The sample function’s body should illustrate a use case of a library method, and the sample function and its arguments are there for sample testing purposes, and we are not expecting the user to run the sample function.

We expect the illustrated library method to have sensible defaults for most of its arguments.  As such the sample function should expose only what is needed at sample testing time. As is right now the sample function appears to be doing nothing but flattening some of the parameters.

dizcology (changes requested):

samples/model-builder/create_model_monitoring_job_example.py
dizcology: Here is an example where the sample function exposes only what’s needed for sample testing and the sample function body illustrates a use case: https://github.com/googleapis/python-aiplatform/blob/main/samples/snippets/job_service/create_custom_job_sample.py

rosiezou:
closing this for now. will revisit at a later time.

PR #1588, opened 2022-08-11
fix: log_metrics docstring error

jaycee-li:


SinaChavoshi (approved):
Thank you for the PR!

PR #1591, opened 2022-08-12
feat: Add support for SDK Method metrics tracking via _USER_AGENT_SDK…

SinaChavoshi:
…_COMMAND


jialuzh (approved):

google/cloud/aiplatform/metadata/context.py
```diff
@@ -136,6 +137,13 @@ def create(
         Returns:
             Context: Instantiated representation of the managed Metadata Context.
         """
+        # Add User Agent Header for metrics tracking if one is not specified
+        # If one is alreayd specified this call was initiated by a sub class.
```
sararob, line 141: Nit: "already", also applies to other instances of this comment.

google/cloud/aiplatform/metadata/execution.py
```diff
@@ -142,18 +142,110 @@ def create(
             Execution: Instantiated representation of the managed Metadata Execution.
 
         """
-        self = cls._empty_constructor(
-            project=project, location=location, credentials=credentials
+        # Add User Agent Header for metrics tracking if one is not specified
+        # If one is alreayd specified this call was initiated by a sub class.
+        if not base_constants.USER_AGENT_SDK_COMMAND:
+            base_constants.USER_AGENT_SDK_COMMAND = (
+                "aiplatform.metadata.execution.Execution.create"
```
sararob, line 149: So that you don't have to hard code all of these module name strings, I wonder if you could save the prefix to a variable and store it in `constants/base.py` using:

```
user_agent_sdk_prefix = f"{self.__module__}.{self.__class__.__name__}."
```

In this case I think that would give you `"aiplatform.metadata.execution.Execution."`. And then you'd append the method name to that prefix. I'm not sure if there's a way to get the current method name, but if there is you could use that here too.

google/cloud/aiplatform/metadata/execution.py
```diff
@@ -142,18 +142,110 @@ def create(
...
+            location (str):
+                Optional. Location used to create this Execution. Overrides location set in
+                aiplatform.init.
+            credentials (auth_credentials.Credentials):
+                Optional. Custom credentials used to create this Execution. Overrides
+                credentials set in aiplatform.init.
+
+        Returns:
+            Execution: Instantiated representation of the managed Metadata Execution.
+
```
sararob, line 221: Just want to double check if this change is related to the addition of the user agent header?

google/cloud/aiplatform/metadata/schema/base_artifact.py
```diff
@@ -114,6 +115,11 @@ def _init_with_resource_name(
                 Artifact name with the following format, this is globally unique in a metadataStore:
                 projects/123/locations/us-central1/metadataStores/<metadata_store_id>/artifacts/<resource_id>.
         """
+        # Add User Agent Header for metrics tracking if one is not specified
+        # If one is alreayd specified this call was initiated by a sub class.
+        if not base_constants.USER_AGENT_SDK_COMMAND:
+            base_constants.USER_AGENT_SDK_COMMAND = "aiplatform.metadata.schema.base_artifact.BaseArtifactSchema._init_with_resource_name"
```
sararob, line 121: Same comment as above re: storing this prefix as a variable.

sararob (changes requested):
Thanks for adding this! Left a few comments. Before we submit this, is there a way to test that these method calls are getting added to a dashboard in the way we're expecting?

google/cloud/aiplatform/metadata/execution.py
SinaChavoshi: Good question, the previous implementation of the `Execution.create` was different from `Artifact` & `Context` create method. This would significantly complicate adding the user agent header. Instead of working around that I decided to refactor the code to use the same implementation for `Execution` as `Artifact` and `Context` create methods. 

This will additionally allow us to remove the `_create` method from individual classes and move it to `resource._Resource`, this will be covered in a follow up PR , added a TODO to track this.

google/cloud/aiplatform/metadata/context.py
SinaChavoshi: oops :) good catch.

google/cloud/aiplatform/metadata/execution.py
SinaChavoshi: good question, we experimented with using a flexible prefix as you mentioned however it will result in too many variations based on the calling subclass, for example for Schema classes depending on the artifact type the path will be different such as ``` sdk_command/google.cloud.aiplatform.metadata.schema.system.artifact_schema.Metrics.create``` if the create is called from Metrics schema subclass versus ```sdk_command/google.cloud.aiplatform.metadata.schema.system.artifact_schema.Model.create```

Based on offline discussion with Jialu I think for now we will keep the metadata implementation limited to the hard coded string. 

That being said for long term maintenance using a shared method to set the name would be much easier.

google/cloud/aiplatform/metadata/schema/base_artifact.py
SinaChavoshi: same comment as above.

sararob (approved):

PR #1593, opened 2022-08-12
feat: add input artifact when creating a pipeline

### LINKED ISSUE #1592: Support input artifacts when creating pipeline job
Thanks for stopping by to let us know something could be better!

**PLEASE READ**: If you have a support contract with Google, please create an issue in the [support console](https://cloud.google.com/support/) instead of filing on GitHub. This will ensure a timely response.

This feature is required by Eval team.

rui5i:

Fixes #1592 🦕


sararob (commented):
Thanks for starting this! Do you have an example pipeline that takes an input artifact uri as a parameter? I think that would help me understand how this feature will be used in the SDK.

google/cloud/aiplatform/pipeline_jobs.py
```diff
@@ -685,6 +692,10 @@ def clone(
                 Optional. The mapping from runtime parameter names to its values that
                 control the pipeline run. Defaults to be the same values as original
                 PipelineJob.
+            input_artifacts (Dict[str, str]):
+                Optional. The mapping from runtime input artifact names to its id that
+                control the pipeline run. Defaults to be the same values as original
+                PipelineJob.
```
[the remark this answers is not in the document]
rui5i: For example there is an existing model artifact on MLMD`"projects/123/locations/us-central1/metadataStores/default/artifacts/456"`, the mapping would look like user inputs `{"vertex_model": "456"}`, and pipeline proto is also expecting the artifact id without project and locations. Because the project id and locations must be the same as where pipeline job is running.

tests/unit/aiplatform/test_pipeline_jobs.py
```diff
@@ -850,6 +854,86 @@ def test_run_call_pipeline_service_create_with_failure_policy(
             gca_pipeline_state.PipelineState.PIPELINE_STATE_SUCCEEDED
         )
```
jaycee-li, line 856: Could you please also update the `test_clone_pipeline_job_with_all_args` with `input_artifacts`? Thanks!

tests/unit/aiplatform/test_pipeline_jobs.py
```diff
@@ -850,6 +854,86 @@ def test_run_call_pipeline_service_create_with_failure_policy(
...
 
+    @pytest.mark.parametrize(
+        "job_spec",
+        [_TEST_PIPELINE_SPEC_JSON, _TEST_PIPELINE_SPEC_YAML, _TEST_PIPELINE_JOB],
+    )
+    @pytest.mark.parametrize("sync", [True, False])
+    def test_run_call_pipeline_service_create_with_input_artifact(
```
jaycee-li, line 862: I wonder if it is necessary to add this test? An alternative option is to add `input_artifacts` in the existing `test_run_call_pipeline_service_create`.

google/cloud/aiplatform/pipeline_jobs.py
```diff
@@ -139,6 +140,9 @@ def __init__(
             parameter_values (Dict[str, Any]):
                 Optional. The mapping from runtime parameter names to its values that
                 control the pipeline run.
+            input_artifacts (Dict[str, str]):
```
jaycee-li, line 143: As Sara mentioned, please add the example `{"vertex_model":"456"}` to the docstring. 

Please also specify that the full resource name(`"projects/123/locations/us-central1/metadataStores/default/artifacts/456"`) cannot be used. Because in some other SDK methods, we do allow both resource id and full resource name.

jaycee-li (changes requested):

google/cloud/aiplatform/pipeline_jobs.py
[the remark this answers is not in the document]
rui5i: Done.

tests/unit/aiplatform/test_pipeline_jobs.py
rui5i: Done.

tests/unit/aiplatform/test_pipeline_jobs.py
rui5i: Done.

google/cloud/aiplatform/pipeline_jobs.py
rui5i: Done.

tests/unit/aiplatform/test_pipeline_jobs.py
```diff
@@ -1467,6 +1473,7 @@ def test_clone_pipeline_job_with_all_args(
             template_path=_TEST_TEMPLATE_PATH,
             job_id=_TEST_PIPELINE_JOB_ID,
             parameter_values=_TEST_PIPELINE_PARAMETER_VALUES,
+            input_artifacts=_TEST_PIPELINE_INPUT_ARTIFACTS,
```
jaycee-li, line 1476: Could you please remove this line? The purpose of this test is to check if clone method can update `input_artifacts` while cloning an existing job. I think it makes more sense to create a job without setting `input_artifacts`, clone it with artifacts being set, and check if the cloned job has the setting.

jaycee-li (changes requested):

tests/unit/aiplatform/test_pipeline_jobs.py
rui5i: Done.

jaycee-li (approved):

google/cloud/aiplatform/pipeline_jobs.py
Ark-kun: Can we automatically handle full artifact resource names?
We can just parse the resource names and get artifact ID.

P.S. I see that currently components may have input artifacts from different projects.
So it makes some sense to have that for pipeline input artifacts as well...

rui5i: Since we are retrieving the existing artifact from pipeline backend, I don't think cross project/location artifact is supported yet. One todo item we can do is to add validation rules either SDK and UI, or pipeline API layer to graceful handle the full resource name.

SinaChavoshi: I'm a bit confused here, could you clarify what is an artifact name?  is that the artifact resourceName such as"projects/123/locations/us-central1/metadataStores/default/artifacts/456" or the artifact display_name "some_user_specified_string" or the artifact resourceID which is the "456" portion of the resourceName?  also this mapping is already maintained in metadata system we should not ask the user to provide it separately ( with a caveat being that display_name is not unique )

SinaChavoshi (changes requested):

google/cloud/aiplatform/pipeline_jobs.py
Ark-kun: >I don't think cross project/location artifact is supported yet.

I see. Maybe I was mistaken.

rui5i: Updated the docstring after the offline sync.

SinaChavoshi: thank you! lgtm

SinaChavoshi (approved):
Thank you for the update! lgtm.

PR #1600, opened 2022-08-17
feat: update the samples of hyperparameter tuning in the public doc

gericdong:

Fixes #240582178 🦕


samples/model-builder/create_hyperparameter_tuning_job_sample.py
```diff
@@ -0,0 +1,74 @@
...
+
+    custom_job = aiplatform.CustomJob(
+        display_name='custom_job',
+        worker_pool_specs=worker_pool_specs,
+    )
+
+    hpt_job = aiplatform.HyperparameterTuningJob(
+        display_name=display_name,
+        custom_job=custom_job,
+        metric_spec={
```
ivanmkc, line 54: We usually pass in some of these as parameters to the function but I'm not sure how to include things like parameter specs without cluttering the parameter space.

ivanmkc: @dizcology do you have guidance or is this okay?

samples/model-builder/create_hyperparameter_tuning_job_sample.py
```diff
@@ -0,0 +1,74 @@
...
+    worker_pool_specs = [
+        {
+            "machine_spec": {
+                "machine_type": "n1-standard-4",
+                "accelerator_type": "NVIDIA_TESLA_K80",
+                "accelerator_count": 1,
+            },
+            "replica_count": 1,
+            "container_spec": {
+                "image_uri": container_uri,
```
ivanmkc, line 39: Seems weird to pass in `container_uri` when it's not used in the tuning job. 

@dizcology thoughts?

ivanmkc:
In the future, please have one PR per sample.

samples/model-builder/create_hyperparameter_tuning_job_sample.py
halio-g: Hptuning job creates sub jobs(trials) using the container image uri, it's actually used by the vertex hptuning jobs.

halio-g (approved):

ivanmkc (approved):
Let's get this in for now.

We can modify the parameters later if required.

PR #1608, opened 2022-08-23
feat: Support filters in matching engine vector matching

jaycee-li:

This PR adds the feature that was requested in [b/233694218](https://buganizer.corp.google.com/issues/233694218)

### DIFF
--- a/google/cloud/aiplatform/matching_engine/matching_engine_index_endpoint.py
+++ b/google/cloud/aiplatform/matching_engine/matching_engine_index_endpoint.py
@@ -15,7 +15,7 @@
 # limitations under the License.
 #
 
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from typing import Dict, List, Optional, Sequence, Tuple
 
 from google.auth import credentials as auth_credentials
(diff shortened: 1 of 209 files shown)


nayaknishant:
Hey @jaycee-li is it necessary to have the Namespace class? I'm wondering what's the easiest way for users to understand how to go about making/using filters.

jaycee-li:
> Hey @jaycee-li is it necessary to have the Namespace class? I'm wondering what's the easiest way for users to understand how to go about making/using filters.

@nayaknishant Yeah I thought about it. But the filter message is defined in gapic, I didn't find a way to make the structure simpler in SDK. If not using Namespace class, users will need to construct the filter (in dictionary or list) by themselves. I feel like this Namespace class can help users better understand the filter message.

nayaknishant:
Sounds good. To emphasize how Namespace is used can we change the example in the comment to `For example, [Namespace("color", ["red"], []), Namespace("shape", [], ["squared"])] will match datapoints that satisfy "red color" and not include datapoints that are "squared"`, or something like that (showing how both `allow_tokens` and `deny_tokens` are used).

jaycee-li:
> Sounds good. To emphasize how Namespace is used can we change the example in the comment to `For example, [Namespace("color", ["red"], []), Namespace("shape", [], ["squared"])] will match datapoints that satisfy "red color" and not include datapoints that are "squared"`, or something like that (showing how both `allow_tokens` and `deny_tokens` are used).

@nayaknishant  Updated.

google/cloud/aiplatform/matching_engine/matching_engine_index_endpoint.py
```diff
@@ -51,6 +51,24 @@ class MatchNeighbor:
...
+class Namespace:
+    """Namespace specifies the rules for determining the datapoints that are eligible for each matching query, overall query is an AND across namespaces.
+    Args:
+        name (str):
+            Required. The name of this Namespace.
+        allow_tokens (List(str)):
+            Optional. The allowed tokens in the namespace.
+        deny_tokens (List(str)):
+            Optional. The denied tokens in the namespace. When a token is denied, then matches will be excluded whenever the other datapoint has that token.
+            For example, if a query specifies {color: red, blue, !purple}, then that query will match datapoints that are red or blue, but if those points are also purple, then they will be excluded even if they are red/blue.
```
nayaknishant, line 64: Let's also change `if a query specifies {color: red, blue, !purple}` to `if a query specifies [Namespace("color", ["red"], ["purple"]), Namespace("color", ["blue], [])]`

jaycee-li: Updated

nayaknishant (approved):

PR #1614, opened 2022-08-25
feat: add enable_simple_view to PipelineJob.list()

sararob:
Add support for `read_mask` option in `PipelineJob.list()` to improve performance. This will reduce the number of fields returned in the response for each PipelineJob.


google/cloud/aiplatform/pipeline_jobs.py
```diff
@@ -538,9 +544,34 @@ def list(
...
+                    "start_time",
+                    "end_time",
+                    "update_time",
+                    "labels",
+                    "template_uri",
+                    "template_metadata.version",
+                    "job_detail.pipeline_run_context",
+                    "job_detail.pipeline_context",
+                ]
+            )
```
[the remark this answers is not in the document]
sararob: Done

google/cloud/aiplatform/base.py
```diff
@@ -1026,6 +1027,7 @@ def _list(
         cls_filter: Callable[[proto.Message], bool] = lambda _: True,
         filter: Optional[str] = None,
         order_by: Optional[str] = None,
+        read_mask: Optional[field_mask.FieldMask] = None,
```
jaycee-li, line 1030: Could you please add doc string for this and the one in the next method?

google/cloud/aiplatform/pipeline_jobs.py
```diff
@@ -524,6 +528,10 @@ def list(
                 Optional. A comma-separated list of fields to order by, sorted in
                 ascending order. Use "desc" after a field name for descending.
                 Supported fields: `display_name`, `create_time`, `update_time`
+            enable_simple_view (bool):
+                Optional. Whether to pass the `read_mask` parameter to the list call.
+                This will improve the performance of calling list(). However, the
+                returned PipelineJob list will not include all fields for each PipelineJob.
```
jaycee-li, line 534: I think here we can specify what fields are missing.

jaycee-li (changes requested):

jaycee-li (approved):

PR #1619, opened 2022-08-29
feat: Import client_abc from open source Vizier in Vertex Vizier SDK. 

halio-g:


google/cloud/aiplatform/vizier/study.py
```diff
@@ -232,22 +232,30 @@ def materialize_study_config(self) -> vz.StudyConfig:
...
+        name: str,
         project: Optional[str] = None,
         location: Optional[str] = None,
         credentials: Optional[auth_credentials.Credentials] = None,
     ) -> _T:
         """Fetches an existing study from the Vizier service.
 
         Args:
-          uid (str): Required. Unique identifier of the study.
+          name (str): Globally unique identifier of the study.
```
sagipe, line 252: Do you want to provide an example of what format is expected here for name?

google/cloud/aiplatform/vizier/trial.py
```diff
@@ -178,3 +207,16 @@ def materialize(self, *, include_all_measurements: bool = True) -> vz.Trial:
         """
         trial = self.api_client.get_trial(name=self.resource_name)
         return copy.deepcopy(vz.TrialConverter.from_proto(trial))
```
sagipe, line 209: Shouldn't this method return a boolean?
We is it returning a pyvizier.Trial?

Can you check that the unit test is not implicitly converting pyvizier.Trial to a boolean? If there's one test where the method should return False, it should catch this bug.

halio-g: from_proto translate the service proto into vz.Trial. It's not boolean value. We have unit test test_materialize to verify it.

google/cloud/aiplatform/vizier/trial.py
```diff
@@ -109,6 +113,19 @@ def status(self) -> vz.TrialStatus:
         trial = self.api_client.get_trial(name=self.resource_name)
         return vz.TrialConverter.from_proto(trial).status
 
+    @property
+    def study(self) -> client_abc.StudyInterface:
+        """Returns the Study that this Trial belongs to."""
+        trial_path_components = self._parse_resource_name(self.resource_name)
+        return study.Study(
+            study.Study._format_resource_name(
+                project=trial_path_components["project"],
```
rosiezou, line 122: I think if you're inheriting from `VertexAiResourceNounWithFutureManager`, you should be able to do `project=self.project` and `location=self.location`, since `project` and `location` are also properties in the base class.

rosiezou (changes requested):

google/cloud/aiplatform/vizier/study.py
```diff
@@ -232,22 +232,34 @@ def materialize_study_config(self) -> vz.StudyConfig:
...
-        return copy.deepcopy(vz.StudyConfig.from_proto(study.study_spec))
+        return copy.deepcopy(
+            vz.ProblemStatement.from_problem(
+                vz.StudyConfig.from_proto(study.study_spec)
+            )
+        )
 
     @classmethod
-    def from_uid(
+    def from_resource_name(
```
rosiezou, line 242: This is a change to the public interface. It could potentially break existing customers' code.

google/cloud/aiplatform/vizier/study.py
```diff
@@ -223,7 +223,7 @@ def optimal_trials(self) -> Collection[client_abc.TrialInterface]:
             for trial in optimal_trials_response.optimal_trials
         ]
 
-    def materialize_study_config(self) -> vz.StudyConfig:
+    def materialize_problem_statement(self) -> vz.ProblemStatement:
```
rosiezou, line 226: same comment as the one I left for `def from_resource_name`. I think this constitutes a breaking change.

google/cloud/aiplatform/vizier/study.py
```diff
@@ -88,7 +88,7 @@ def __init__(
     def create_or_load(
         cls,
         display_name: str,
-        problem: vz.ProblemStatement,
+        problem: vz.StudyConfig,
```
rosiezou, line 91: Looks like `ProblemStatement` is imported in `pyvizier/__init__.py`. Is the class no longer being used and its features are now completely replaced and/or incapsulated in `StudyConfig`?

google/cloud/aiplatform/vizier/study.py
```diff
@@ -232,22 +232,34 @@ def materialize_study_config(self) -> vz.StudyConfig:
...
+                vz.StudyConfig.from_proto(study.study_spec)
+            )
+        )
 
     @classmethod
-    def from_uid(
+    def from_resource_name(
         cls: Type[_T],
-        uid: str,
+        name: str,
```
rosiezou, line 244: same comment as above. This is a breaking change for customers who use keyword arguments in their function calls

google/cloud/aiplatform/vizier/trial.py
```diff
@@ -146,8 +163,20 @@ def complete(
             else None
         )
 
-    def should_stop(self) -> bool:
-        """Returns true if the Trial should stop."""
+    def check_early_stopping(self) -> bool:
```
rosiezou, line 166: same comment as above. this is a breaking change

rosiezou (changes requested):

google/cloud/aiplatform/vizier/trial.py
halio-g: Looks like the Vertex Vizier SDK is not used too much by the customers. I think it's fine to make the change on the API ? https://datastudio.google.com/c/u/0/reporting/d8a64931-8a91-4f07-9abc-93fece2e318d/page/p_n2odnmsgmc

google/cloud/aiplatform/vizier/study.py
halio-g: The ProblemStatement is the rename of Study config.

google/cloud/aiplatform/vizier/study.py
halio-g: We can create another function of it. If it makes more sense.

google/cloud/aiplatform/vizier/study.py
halio-g: We can create another function of it. If it makes more sense.

google/cloud/aiplatform/vizier/study.py
sagipe: ProblemStatement is the base class for StudyConfig, so both can be used from a pytype perspective.
Using StudyConfig here means that we want the Vertex specific StudyConfig, and not any subclass of ProblemStatement.

We expect Vertex users to only create StudyConfig objects, not ProblemStatement objects.

sagipe (approved):
halio@ this LGTM from my side. Please address other comments.

sararob:

PR #1620, opened 2022-08-30
feat: support raw_predict for Endpoint

rosiezou:


nayaknishant (commented):
Is `test_model_interactions.py` testing permanent models? If it is I think we should keep it as two separate files (as you have it), but if it's using temp models created in `test_model_upload.py` wouldn't it make more sense to have a `predict()` call and then a `raw_predict()` call after in the same file (maybe change the name to a more broad `test_model.py` ?

rosiezou:
@nayaknishant to answer your question: in addition to feature work, this PR also introduces a minor refactor on `test_model_upload.py`. Previously, `test_model_upload.py` tested model upload, update, and also predict. I figured that we could move predict out of the file into a separate one that only handles model interactions. That way we can make the tests more atomic and independent, since if we're only testing interactions then we can test `predict` and `raw_predict` against existing models and endpoints. It makes for a cleaner system test design.

google/cloud/aiplatform/models.py
```diff
@@ -1505,29 +1511,67 @@ def predict(
...
+        if use_raw_predict:
+            raw_predict_response = self.raw_predict(
+                body=json.dumps({"instances": instances, "parameters": parameters}),
+                headers={"Content-Type": "application/json"},
+            )
+            response_text = json.loads(raw_predict_response.text)
+            return Prediction(
+                predictions=response_text["predictions"],
+                deployed_model_id=raw_predict_response.headers[
+                    "X-Vertex-AI-Deployed-Model-Id"
```
nayaknishant, line 1533: Interesting, if you can clarify @rosiezou, I'm guessing `X-Vertex-AI-Deployed-Model-Id` is populated when the response is returned?

nayaknishant (commented):
I see that system tests have been added, but can we add also unit tests to https://github.com/googleapis/python-aiplatform/blob/main/tests/unit/aiplatform/test_models.py for `raw_predict()` to ensure the function works as intended during future development.

google/cloud/aiplatform/models.py
```diff
@@ -2062,6 +2106,30 @@ def predict(self, instances: List, parameters: Optional[Dict] = None) -> Predict
             deployed_model_id=self._gca_resource.deployed_models[0].id,
         )
 
+    def raw_predict(
+        self, body: bytes = None, headers: Dict[str, str] = None
```
dizcology, line 2110: Same comments about the arguments' default values here.

google/cloud/aiplatform/models.py
```diff
@@ -2062,6 +2106,30 @@ def predict(self, instances: List, parameters: Optional[Dict] = None) -> Predict
...
+    ) -> requests.models.Response:
+        """Make a prediction request using arbitrary headers.
+        This method must be called within the network the PrivateEndpoint is peered to.
+        The function call will fail otherwise. To check, use `PrivateEndpoint.network`.
```
dizcology, line 2114: Do we know the specific error that would be raised?  If so perhaps add a `Raises:` section to the docstring: https://google.github.io/styleguide/pyguide.html#doc-function-raises

tests/system/aiplatform/test_model_interactions.py
```diff
@@ -0,0 +1,61 @@
...
+            == prediction_response.model_resource_name
+        )
+
+        # test raw_predict
+        raw_prediction_response = self.endpoint.raw_predict(
+            json.dumps({"instances": [_PREDICTION_INSTANCE]}),
+            {"Content-Type": "application/json"},
+        )
+        assert raw_prediction_response.status_code == 200
+        assert len(json.loads(raw_prediction_response.text).items()) == 1
```
dizcology, line 61: `.items()` is not needed if we are only checking the dictionary's length.

dizcology (changes requested):

google/cloud/aiplatform/constants/base.py
```diff
@@ -92,3 +92,6 @@
 # that is being used for usage metrics tracking purposes.
 # For more details on go/oneplatform-api-analytics
 USER_AGENT_SDK_COMMAND = ""
+
+# Needed for Endpoint.raw_predict
+DEFAULT_AUTHED_SCOPES = ["https://www.googleapis.com/auth/cloud-platform"]
```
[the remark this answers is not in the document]
rosiezou: Yes I believe so, based on the [full list of valid auth scopes](https://developers.google.com/identity/protocols/oauth2/scopes#ml).

google/cloud/aiplatform/models.py
```diff
@@ -1505,29 +1511,67 @@ def predict(
                 [PredictSchemata's][google.cloud.aiplatform.v1beta1.Model.predict_schemata]
                 ``parameters_schema_uri``.
             timeout (float): Optional. The timeout for this request in seconds.
+            use_raw_predict (bool):
+                Optional. If set to True, the underlying prediction call will be made
+                against Endpoint.raw_predict(). Currently, model version information will
```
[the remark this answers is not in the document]
rosiezou: I will remove "currently". I have checked with the service team and they mentioned that it's not on the roadmap to make model version info available for raw_predict due to performance reasons.

google/cloud/aiplatform/models.py
```diff
@@ -35,9 +36,11 @@
 from google.api_core import operation
 from google.api_core import exceptions as api_exceptions
 from google.auth import credentials as auth_credentials
+from google.auth.transport.requests import AuthorizedSession
```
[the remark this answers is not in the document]
rosiezou: updated import statement

google/cloud/aiplatform/models.py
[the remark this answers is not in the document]
rosiezou: updated doc strings

google/cloud/aiplatform/models.py
```diff
@@ -1505,29 +1511,67 @@ def predict(
...
             prediction (aiplatform.Prediction):
                 Prediction with returned predictions and Model ID.
         """
         self.wait()
+        if use_raw_predict:
+            raw_predict_response = self.raw_predict(
+                body=json.dumps({"instances": instances, "parameters": parameters}),
+                headers={"Content-Type": "application/json"},
+            )
+            response_text = json.loads(raw_predict_response.text)
```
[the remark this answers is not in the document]
rosiezou: updated variable name

tests/system/aiplatform/test_model_interactions.py
rosiezou: updated

google/cloud/aiplatform/models.py
dizcology: I meant specifically the part “the function call will fail” in the docstring.

google/cloud/aiplatform/models.py
```diff
@@ -69,6 +72,8 @@
 _DEFAULT_MACHINE_TYPE = "n1-standard-2"
 _DEPLOYING_MODEL_TRAFFIC_SPLIT_KEY = "0"
 _SUCCESSFUL_HTTP_RESPONSE = 300
+_RAW_PREDICT_MODEL_ID_KEY = "X-Vertex-AI-Deployed-Model-Id"
```
dizcology, line 75: Probably call this ”deployed model id” instead of “model id” (which sometimes refers to the last part of the resource full name).

rosiezou: updated variable name

google/cloud/aiplatform/models.py
rosiezou: Got it. I updated the doc strings in `PrivateEndpoint.predict` and `PrivateEndpoint.raw_predict` to contain information about the error code. Most common error code will be 404 with a message saying "request not found".

dizcology (approved):

nayaknishant (approved):

google/cloud/aiplatform/models.py
```diff
@@ -200,6 +205,8 @@ def __init__(
             location=self.location,
             credentials=credentials,
         )
+        self.authorized_session = None
+        self.raw_predict_request_url = None
```
Ark-kun, line 209: Should these attributes be public?

google/cloud/aiplatform/base.py
```diff
@@ -974,7 +974,11 @@ def _sync_object_with_future_result(
             "_gca_resource",
             "credentials",
         ]
-        optional_sync_attributes = ["_prediction_client"]
+        optional_sync_attributes = [
+            "_prediction_client",
+            "_authorized_session",
+            "_raw_predict_request_url",
```
Ark-kun, line 980: These attribute names start with underscores, but the attributes in `models.py` do not have leading underscores. Is this correct?

PR #1640, opened 2022-09-03
Adds enable probabilistic inference for forecasting training jobs.

AndrewLeach:
- [ x] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/python-aiplatform/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [ x] Ensure the tests and linter pass
- [ x] Code coverage does not decrease (if any source code was changed)
- [ x] Appropriate docs were updated (if necessary)

Fixes b/244643824 🦕

### DIFF
--- a/google/cloud/aiplatform/training_jobs.py
+++ b/google/cloud/aiplatform/training_jobs.py
@@ -1793,6 +1793,7 @@ class _ForecastingTrainingJob(_TrainingJob):
         window_stride_length: Optional[int] = None,
         window_max_count: Optional[int] = None,
         holiday_regions: Optional[List[str]] = None,
+        enable_probabilistic_inference: Optional[bool] = None,
         sync: bool = True,
         create_request_timeout: Optional[float] = None,
     ) -> models.Model:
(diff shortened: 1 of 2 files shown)


google/cloud/aiplatform/training_jobs.py
```diff
@@ -2385,6 +2398,15 @@ def _run(
                     EMEA: Europe, the Middle East and Africa
                     LAC: Latin America and the Caribbean
                   Third level: countries from ISO 3166-1 Country codes.
+            enable_probabilistic_inference (bool):
+                If probabilistic inference is enabled, the model will fit a
```
sasha-gitg, line 2402:
```suggestion
                Optional. If probabilistic inference is enabled, the model will fit a
```

google/cloud/aiplatform/training_jobs.py
```diff
@@ -1793,6 +1793,7 @@ def run(
         window_stride_length: Optional[int] = None,
         window_max_count: Optional[int] = None,
         holiday_regions: Optional[List[str]] = None,
+        enable_probabilistic_inference: Optional[bool] = None,
```
sasha-gitg, line 1796: If I'm reading the argument description correctly, it seems be either enabled or disabled. It may make more sense to default to `False` unless `None` has a different meaning than disabled.

sasha-gitg (changes requested):

sasha-gitg:
Closing in favor of #1643

PR #1648, opened 2022-09-07
feat: add eval metrics types to get_experiment_df

sararob:
Add new Model Evaluation metric schema types to `get_experiment_df`. With this addition, calling `get_experiment_df` will return metrics produced by Model Evaluation pipelines.


sasha-gitg (approved):

SinaChavoshi (approved):
I think we should change the type name in schema classes [artifact_schema.py](https://github.com/googleapis/python-aiplatform/blob/main/google/cloud/aiplatform/metadata/schema/google/artifact_schema.py) to use the same constant as definitions in const file. However we can cover that in a separate cl.

PR #1659, opened 2022-09-13
chore: Replace collections.Iterable reference to collections.abc.Iterable for Python 3.10

sasha-gitg:
Fixes [b/246541115](http://b/246541115)

### DIFF
--- a/tests/unit/aiplatform/test_pipeline_jobs.py
+++ b/tests/unit/aiplatform/test_pipeline_jobs.py
@@ -15,7 +15,6 @@
 # limitations under the License.
 #
 
-import yaml
 import pytest
 import json
 
@@ -462,6 +461,8 @@ class TestPipelineJob:
         mock_load_yaml_and_json,
         sync,
     ):
+        import yaml
+
         aiplatform.init(
             project=_TEST_PROJECT,
             staging_bucket=_TEST_GCS_BUCKET_NAME,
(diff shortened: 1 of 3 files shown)


nayaknishant:
I'm wondering is there a reason why the `import yaml` is inside each test function instead of being global? Also if it is inside each function do we need to wrap it in a try-except statement?

tests/unit/aiplatform/test_pipeline_jobs.py
```diff
@@ -462,6 +461,8 @@ def test_run_call_pipeline_service_create(
         mock_load_yaml_and_json,
         sync,
     ):
+        import yaml
```
nayaknishant, line 464: I'm wondering is there a reason why the `import yaml` is inside each test function instead of being global? Also if it is inside each function do we need to wrap it in a try-except statement?

sasha-gitg: Responded above to Jialu's comment on the reasoning.

It's arguable that tests are not intended to be used/depended-on so we don't require as much UX friendly boilerplate on these optional dependency imports.

nayaknishant: Ok, sounds good

nayaknishant (approved):

PR #1665, opened 2022-09-14
chore: change enable_simple_view default to False

sararob:
Changed `enable_simple_view` param in PipelineJob.list() from Optional to `False`.

### DIFF
--- a/google/cloud/aiplatform/pipeline_jobs.py
+++ b/google/cloud/aiplatform/pipeline_jobs.py
@@ -510,9 +510,9 @@ class PipelineJob(
     @classmethod
     def list(
         cls,
-        enable_simple_view: bool = False,
         filter: Optional[str] = None,
         order_by: Optional[str] = None,
+        enable_simple_view: bool = False,
         project: Optional[str] = None,
         location: Optional[str] = None,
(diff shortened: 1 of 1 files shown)


google/cloud/aiplatform/pipeline_jobs.py
```diff
@@ -535,11 +535,11 @@ def list(
                 ascending order. Use "desc" after a field name for descending.
                 Supported fields: `display_name`, `create_time`, `update_time`
             enable_simple_view (bool):
-                Optional. Whether to pass the `read_mask` parameter to the list call.
-                This will improve the performance of calling list(). However, the
-                returned PipelineJob list will not include all fields for each PipelineJob.
-                Setting this to True will exclude the following fields in your response:
-                `runtime_config`, `service_account`, `network`, and some subfields of
+                Whether to pass the `read_mask` parameter to the list call.
```
sasha-gitg, line 538:
```suggestion
                Optional. Whether to pass the `read_mask` parameter to the list call.
```
Indicates the argument has a default and doesn't need to be passed in. Unrelated to the `Optional` type.

sasha-gitg (approved):

PR #1671, opened 2022-09-16
fix: fix endpoint parsing in ModelDeploymentMonitoringJob.update

rosiezou:
Also modifying existing system test to exercise previously timed-out code path


tests/system/aiplatform/test_model_monitoring.py
```diff
@@ -156,22 +176,66 @@ def test_mdm_two_models_one_valid_config(self):
...
 
+        # delete this job and re-configure it to only enable drift detection for faster testing
+        job.delete()
         job_resource = job._gca_resource.name
 
-        # test job update and delete()
-        timeout = time.time() + 3600
-        new_obj_config = model_monitoring.ObjectiveConfig(skew_config)
+        # test job delete
+        with pytest.raises(core_exceptions.NotFound):
```
sasha-gitg, line 184: Preference for each additional test to be added as a new test method. The test methods get executed in order. Reference example: https://github.com/googleapis/python-aiplatform/blob/5fe5ae5ccc58a547060cbe04539be963930d5828/tests/system/aiplatform/test_experiments.py#L44

tests/system/aiplatform/test_model_monitoring.py
```diff
@@ -156,22 +176,66 @@ def test_mdm_two_models_one_valid_config(self):
...
+                drift_detection_config=drift_config
+            ),
+            create_request_timeout=3600,
+            project=e2e_base._PROJECT,
+            location=e2e_base._LOCATION,
+            endpoint=self.endpoint,
+            predict_instance_schema_uri="",
+            analysis_instance_schema_uri="",
+        )
+        assert new_job is not None
```
sasha-gitg, line 202: Is there a case where this could be None and the above line succeeds?

tests/system/aiplatform/test_model_monitoring.py
```diff
@@ -156,22 +176,66 @@ def test_mdm_two_models_one_valid_config(self):
...
+                assert str(gca_obj_config.prediction_drift_detection_config) == ""
+                assert (
+                    gca_obj_config.training_prediction_skew_detection_config
+                    == skew_config.as_proto()
+                )
                 break
-            time.sleep(5)
+        new_job.pause()
+        while new_job.state != gca_job_state.JobState.JOB_STATE_PAUSED:
+            if new_job.state == gca_job_state.JobState.JOB_STATE_PAUSED:
```
sasha-gitg, line 232: It seems like we want to sleep here as well or this will repeatedly hit the service.

google/cloud/aiplatform/jobs.py
```diff
@@ -2485,7 +2485,7 @@ def update(
             current_job.model_deployment_monitoring_objective_configs = (
                 ModelDeploymentMonitoringJob._parse_configs(
                     objective_configs,
-                    current_job.endpoint,
+                    aiplatform.Endpoint(current_job.endpoint),
```
sasha-gitg, line 2488: Please add a unit test to test this path.

sasha-gitg (changes requested):

tests/system/aiplatform/test_model_monitoring.py
rosiezou: I have separated out the test cases

tests/system/aiplatform/test_model_monitoring.py
rosiezou: added sleep

tests/system/aiplatform/test_model_monitoring.py
rosiezou: none that I can think of, but I added job=None at the beginning just to be extra certain

google/cloud/aiplatform/jobs.py
rosiezou: unit test added

google/cloud/aiplatform/jobs.py
```diff
@@ -2485,7 +2485,7 @@ def update(
...
                     deployed_model_ids,
                 )
             )
```
sasha-gitg, line 2491: This call seems to have no effect if the job isn't running but doesn't raise or warn in that case. Also this silent noop behavior isn't documented. This should be changed because it seems like the update is executed successfully when no update is actually performed.

sasha-gitg (changes requested):

google/cloud/aiplatform/jobs.py
rosiezou: Service side will throw an error if the job cannot be updated (i.e. when it's not in RUNNING stage). It doesn't fail silently.

sasha-gitg: It seems like the update is only attempted when RUNNING:

https://github.com/googleapis/python-aiplatform/blob/9a506eeae6be3fa33b262dc7782d0bb4f10bf765/google/cloud/aiplatform/jobs.py#L2492-L2497

tests/system/aiplatform/test_model_monitoring.py
```diff
@@ -156,22 +177,71 @@ def test_mdm_two_models_one_valid_config(self):
...
+                drift_detection_config=drift_config
+            ),
+            create_request_timeout=3600,
+            project=e2e_base._PROJECT,
+            location=e2e_base._LOCATION,
+            endpoint=self.endpoint,
+            predict_instance_schema_uri="",
+            analysis_instance_schema_uri="",
+        )
+        assert job is not None
```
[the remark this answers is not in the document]
rosiezou: removed

google/cloud/aiplatform/jobs.py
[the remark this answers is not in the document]
rosiezou: added

google/cloud/aiplatform/jobs.py
rosiezou: Oh wait you're right. I have removed the if conditions in the update/pause/delete functions to avoid silently masking failed actions. I also added system test coverage for failed update when job is in PENDING state.

tests/unit/aiplatform/test_jobs.py
```diff
@@ -969,3 +977,93 @@ def test_batch_predict_job_with_versioned_model(
...
+class TestModelDeploymentMonitoringJob:
+    def setup_method(self):
+        reload(initializer)
+        reload(aiplatform)
+
+    def teardown_method(self):
+        initializer.global_pool.shutdown(wait=True)
+
+    @pytest.mark.usefixtures(
+        "create_mdm_job_mock", "get_mdm_job_mock", "update_mdm_job_mock"
```
sasha-gitg, line 1057: Why does create need to be mocked?

tests/unit/aiplatform/test_jobs.py
```diff
@@ -969,3 +977,93 @@ def test_batch_predict_job_with_versioned_model(
...
+    )
+    def test_update_mdm_job(self):
+        job = jobs.ModelDeploymentMonitoringJob(
+            model_deployment_monitoring_job_name=_TEST_MDM_JOB_NAME
+        )
+        new_config = aiplatform.model_monitoring.ObjectiveConfig(
+            explanation_config=aiplatform.model_monitoring.ExplanationConfig()
+        )
+        job.update(objective_configs=new_config)
+        assert job._gca_resource.model_deployment_monitoring_objective_configs[
```
sasha-gitg, line 1067: This test should also assert all client requests were made as expected.

tests/unit/aiplatform/test_jobs.py
```diff
@@ -969,3 +977,93 @@ def test_batch_predict_job_with_versioned_model(
...
+        yield mock_deployed_model
+
+
+@pytest.fixture
+def endpoint_list_models_mock(mock_deployed_model):
+    with mock.patch.object(
+        aiplatform.Endpoint, "list_models"
+    ) as endpoint_list_models_mock:
+        endpoint_list_models_mock.return_value = [mock_deployed_model]
+        yield endpoint_list_models_mock
```
sasha-gitg, line 1027: The underlying GAPIC clients should be mocked. The mocked client APIs should return an instance of the proto resource as expected instead of mocking the proto class. The mocks in the endpoint tests can be reused.

tests/system/aiplatform/test_model_monitoring.py
```diff
@@ -107,6 +127,7 @@ def test_mdm_two_models_one_valid_config(self):
         Enable model monitoring on two existing models deployed to the same endpoint.
         """
         # test model monitoring configurations
+        job = None
```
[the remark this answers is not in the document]
rosiezou: removed

tests/system/aiplatform/test_model_monitoring.py
```diff
@@ -156,22 +176,82 @@ def test_mdm_two_models_one_valid_config(self):
...
-        timeout = time.time() + 3600
-        new_obj_config = model_monitoring.ObjectiveConfig(skew_config)
+        # test job delete
+        with pytest.raises(core_exceptions.NotFound):
+            job.api_client.get_model_deployment_monitoring_job(name=job_resource)
 
-        while time.time() < timeout:
+    def test_mdm_pause_and_update_config(self):
+        """Test objective config updates for existing MDM job"""
+        job = None
```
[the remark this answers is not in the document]
rosiezou: removed

tests/system/aiplatform/test_model_monitoring.py
```diff
@@ -156,22 +176,82 @@ def test_mdm_two_models_one_valid_config(self):
...
+            alert_config=alert_config,
+            objective_configs=model_monitoring.ObjectiveConfig(
+                drift_detection_config=drift_config
+            ),
+            create_request_timeout=3600,
+            project=e2e_base._PROJECT,
+            location=e2e_base._LOCATION,
+            endpoint=self.endpoint,
+            predict_instance_schema_uri="",
+            analysis_instance_schema_uri="",
```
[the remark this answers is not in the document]
rosiezou: removed

google/cloud/aiplatform/jobs.py
```diff
@@ -2485,31 +2485,30 @@ def update(
             current_job.model_deployment_monitoring_objective_configs = (
                 ModelDeploymentMonitoringJob._parse_configs(
                     objective_configs,
```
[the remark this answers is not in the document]
rosiezou: updated

tests/unit/aiplatform/test_jobs.py
rosiezou: Good catch. I forgot to delete that code block when I added the mock for getter. We actually don't need that function for this unit test. I have removed it.

tests/unit/aiplatform/test_jobs.py
rosiezou: Hmm somehow when I tried to call `update_mdm_job_mock.assert_called_once...`, I got an attribute error saying `AttributeError: 'function' object has no attribute 'assert_called_once'`. Not sure what I did wrong here because it looks like the fixture is set up similarly as other fixtures in this test file.

tests/system/aiplatform/test_model_monitoring.py
```diff
@@ -156,22 +173,79 @@ def test_mdm_two_models_one_valid_config(self):
...
+                    == new_obj_config.drift_detection_config.as_proto()
+                )
+                break
+
+        # test pause
+        job.pause()
+        while job.state != gca_job_state.JobState.JOB_STATE_PAUSED:
+            time.sleep(1)
+            if job.state == gca_job_state.JobState.JOB_STATE_PAUSED:
+                break
```
sasha-gitg, line 243: It seems like the while condition should suffice.

tests/unit/aiplatform/test_jobs.py
rosiezou: update: I have fixed this issue. It looks like marking the use of a fixture isn't enough to import all the context & I needed to include the fixture as an argument in the test function too.

tests/system/aiplatform/test_model_monitoring.py
rosiezou: removed this code block

tests/unit/aiplatform/test_jobs.py
rosiezou: update: I have removed the endpoint-related mocks and imported them from the endpoint unit tests. I have also opted to use adding drift detection configs as a way to test the update function instead of enabling/disabling XAI. Test coverage for XAI configs already exists in other model monitoring unit and system tests.

tests/unit/aiplatform/test_jobs.py
```diff
@@ -969,3 +981,87 @@ def test_batch_predict_job_with_versioned_model(
...
+                0
+            ].objective_config.prediction_drift_detection_config
+            == drift_detection_config.as_proto()
+        )
+        update_mdm_job_mock.assert_called_once_with(
+            model_deployment_monitoring_job=get_mdm_job_mock.return_value,
+            update_mask=field_mask_pb2.FieldMask(
+                paths=["model_deployment_monitoring_objective_configs"]
+            ),
+        )
```
sasha-gitg, line 1067: Also assert `get_mdm_job_mock` is called as expected since it's not unit tested elsewhere

sasha-gitg (approved):

tests/unit/aiplatform/test_jobs.py
```diff
@@ -969,3 +981,87 @@ def test_batch_predict_job_with_versioned_model(
...
+@pytest.mark.usefixtures("google_auth_mock")
+class TestModelDeploymentMonitoringJob:
+    def setup_method(self):
+        reload(initializer)
+        reload(aiplatform)
+
+    def teardown_method(self):
+        initializer.global_pool.shutdown(wait=True)
+
+    @pytest.mark.usefixtures("get_mdm_job_mock", "update_mdm_job_mock")
```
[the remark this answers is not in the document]
rosiezou: updated

tests/unit/aiplatform/test_jobs.py
rosiezou: updated

PR #1681, opened 2022-09-21
chore: Add headers for CPR model server errors (controled in handler …

abcdefgs0324:
…for data plane).

### DIFF
--- a/google/cloud/aiplatform/prediction/handler.py
+++ b/google/cloud/aiplatform/prediction/handler.py
@@ -16,9 +16,12 @@
 #
 
 from abc import ABC, abstractmethod
+import logging
 from typing import Optional, Type
+import traceback
 
 try:
+    from fastapi import HTTPException
     from fastapi import Request
     from fastapi import Response
 except ImportError:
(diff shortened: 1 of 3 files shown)


google/cloud/aiplatform/prediction/handler.py
```diff
@@ -103,15 +108,56 @@ async def handle(self, request: Request) -> Response:
...
+            )
+
+            # Converts all other exceptions to HTTPException.
+            raise HTTPException(
+                status_code=500,
+                detail=error_message,
+                headers={
+                    prediction.CUSTOM_PREDICTION_ROUTINES_SERVER_ERROR_HEADER_KEY: version.__version__
+                },
+            )
```
kthytang, line 137: Initial thought is that we should wrap the user's code in try, instead of trying to cover all the places our code is executed. Still thinking about the best way to do that.

abcdefgs0324: If I understand correctly, what your thought will look like:
handler:
```
try:
  ...user code...
except:
  raises HTTPException without headers
```
model server:
```
def predict:
  try:
    return self.handler.handle(...)
  except HTTPException:
    raise (without headers)
  except Exception:
    raise with headers
```
The issue is that we allow custom handlers which users may not raise all potential exceptions in their code. That's why I did a reverse way.

rosiezou:
@jaycee-li please review

jaycee-li (approved):
LGTM

Issue #1684, opened 2022-09-21
Fix FeatureStore list_entity_types() and EntityType list_features()


MichaelisTrofficus:
#### Environment details

  - OS type and version:
  - Python version: `3.7.12`
  - pip version: `22.2.2 `
  - `google-cloud-aiplatform` version: `1.16.1`

#### Steps to reproduce

I was trying to list the the EntityTypes and the Features created in a FeatureStore from Vertex AI Workbench (I've already detailed the Environment details). The region I built the FeatureStore into is `europe-west4`, son when I try to list the entities and the features it keeps failing (more details in the Stack Trace and Code Example).

#### Code example

```python
PROJECT_ID = "my-project"
REGION = "europe-west4"
FEATURESTORE_ID = "my_feature_store_id"

fs = FeaturestoreCustom(
    featurestore_name=FEATURESTORE_ID,
    project=PROJECT_ID,
    location=REGION,
)

fs.list_entity_types() # It fails here


entity_type = fs.get_entity_type(entity_type_id="my_entity_type_id")
entity_type.list_features() # It also fails here
```

#### Stack trace
```
InvalidArgument: 400 List of found errors:	1.Field: parent; Message: The provided location ID doesn't match the endpoint. The valid location ID is `us-central1`.	 [field_violations {
  field: "parent"
  description: "The provided location ID doesn\'t match the endpoint. The valid location ID is `us-central1`."
}
]
```

I'm going to drop a PR to solve this issue. 

Thanks in advance!!

rosiezou:
Hi @MichaelisTrofficus. Developers are encouraged to use `aiplatform.init(...)` to set global configs for project & location since it'll facilitate the rest of the development workflow. It will also resolve the error that you're seeing. You will have to manually pass in project and location for most other SDK function calls pertaining to getting or creating resources if you don't set them globally. I'm closing this issue & marking it as intended behavior.

PR #1700, opened 2022-09-27
fix: project/location parsing for nested resources

nayaknishant:
For nested resources like EntityType (Featurestore -> EntityType -> Feature), the project/location needs to be extracted from the full resource name and passed into the client constructor. Adding check to see if parent exists and utility method to extract project and location from the parent resource name.
 
Fixes [b/249363374](https://b.corp.google.com/issues/249363374) and https://github.com/googleapis/python-aiplatform/issues/1684 🦕

### DIFF
--- a/google/cloud/aiplatform/utils/__init__.py
+++ b/google/cloud/aiplatform/utils/__init__.py
@@ -324,6 +324,32 @@ def extract_bucket_and_prefix_from_gcs_path(gcs_path: str) -> Tuple[str, Optiona
 
     return (gcs_bucket, gcs_blob_prefix)
 
+def extract_project_and_location_from_parent(parent: str) -> Tuple[str, str]:
+    """Given a complete parent resource name, return the project and location as a tuple.
+
+    Example Usage:
+
+        project, location = extract_project_and_location_from_parent(
+            "projects/123/locations/us-central1/datasets/456"
+        )
+
+        project = "123"
+        location = "us-central1"
+
+    Args:
+        parent (str):
+            Required. A complete parent resource name.
+
+    Returns:
+        Tuple[str, str]
+            A (project, location) pair from provided parent resource name.
+    """
+    parent_parts = parent.split("/", 4)
+    parent_project =  None if len(parent_parts) == 1 else parent_parts[1]
+    parent_location = None if len(parent_parts) < 4 else parent_parts[3]
+
+    return (parent_project, parent_location)
+
 
 class ClientWithOverride:
     class WrappedClient:
--- a/google/cloud/aiplatform/utils/__init__.py
+++ b/google/cloud/aiplatform/utils/__init__.py
@@ -327,7 +327,7 @@ def extract_bucket_and_prefix_from_gcs_path(gcs_path: str) -> Tuple[str, Optiona
 
 def extract_project_and_location_from_parent(
     parent: str,
-) -> Dict[Optional[str], Optional[str]]:
+) -> Dict[str, str]:
     """Given a complete parent resource name, return the project and location as a dict.
 
     Example Usage:
@@ -344,7 +344,7 @@ def extract_project_and_location_from_parent(
             Required. A complete parent resource name.
 
     Returns:
-        Dict[Optional[str], Optional[str]
+        Dict[str, str]
             A project, location dict from provided parent resource name.
     """
     parent_resources = re.match(
--- a/google/cloud/aiplatform/utils/__init__.py
+++ b/google/cloud/aiplatform/utils/__init__.py
@@ -324,7 +324,10 @@ def extract_bucket_and_prefix_from_gcs_path(gcs_path: str) -> Tuple[str, Optiona
 
     return (gcs_bucket, gcs_blob_prefix)
 
-def extract_project_and_location_from_parent(parent: str) -> Dict[Optional[str], Optional[str]]:
+
+def extract_project_and_location_from_parent(
+    parent: str,
+) -> Dict[Optional[str], Optional[str]]:
     """Given a complete parent resource name, return the project and location as a dict.
 
     Example Usage:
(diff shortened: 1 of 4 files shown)


google/cloud/aiplatform/utils/__init__.py
```diff
@@ -324,6 +324,32 @@ def extract_bucket_and_prefix_from_gcs_path(gcs_path: str) -> Tuple[str, Optiona
...
+
+    Args:
+        parent (str):
+            Required. A complete parent resource name.
+
+    Returns:
+        Tuple[str, str]
+            A (project, location) pair from provided parent resource name.
+    """
+    parent_parts = parent.split("/", 4)
```
sasha-gitg, line 347: Preference to adapt regex similar to https://github.com/googleapis/python-aiplatform/blob/main/google/cloud/aiplatform_v1/services/migration_service/client.py#L395 to handle both location paths and arbitrary resource paths so there's additional validation on the parent.

google/cloud/aiplatform/utils/__init__.py
```diff
@@ -324,6 +324,32 @@ def extract_bucket_and_prefix_from_gcs_path(gcs_path: str) -> Tuple[str, Optiona
...
+    parent_project =  None if len(parent_parts) == 1 else parent_parts[1]
+    parent_location = None if len(parent_parts) < 4 else parent_parts[3]
+
+    return (parent_project, parent_location)
+
```
sasha-gitg, line 352: Include unit tests for the utility as well as the exact bug this is fixing; https://github.com/googleapis/python-aiplatform/issues/1684

sasha-gitg (changes requested):

google/cloud/aiplatform/utils/__init__.py
```diff
@@ -324,6 +324,32 @@ def extract_bucket_and_prefix_from_gcs_path(gcs_path: str) -> Tuple[str, Optiona
 
     return (gcs_bucket, gcs_blob_prefix)
 
+def extract_project_and_location_from_parent(parent: str) -> Tuple[str, str]:
```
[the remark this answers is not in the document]
nayaknishant: Fixed

google/cloud/aiplatform/utils/__init__.py
nayaknishant: Converted to regex check instead of using Python string `split` method

google/cloud/aiplatform/utils/__init__.py
```diff
@@ -325,6 +325,34 @@ def extract_bucket_and_prefix_from_gcs_path(gcs_path: str) -> Tuple[str, Optiona
     return (gcs_bucket, gcs_blob_prefix)
 
 
+def extract_project_and_location_from_parent(
+    parent: str,
+) -> Dict[Optional[str], Optional[str]]:
```
dizcology, line 330: `Optional[str]` is a synonym for `Union[str, None]`, which would imply that the return value could have `None` in the key or value.  The return type should simply be `Dict[str, str]` here.

google/cloud/aiplatform/utils/__init__.py
```diff
@@ -325,6 +325,34 @@ def extract_bucket_and_prefix_from_gcs_path(gcs_path: str) -> Tuple[str, Optiona
...
+
+        parent_resources["project"] = "123"
+        parent_resources["location"] = "us-central1"
+
+    Args:
+        parent (str):
+            Required. A complete parent resource name.
+
+    Returns:
+        Dict[Optional[str], Optional[str]
```
dizcology, line 347: Same comment here.

google/cloud/aiplatform/utils/__init__.py
```diff
@@ -325,6 +325,34 @@ def extract_bucket_and_prefix_from_gcs_path(gcs_path: str) -> Tuple[str, Optiona
...
+            A project, location dict from provided parent resource name.
+    """
+    parent_resources = re.match(
```
dizcology, line 350: Is this code basically copied from the GAPIC library such as this?  https://github.com/googleapis/python-aiplatform/blob/main/google/cloud/aiplatform_v1/services/dataset_service/client.py#L377

Why not just use that instead?

dizcology (changes requested):

tests/unit/aiplatform/test_featurestores.py
```diff
@@ -1023,7 +1023,25 @@ def test_list_entity_types(self, list_entity_types_mock):
         aiplatform.init(project=_TEST_PROJECT)
```
sasha-gitg, line 1023: To recreate the failure here: https://github.com/googleapis/python-aiplatform/issues/1684

You need an additional test that doesn't use `init` and passes in featurestore_id, project, and location.

sasha-gitg (changes requested):

google/cloud/aiplatform/utils/__init__.py
nayaknishant: When that code extracts the project and location from the string `"projects/123/locations/us-central1/datasets/456"` it extracts it as `{"project" : "123", "location" : "us-central1/datasets/456"}`, it takes everything remaining after `locations/`. My regex is based on the one you and Sasha sent, but the end parsing is different to only grab `us-central1`.

google/cloud/aiplatform/utils/__init__.py
nayaknishant: The function returns either a `Dict[str, str]` or an empty dict `{}`. Does having `Dict[str, str]` also capture `{}`. But sure, I'll add the change

google/cloud/aiplatform/utils/__init__.py
nayaknishant: Fixed

tests/unit/aiplatform/test_featurestores.py
nayaknishant: Good catch Sasha, added a test for EntityType `list_features` as well that doesn't use `init()`

google/cloud/aiplatform/utils/__init__.py
dizcology: That is correct, the typing annotation `Dict[str, str]` includes the empty dict.

The confusion here is that the word `Optional` does not mean what it means in English: https://docs.python.org/3/library/typing.html#typing.Optional, specifically "Note that this is not the same concept as an optional argument..."

nayaknishant: Gotcha, thanks Yuhan

google/cloud/aiplatform/utils/__init__.py
dizcology: I missed that - thanks for the explanation!

dizcology (approved):

sasha-gitg (approved):

PR #1702, opened 2022-09-28
feat: adding global network arg

nayaknishant:
Previously, network could be set explicitly in specific methods in `models.py` (PrivateEndpoint) and certain job requests. Adding a global network arg that can be set with 
```python
aiplatform.init(
    project="your-project-id"
    location="us-central1"
    network="projects/123456789123/global/networks/my_vpc"
)
```
Will require adding a helper function with `@base.optional_sync` decorator to certain methods to ensure proper sync of the global `network` arg

Fixes [b/211371627](https://b.corp.google.com/issues/211371627) 🦕


google/cloud/aiplatform/initializer.py
```diff
@@ -130,6 +138,8 @@ def init(
             self._credentials = credentials
         if encryption_spec_key_name:
             self._encryption_spec_key_name = encryption_spec_key_name
+        if network:
```
dizcology, line 141: To test against `None` default values, use `is not None` instead.  So here this should be `if network is not None: ...`.

tests/unit/aiplatform/test_initializer.py
```diff
@@ -90,6 +91,10 @@ def test_init_location_with_invalid_location_raises(self):
         with pytest.raises(ValueError):
             initializer.global_config.init(location=_TEST_INVALID_LOCATION)
 
+    def test_init_network_sets_network(self):
```
dizcology, line 94: This only tests whether the `_network` attribute is correctly set in the config.  Can we add a test where that attribute is used somewhere downstream too?

dizcology (changes requested):

google/cloud/aiplatform/models.py
```diff
@@ -3421,8 +3422,7 @@ def _deploy(
                 Optional. The full name of the Compute Engine network to which
                 this Endpoint will be peered. E.g. "projects/12345/global/networks/myVPC".
                 Private services access must already be configured for the network.
-                If left unspecified, the job is not peered with any network or
```
sasha-gitg, line 3424: This is actually not true on this method.

google/cloud/aiplatform/pipeline_jobs.py
```diff
@@ -304,12 +304,15 @@ def run(
...
+                If left unspecified, the network set in aiplatform.init will be used.
+                Otherwise, the Endpoint is not peered with any network.
             sync (bool):
                 Optional. Whether to execute this method synchronously. If False, this method will unblock and it will be executed in a concurrent Future.
             create_request_timeout (float):
                 Optional. The timeout for the create request in seconds.
         """
+        network = network or initializer.global_config.network
+
         self.submit(
```
sasha-gitg, line 316: `submit` should also be updated

google/cloud/aiplatform/pipeline_jobs.py
```diff
@@ -304,12 +304,15 @@ def run(
...
                 Private services access must already be configured for the network.
-                If left unspecified, the job is not peered with any network.
+                If left unspecified, the network set in aiplatform.init will be used.
+                Otherwise, the Endpoint is not peered with any network.
             sync (bool):
                 Optional. Whether to execute this method synchronously. If False, this method will unblock and it will be executed in a concurrent Future.
             create_request_timeout (float):
                 Optional. The timeout for the create request in seconds.
         """
+        network = network or initializer.global_config.network
```
sasha-gitg, line 314: Note that for methods decorated with `optional_sync` and with `sync` parameter the default argument needs to be captured synchronously at call time. Otherwise the argument may change between call time and execution time if the use uses `init(network=...)` again. So something like

```Python

def run( self, ..., network, sync=True):
    network = network of initializer.global_config.network
    return self._run(..., network=network, sync=sync)

@base.optional_sync
def _run(..., network, sync=True):
  ...
```

sasha-gitg (changes requested):

google/cloud/aiplatform/initializer.py
nayaknishant: Fixed

google/cloud/aiplatform/models.py
nayaknishant: Changed docstring to Endpoint instead of job. If left unspecified (kept as None), wouldn't the Endpoint not be peered with any network?

google/cloud/aiplatform/pipeline_jobs.py
nayaknishant: Fixed

google/cloud/aiplatform/pipeline_jobs.py
nayaknishant: Hey Sasha, I checked through the files that would be affected by this (`jobs.py`, `pipeline_jobs.py`, `training_jobs.py`, `models.py`, `matching_engine_index_endpoint.py`) and they all have explicit passing of the `network` arg into their associated helper methods. So, I think they are good to go

dizcology: Can we look into this some more?  I think Sasha is saying the public method should not be decorated with `optional_sync`, and the public method should call a private method decorated with `optional_sync`.

nayaknishant: hey @sasha-gitg can we get clarification on this?

google/cloud/aiplatform/models.py
sasha-gitg: The purpose of this comment was to capture that `_deploy` does not default to the `network` set in `init` but `deploy` does. 

ie: This line should be added back in for `deploy` and left out for `_deploy`:
```
If left unspecified, the job is not peered with any network or
                the network set in aiplatform.init will be used.
```

Preference to change to something like this because the statement above is slightly ambiguous:
```
If left unspecified, the network set in aiplatform.init will be used.
```

google/cloud/aiplatform/pipeline_jobs.py
sasha-gitg: Yu-han is right. The current implementation, without the proposed change above, could result in thisunintened behavior:

```Python
aiplatform.init(project='my-project')

job = aiplatform.PipelineJob(...)
job.run(..., sync=False)

aipaltform.init(network='my_network') # job is potentially executed with this network which is not the intention
```

sasha-gitg (changes requested):

google/cloud/aiplatform/models.py
nayaknishant: Fixed. Left out the lines about `aiplatform.init(...)` in the helper functions and tried to be more specific with the docstrings for the public, main functions (`deploy` for Model and `create` for PrivateEndpoint)

google/cloud/aiplatform/pipeline_jobs.py
nayaknishant: Fixe, added and moved `@base.optional_sync` to a helper method to capture network arg synchronously (`network = network of initializer.global_config.network`) to `matching_engine_index_endpoint.py`, `pipeline_jobs.py`, and `jobs.py`

google/cloud/aiplatform/pipeline_jobs.py
```diff
@@ -352,8 +385,8 @@ def submit(
         if service_account:
             self._gca_resource.service_account = service_account
 
-        if network:
-            self._gca_resource.network = network
+        if network is not None:
+            self._gca_resource.network = network or initializer.global_config.network
```
sasha-gitg, line 389: This seems to not set `network` from `initializer.gloabel_config` by default.

sasha-gitg (changes requested):

google/cloud/aiplatform/matching_engine/matching_engine_index_endpoint.py
```diff
@@ -201,10 +200,90 @@ def create(
             MatchingEngineIndexEndpoint - IndexEndpoint resource object
 
         """
-        gapic_index_endpoint = gca_matching_engine_index_endpoint.IndexEndpoint(
+        network = network or initializer.global_config.network
```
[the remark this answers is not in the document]
nayaknishant: Fixed, added a ValueError for when `network` is not passed into IndexEndpoint.create()

google/cloud/aiplatform/matching_engine/matching_engine_index_endpoint.py
```diff
@@ -201,10 +200,90 @@ def create(
...
+            labels=labels,
+            project=project,
+            location=location,
+            credentials=credentials,
+            request_metadata=request_metadata,
+            sync=sync,
+        )
+
+    @base.optional_sync()
+    @classmethod
```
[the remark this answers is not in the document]
nayaknishant: Fixes

google/cloud/aiplatform/matching_engine/matching_engine_index_endpoint.py
```diff
@@ -201,10 +200,90 @@ def create(
             MatchingEngineIndexEndpoint - IndexEndpoint resource object
 
         """
-        gapic_index_endpoint = gca_matching_engine_index_endpoint.IndexEndpoint(
+        network = network or initializer.global_config.network
+
+        cls._create(
```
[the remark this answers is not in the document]
nayaknishant: Fixed

google/cloud/aiplatform/jobs.py
```diff
@@ -1570,7 +1570,73 @@ def run(
...
+            network=network,
+            timeout=timeout,
+            restart_job_on_worker_restart=restart_job_on_worker_restart,
+            enable_web_access=enable_web_access,
+            tensorboard=tensorboard,
+            sync=sync,
+            create_request_timeout=create_request_timeout
+        )
+
+    @base.optional_sync
```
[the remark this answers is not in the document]
nayaknishant: Fixed

google/cloud/aiplatform/pipeline_jobs.py
nayaknishant: Fixed, set `network` argument beforehand and check if `network` arg has been set in the if statement

dizcology (approved):

google/cloud/aiplatform/matching_engine/matching_engine_index_endpoint.py
```diff
@@ -201,10 +199,96 @@ def create(
             MatchingEngineIndexEndpoint - IndexEndpoint resource object
 
         """
-        gapic_index_endpoint = gca_matching_engine_index_endpoint.IndexEndpoint(
+        network = network or initializer.global_config.network
+
+        if not network:
+            raise ValueError(
+                "Please provide `network` argument or set "
```
sasha-gitg, line 206:
```suggestion
                "Please provide `network` argument or set network "
```

sasha-gitg (approved):

google/cloud/aiplatform/matching_engine/matching_engine_index_endpoint.py
```diff
@@ -201,10 +199,96 @@ def create(
             MatchingEngineIndexEndpoint - IndexEndpoint resource object
 
         """
-        gapic_index_endpoint = gca_matching_engine_index_endpoint.IndexEndpoint(
+        network = network or initializer.global_config.network
+
+        if not network:
+            raise ValueError(
```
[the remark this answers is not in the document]
nayaknishant: Added section to docstring

google/cloud/aiplatform/matching_engine/matching_engine_index_endpoint.py
nayaknishant: Fixed

IronPan (approved):

sasha-gitg (approved):

sasha-gitg (approved):

PR #1707, opened 2022-09-29
docs: show inherited SDK methods in pydoc

jaycee-li:
Fixes b/249069357

Show inherited methods for all the classes in SDK. And delete repeated methods for `AutoMLForecastingTrainingJob` and `SequenceToSequencePlusForecastingTrainingJob`


jaycee-li:
Please download and view the [html](https://drive.google.com/file/d/1y_lzBPTDo-uSM9xjuxxouYK43foSBE7M/view?usp=sharing&resourcekey=0-Ol2ankH2_om3MpoDnoNk4w) file for the updated pydoc page.

sasha-gitg (approved):

jaycee-li:
@dandhlee  Thanks for the review! What I'm trying to change is [this page](https://googleapis.dev/python/aiplatform/latest/aiplatform/services.html). I'm not familiar with c.g.c docs so please let me know if deleting those two methods breaks c.g.c

dandhlee:
I don't think it'll break anything since you're only removing duplicate entries. The inheritance may not show, but that's on me - don't have to worry for the scope of this PR 😄

jaycee-li:
> I don't think it'll break anything since you're only removing duplicate entries. The inheritance may not show, but that's on me - don't have to worry for the scope of this PR 😄

Got it. Thank you!

nayaknishant (approved):

PR #1717, opened 2022-10-05
fix: PipelineJob should only pass bearer tokens for AR URIs

TheMichaelHu:
When downloading compiled KFP pipelines over HTTPS, we only need to pass a bearer token when we need to authenticate for services like Artifact Registry. We may get unexpected behavior passing this token in all HTTPS requests, which is the current behavior.

Fixes b/251143831 🦕


chensun (approved):
/lgtm

PR #1718, opened 2022-10-05
fix: added proto message conversion to MDMJob.update fields

rosiezou:

Fixes #<251194445> 🦕


sasha-gitg (changes requested):

tests/unit/aiplatform/test_jobs.py
```diff
@@ -1065,6 +1095,15 @@ def test_update_mdm_job(self, get_mdm_job_mock, update_mdm_job_mock):
         update_mdm_job_mock.assert_called_once_with(
             model_deployment_monitoring_job=get_mdm_job_mock.return_value,
```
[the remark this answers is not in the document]
rosiezou: `get_mdm_job_mock` is only called once at the beginning of `ModelDeploymentJob.update`, then the underlying `return_value` is used and modified in the `update` function during test execution. You can insert debug print statements during multiple points in the test execution to verify that only a single MDM job proto (the one mocked out by `get_mdm_job_mock`) is created and updated throughout the test.

tests/unit/aiplatform/test_jobs.py
```diff
@@ -992,44 +998,46 @@ def get_mdm_job_mock():
...
-                drift_thresholds={
-                    "TEST_KEY": gca_model_monitoring_compat.ThresholdConfig(value=0.01)
-                }
-            )
-        )
-        all_configs = []
-        for model in get_endpoint_with_models_mock.return_value.deployed_models:
-            all_configs.append(
+        expected_output = gca_model_deployment_monitoring_job_compat.ModelDeploymentMonitoringJob(
+            display_name="NEW_NAME",
```
sasha-gitg, line 1013: Preference to use the constant.

tests/unit/aiplatform/test_jobs.py
```diff
@@ -1049,10 +1057,40 @@ def test_update_mdm_job(self, get_mdm_job_mock, update_mdm_job_mock):
...
+            display_name=display_name,
+            schedule_config=schedule_config,
+            alert_config=alert_config,
+            logging_sampling_strategy=sampling_strategy,
+            labels=labels,
+            bigquery_tables_log_ttl=log_ttl,
+            enable_monitoring_pipeline_logs=True,
+            objective_configs=new_config,
+        )
+        gapic_job = job._gca_resource
```
sasha-gitg, line 1083: It seems like we are testing that the `_gca_resource` is updated but it doesn't seem to be assigned after the update call here: 
https://github.com/googleapis/python-aiplatform/blob/0cdfc9b7cf784317bb810174862d2ea9c2e9497e/google/cloud/aiplatform/jobs.py#L2499-L2504

sasha-gitg (changes requested):

tests/unit/aiplatform/test_jobs.py
rosiezou: Interesting..I added `print(self._gca_resource)` right before line 2503 and ran both the unit and system tests to double check the contents. Both debug prints successfully showed the updated job contents. I also didn't explicitly run `_sync_resource` on the existing job in the system test and just queried `_gca_resource` directly to verify the update results and it also worked. My guess is that because I'm not decorating the model monitoring methods with `optional_sync()`, the resource is sync'd whenever there is a change.
https://github.com/googleapis/python-aiplatform/blob/0cdfc9b7cf784317bb810174862d2ea9c2e9497e/tests/system/aiplatform/test_model_monitoring.py#L227-L235

rosiezou:
Closing this PR for now. Will re-open once SDK maintenance work is completed.

tests/unit/aiplatform/test_jobs.py
```diff
@@ -988,48 +1037,22 @@ def get_mdm_job_mock():
...
-                endpoint=_TEST_ENDPOINT,
-            )
-        )
+        get_mdm_job_mock.side_effect = [
+            _TEST_MDM_OLD_JOB,
+            _TEST_MDM_OLD_JOB,
+            _TEST_MDM_OLD_JOB,
+            _TEST_MDM_EXPECTED_NEW_JOB,
+            _TEST_MDM_EXPECTED_NEW_JOB,
+        ]
```
rosiezou, line 1046: This test fixture mocks out the states of the MDM job during the execution of `test_update_mdm_job`. The pytest mock of `get_model_deployment_monitoring_job` is called in the following order:
1. at the beginning of the test definition, `job = jobs.ModelDeploymentMonitoringJob(...)`, during init call of underlying base job class
2.  at the beginning of the test definition, `job = jobs.ModelDeploymentMonitoringJob(...)`, during init call of the `ModelDeploymentMonitoringJob` class
3. at the beginning of `MDMJob.update()`, during `self._sync_gca_resource()`
4. at the beginning of `MDMjob.update()`, during `current_job = self.api_client.get_model_deployment_monitoring_job`
5. at the end of `MDMJob.update()`, during re-assignment of `self._gca_resource`

tests/unit/aiplatform/test_jobs.py
```diff
@@ -988,48 +1041,22 @@ def get_mdm_job_mock():
...
-                state=_TEST_JOB_STATE_RUNNING,
-                endpoint=_TEST_ENDPOINT,
-            )
-        )
+        get_mdm_job_mock.side_effect = [
+            _TEST_MDM_OLD_JOB,
+            _TEST_MDM_OLD_JOB,
+            _TEST_MDM_OLD_JOB,
+            _TEST_MDM_OLD_JOB,
+            _TEST_MDM_EXPECTED_NEW_JOB,
```
rosiezou, line 1049: This test fixture mocks out the states of the MDM job during the execution of `test_update_mdm_job`. The pytest mock of `get_model_deployment_monitoring_job` is called in the following order:
1. at the beginning of the test definition, `job = jobs.ModelDeploymentMonitoringJob(...)`, during init call of underlying base job class
2.  at the beginning of the test definition, `job = jobs.ModelDeploymentMonitoringJob(...)`, during init call of the `ModelDeploymentMonitoringJob` class
3. at the beginning of `MDMJob.update()`, during `self._sync_gca_resource()`
4. at the beginning of `MDMjob.update()`, during `current_job = self.api_client.get_model_deployment_monitoring_job`
5. at the end of `MDMJob.update()`, during re-assignment of `self._gca_resource`

google/cloud/aiplatform/jobs.py
```diff
@@ -2495,6 +2501,9 @@ def update(
             model_deployment_monitoring_job=current_job,
             update_mask=field_mask_pb2.FieldMask(paths=update_mask),
         )
+        self._gca_resource = self.api_client.get_model_deployment_monitoring_job(
```
sasha-gitg, line 2504: The update call returns an `Operation`. The result of the `Operation` is the `ModelDeploymentMonitoringJob`. Reference:

https://github.com/googleapis/python-aiplatform/blob/6dc5f12c771c05efb4ec304054400b1b887a0980/google/cloud/aiplatform_v1/services/job_service/client.py#L3599

We should:
1. Block on this operation until we add optional_sync support (that can be captured as a TODO with reference to a ticket)
2. Assign the gca_resource to the result of that operation.

If we do not then the local object  likely becomes out of sync with the service after this update call. 

This should work:

```Python
operation = self.api_client.update_model_deployment_monitoring_job(
            model_deployment_monitoring_job=current_job,
            update_mask=field_mask_pb2.FieldMask(paths=update_mask),
        )

self._gca_resource = operation.result()
```

google/cloud/aiplatform/jobs.py
```diff
@@ -2456,25 +2457,30 @@ def update(
         current_job = self.api_client.get_model_deployment_monitoring_job(
```
sasha-gitg, line 2457: Since, we are calling sync above, which calls get underneath, there is no need to call get here. The self._gca_resource proto can be copied.

tests/unit/aiplatform/test_jobs.py
```diff
@@ -988,48 +1041,22 @@ def get_mdm_job_mock():
...
-        update_mdm_job_mock.return_vaue.result_type = (
-            gca_model_deployment_monitoring_job_compat.ModelDeploymentMonitoringJob(
-                name=_TEST_MDM_JOB_NAME,
-                display_name=_TEST_DISPLAY_NAME,
-                state=_TEST_JOB_STATE_RUNNING,
-                endpoint=_TEST_ENDPOINT,
-                model_deployment_monitoring_objective_configs=all_configs,
-            )
-        )
+        update_mdm_job_mock.return_value.result_type = _TEST_MDM_EXPECTED_NEW_JOB
```
sasha-gitg, line 1059: `result_type` is intended to be a type not an instance of that type. Refer:

https://github.com/googleapis/python-api-core/blob/main/google/api_core/operation.py#L60

Here is an example of mocking an lro: https://github.com/googleapis/python-aiplatform/blob/main/tests/unit/aiplatform/test_datasets.py#L370

sasha-gitg (changes requested):

google/cloud/aiplatform/jobs.py
rosiezou: done

google/cloud/aiplatform/jobs.py
rosiezou: done

tests/unit/aiplatform/test_jobs.py
rosiezou: done

google/cloud/aiplatform/jobs.py
```diff
@@ -64,6 +64,7 @@
     gca_job_state.JobState.JOB_STATE_FAILED,
     gca_job_state.JobState.JOB_STATE_CANCELLED,
     gca_job_state.JobState.JOB_STATE_PAUSED,
+    gca_job_state.JobState.JOB_STATE_RUNNING,
```
sasha-gitg, line 67: This seems like it will break usage in other jobs and isn't consistent with the intention of this variable.

google/cloud/aiplatform/jobs.py
```diff
@@ -2491,10 +2496,13 @@ def update(
                     deployed_model_ids=deployed_model_ids,
                 )
             )
-        self.api_client.update_model_deployment_monitoring_job(
+        operation = self.api_client.update_model_deployment_monitoring_job(
             model_deployment_monitoring_job=current_job,
             update_mask=field_mask_pb2.FieldMask(paths=update_mask),
         )
+        # TODO: b/254285776 add optional_sync support to model monitoring job
+        self._block_until_complete()
```
sasha-gitg, line 2504: This shouldn't be necessary. We should only rely on the operation completing through `operation.result()`.

sasha-gitg (changes requested):

google/cloud/aiplatform/jobs.py
rosiezou: Actually, I looked into this again and we should not block the update call at all. I had an offline chat with Jing a while back and she said that depending on what is requested in the update operation, it could take anywhere between minutes and hours for it to complete (by completing I'm referring to the job state going from UPDATING back to RUNNING). It does not make sense for the SDK to block until the operation is complete, especially when the update request is for something time consuming like updating the training dataset used for skew detection. 

The proto representation of the job in the update response will already contain the latest configuration, so it's sufficient to just re-assign that back to `self._gca_resource`. As for when the MDM job will resume running with the updated configs, that's a service-side behavior. The GAPIC code only schedules for the update operation to be executed and does not block on complete, so I don't think the SDK should deviate from that behavior. It also isn't necessary from a UX perspective because developers most likely wouldn't need to do anything else on the monitoring job immediately after updating it. As of right now, if users were to do something immediately after `job.update`, for instance calling `job.pause`, GAPIC will return a gRPC error telling the users that the job cannot be paused when it's in `UPDATING` state. So I think the function behavior is correct as-is.

google/cloud/aiplatform/jobs.py
```diff
@@ -2491,10 +2495,11 @@ def update(
                     deployed_model_ids=deployed_model_ids,
                 )
             )
-        self.api_client.update_model_deployment_monitoring_job(
+        response = self.api_client.update_model_deployment_monitoring_job(
```
sasha-gitg, line 2498: We should add the TODO ticket reference back to to make this method optionally synchronous because this can still be executed asynchronously w.r.t to the update LRO.

Also, preference to capture that this is an lro when making the service call. ie:

```Python
lro = self.api_client.update_model_deployment_monitoring_job(...)
self._gca_resource = lro.result()
```

sasha-gitg (approved):

google/cloud/aiplatform/jobs.py
rosiezou: done

PR #1721, opened 2022-10-09
Added us-west3 for tables new region

yishan-pu:


sararob:
Thank you for this contribution! To get the checks to pass, could you do the following:
* Update your commit message to use a conventional commits [prefix](https://www.conventionalcommits.org/en/v1.0.0/#summary), for example: `feat: add us-west3 for tables new region`. This will be used in our release notes
* Change the string used in [this test](https://github.com/googleapis/python-aiplatform/blob/main/tests/unit/aiplatform/test_utils.py#L80-L82) to an unsupported region so that test passes

PR #1742, opened 2022-10-19
fix: Error when only log confusion matrix or ROC curve through log_classification_metrics

jaycee-li:
Fix bug: Get error if users only log confusion matrix or ROC curve through `log_classification_metrics`


sasha-gitg (approved):
Preference to add unit tests that test:
1. Passing just labels and metrics
2. Passing just fpr,tpr,and threshold

Can be added in a follow up PR to unblock this fix.

jialuzh (approved):
+1, let's add some tests to cover this bug.

PR #1872, opened 2022-12-29
feat: add SA parameter to BPJ create method

cymarechal-devoteam:
A BatchPredictionJob could not be created with a service account specified.
This PR solves this issue by using the field available in the proto definition.


google/cloud/aiplatform/jobs.py
```diff
@@ -395,6 +395,7 @@ def create(
         encryption_spec_key_name: Optional[str] = None,
         sync: bool = True,
         create_request_timeout: Optional[float] = None,
+        service_account: Optional[str] = None,
```
sasha-gitg, line 398: Please move this to the end of the signature to avoid order of arguments breaking change.

sasha-gitg (approved):

cymarechal-devoteam:
@sasha-gitg not sure why unit tests failed on the last commit... is there anything I can do still?

sasha-gitg (approved):

sasha-gitg:
Merged in. Thanks @cymarechal-devoteam!

PR #2037, opened 2023-03-24
fix: change the parameters to fit the current sklearn predict module

RochaOwng:
This code snippet was giving me an error:

"google.api_core.exceptions.FailedPrecondition: 400 Prediction failed: Exception during sklearn prediction: predict() got an unexpected keyword argument 'parameters'."

After removing the parameters step, I was able to use the endpoint to predict from a local model. Additionally, I suggest returning the predictions rather than printing them, so that the user can decide what to do with the response. With these changes, I can perform the following code:

```
pred = predict_custom_trained_model_sample(
    project=project,
    endpoint_id=endrpoind_id,
    location=location,
    instances=[[5.1, 3.5, 1.4, 0.2], [5.1, 3.5, 1.4, 0.2]]
)
```


matthew29tang:
Closing due to inactivity.

PR #2005, opened 2023-03-04
fix: Remove packaging package upper bound

r-richmond:
Thank you for opening a Pull Request! Before submitting your PR, there are a few things you can do to make sure it goes smoothly:
- [X] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/python-aiplatform/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [x] Ensure the tests and linter pass
- [x] Code coverage does not decrease (if any source code was changed)
- [x] Appropriate docs were updated (if necessary)
- [x] Get the necessary approvals
- [x] Once the last commit on the PR has been approved, add the "ready to pull" label to the Pull Request

Note: do not merge your PR from GitHub. Adding the "ready to pull" label is the final step in the review process.
After approvals, the changes in your PR will be committed to the `main` branch and this PR will be closed.

Fixes #2004 🦕

PR #2078, opened 2023-04-10
chore: Update Dependencies in setup.py

holtskinner:
Change `packaging` to support `packaging-23.0`

To avoid error 

```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
google-cloud-aiplatform 1.23.0 requires packaging<22.0.0dev,>=14.3, but you have packaging 23.0 which is incompatible.
```

Other Google Cloud Client Libraries support or require a newer version


holtskinner:
Closing this - PR #2005 solves this issue

PR #2137, opened 2023-04-27
Removed api_endpoint from snippet methods

### LINKED ISSUE #2136: api_endpoint in `snippets` methods can be reconstructed from the `location` argument.
As per the title, `api_endpoint` can be reconstructed  from the `location` argument. This led to problems because if the model is deployed elsewhere the requests fail. Additionally, the `api_endpoint` meaning is a bit opaque since a user could assume that

msteiner-google:
Fixes #2136 🦕


matthew29tang:
Closing due to inactivity.

PR #2397, opened 2023-08-01
feat: adding back Python 3.7 unit test support

nayaknishant:
Reverting removal of Python 3.7 unit test support (https://github.com/googleapis/python-aiplatform/commit/cb904d772abd0836f67b1034074395ce7a032d66).


yinghsienwu (approved):

yinghsienwu:
Try to merge this one after pull/2401 is merged.

.pre-commit-config.yaml
```diff
@@ -26,6 +26,6 @@ repos:
     hooks:
     - id: black
```
yinghsienwu, line 27: Need to keep it as 6.1.0 to avoid lint errors for new flake8

.pre-commit-config.yaml
```diff
@@ -26,6 +26,6 @@ repos:
...
 -   repo: https://github.com/pycqa/flake8
-    rev: 6.1.0
+    rev: 3.9.2
     hooks:
```
yinghsienwu, line 30: Need to keep it as 6.1.0, also I checked it should be compatible w/ python 3.7.

noxfile.py
```diff
@@ -25,14 +25,13 @@
 
 import nox
 
-FLAKE8_VERSION = "flake8==6.1.0"
```
yinghsienwu, line 28: keep this

noxfile.py
```diff
@@ -86,7 +85,7 @@ def lint(session):
     Returns a failure if the linters find linting errors or sufficiently
     serious code quality issues.
     """
-    session.install(FLAKE8_VERSION, BLACK_VERSION)
```
yinghsienwu, line 89: also this. So we pin on 6.1.0.

The source template was updated as well.
https://github.com/googleapis/synthtool/pull/1842

sasha-gitg (approved):

.pre-commit-config.yaml
nayaknishant: Fixed, all these changes were done automatically with owlbot. Updated to new version of owlbot to sync with synthtool so should be ok now.

noxfile.py
nayaknishant: Fixed, all these changes were done automatically with owlbot. Updated to new version of owlbot to sync with synthtool so should be ok now.

sasha-gitg (approved):

PR #3186, opened 2024-01-11
feat: LLM - Add support for "seed" request parameter

### LINKED ISSUE #3064: Support missing request parameters for TextGenerationModel.predict
Hi,

I am currently integrating PaLM-2 into [CRFM HELM](https://github.com/stanford-crfm/helm).
I noticed that some parameters are supported in an HTTP request (see [link](https://cloud.google.com/vertex-ai/docs/generative-ai/model-reference/text)) such as `echo`, `frequence_penaly` and `presence_penalty` but not in this Python SDK.

Would it be possible to add them please?
Thanks!

meken:
Fixes #3064 🦕

I'm well aware that some of the older models (mainly @001 versions) don't support the `seed` parameter, but that also applies to `presence_penalty` parameter which has been implemented in a similar fashion. In both cases the API endpoint will generate an error if the parameter is not supported by the specific version of the model.


tekumara:
Thanks @lozhn ! Out of interest, did you notice it had an effect on the text generations?

lozhn:
yes, `seed` became a hyperparameter for us

Ben-Epstein:
is there a reason we can't merge this? Being able to specify the seed is very important

Ben-Epstein:
@lozhn did you notice a massive performance degradation with the seed? It's unusably slow when the seed it set. Also, i found that it's not actually guaranteeing determinism...

```
import vertexai
from vertexai.language_models import TextGenerationModel

model = TextGenerationModel.from_pretrained("text-bison@002")
prompt = "tell me a story and make it very creative"

def predict(prompt, seed=None):
    kwargs = {"maxDecodeSteps": 1024, "temperature": 0.0, "topP": 1, "topK": 40}
    if seed is not None:
        kwargs["seed"] = seed
    instance = {"content": prompt}
    raw_pred = model._endpoint.predict(instances=[instance], parameters=kwargs)
    response = vertexai.language_models._language_models._parse_text_generation_model_multi_candidate_response(raw_pred)
    return response.text.strip()

%time predict(prompt, seed=None);  # Wall time: 3.34 s
%time predict(prompt, seed=42);    # Wall time: 6.95 s
```

Ark-kun:
Could you please split out `echo`?

I think that `seed` is important enough to support, but I'm not sure we see a scenario where `echo` is needed.

meken:
@Ark-kun I've now removed the `echo` parameter so the PR only includes `seed`. IMHO, the SDK should be as complete as possible, since the API supports that parameter, we should've considered it too. Anyway, I need the `seed` a little too hard to quarrel about `echo` :smile:

Ark-kun (approved):

Ark-kun (approved):

Ark-kun (approved):

Ark-kun (approved):

sasha-gitg (approved):

Ark-kun:
This has finally been merged and is released in 1.53.

Thank you for your great patience, everyone.

neta-b:
Couldn't make gemini 1.5 pro reproducible with the seed param. Should it work?

PR #3226, opened 2024-01-24
fix: add Content to vertex-ai ChatSession.send_message arg type

nicoloboschi:
in the vertex-ai package, the chat model allows Content to be used as parameter. LlamaIndex and LangChain integrations use the Content class. There are no tests and type hints is wrong, although the code already handles it perfectly

### DIFF
--- a/tests/system/vertexai/test_generative_models.py
+++ b/tests/system/vertexai/test_generative_models.py
@@ -137,6 +137,7 @@ class TestGenerativeModels(e2e_base.TestEndToEnd):
     # Chat
 
     def test_send_message_from_text(self):
+
         model = generative_models.GenerativeModel("gemini-pro")
         chat = model.start_chat()
         response1 = chat.send_message("I really like fantasy books.")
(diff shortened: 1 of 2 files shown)


Ark-kun:
Hello. Thank you for your contribution.

>in the vertex-ai package, the chat model allows Content to be used as parameter.

This seems to be a mistake that has creeped in. The type declaration is correct. The `ChatSession.send_message` is not supposed to accept `Content`. The reason is that unlike the more raw `generate_content`, `ChatSession` controls the integrity of the chat session. One part of that is the need to control the role. `ChatSession` cannot allow the user to control the `role` since that can lead to problematic history that is not accepted by the model. `Content` is basically `parts`+`role`. So the user should provide `parts`.

If for some reason the caller has `Content`, they should just pass `content.parts` to `send_message`.
Would that work for you?

nicoloboschi:
> Hello. Thank you for your contribution.
> 
> > in the vertex-ai package, the chat model allows Content to be used as parameter.
> 
> This seems to be a mistake that has creeped in. The type declaration is correct. The `ChatSession.send_message` is not supposed to accept `Content`. The reason is that unlike the more raw `generate_content`, `ChatSession` controls the integrity of the chat session. One part of that is the need to control the role. `ChatSession` cannot allow the user to control the `role` since that can lead to problematic history that is not accepted by the model. `Content` is basically `parts`+`role`. So the user should provide `parts`.
> 
> If for some reason the caller has `Content`, they should just pass `content.parts` to `send_message`. Would that work for you?

It makes sense. However I believe something has to be done in any case, either to:
1. deprecate and add a warning in the log in case of Content being passed
2. accept the Content in the types. 

Given your explanation, it makes more sense to do 1. Raising an error would break a lot of integrations (llama_index, LangChain AFAIK) 

I can change the PR accordingly

matthew29tang:

PR #3506, opened 2024-03-27
Fix BigFrames Dependencies

tabbyl21:
If a user has installed a bigframes version 1.0.0, they will be unable to leverage the vertex ai langchain extesnion and any other lib that uses this package as a dependency. So we 
1) For texting logic pin a version of BigFrames less than 1.0.0
2) In code that imports bigframes.dataframe verify user has installed a version of bigframes < 1.0.0 .
This is not intended as a permanent solution as bigframes 1.0.0 has been released, this library needs to be further updated so that it uses the new Dataframes module from bigframes. 

### DIFF
--- a/tests/unit/vertexai/test_any_serializer.py
+++ b/tests/unit/vertexai/test_any_serializer.py
@@ -47,6 +47,8 @@ except ImportError:
 
 try:
     import bigframes as bf
+    if bf.__version__ >= "1.0.0":
+        raise ImportError("BigFrames version is greater than or equal to 1.0.0. Please install a version less than 1.0.0.")
 except ImportError:
     bf = None
 
--- a/vertexai/preview/_workflow/executor/training.py
+++ b/vertexai/preview/_workflow/executor/training.py
@@ -53,14 +53,8 @@ try:
 except ImportError:
     import importlib_metadata
 
-try:
-    import bigframes as bf
-    from bigframes.dataframe import DataFrame
-
-    BigframesData = DataFrame
-except ImportError:
-    bf = None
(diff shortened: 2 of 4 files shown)


vertexai/preview/_workflow/executor/training.py
```diff
@@ -53,14 +53,8 @@
...
-try:
-    import bigframes as bf
-    from bigframes.dataframe import DataFrame
-
-    BigframesData = DataFrame
-except ImportError:
-    bf = None
-    BigframesData = Any
+bf = None
+BigframesData = Any
```
kurtisvg, line 57: Can you remove this altogether? I didn't see it being used in the file

tests/unit/vertexai/test_any_serializer.py
```diff
@@ -47,6 +47,8 @@
 
 try:
     import bigframes as bf
+    if bf.__version__ >= "1.0.0":
+        raise ImportError("BigFrames version is greater than or equal to 1.0.0. Please install a version less than 1.0.0.")
 except ImportError:
```
kurtisvg, line 52: the error on line 51 will be caught by 52 without logging

tabbyl21: if bf is set to None it will throw an error here: https://github.com/googleapis/python-aiplatform/blob/67a10e98bd2a55a871bcdfbbc608b96a89d546aa/tests/unit/vertexai/test_any_serializer.py#L313

tabbyl21: added a print statement so its verbose.

PR #4070, opened 2024-07-09
fix(deps): Require proto-plus 1.22.3

parthea:
Update the minimum version of `proto-plus` to 1.22.3 to match [gapic-generator-python](https://github.com/googleapis/gapic-generator-python/blob/6fc3d22b697ed14ef2daa18bbd0964026caed850/gapic/templates/setup.py.j2#L37) similar to https://github.com/googleapis/gapic-generator-python/pull/1863. This will prevent issues like https://github.com/googleapis/proto-plus-python/issues/469.


sasha-gitg (approved):

jaycee-li:
Merged in 4131e6583799d16b5032fecd73d4539fb05f0cd1

PR #4156, opened 2024-08-05
Draft: add cpu and memory config

iliya-malecki:
This is a draft, as writing a code for this tiny change is much easier than explaining what i would like to change.

The main motivator of this one is being able to test / observe the OOM behavior of LocalModel-s when deploying locally with `LocalModel.deploy_to_local_endpoint`. Thus, ive added some flags the user can set to control `docker run` behavior. I have considered:
- passing arbitrary `**kwargs` to `docker run` as flags via the same mechanism im using in this PR
- passing a config object to `docker run` as flags via the same mechanism
- selecting the most useful subset of flags to expose to `deploy_to_local_endpoint` 
Ive chosen the third approach as it is the least sloppy and/or bloated and ive felt like this was the authors intention, even though it is the least flexible. Feel free to choose another approach

### DIFF
--- a/google/cloud/aiplatform/docker_utils/run.py
+++ b/google/cloud/aiplatform/docker_utils/run.py
@@ -19,7 +19,7 @@ import logging
 import os
 from pathlib import Path
 import re
-from typing import Dict, List, Optional, Sequence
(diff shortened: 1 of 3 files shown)


matthew29tang:

PR #4558, opened 2024-10-19
fix: Remove check on serving_container_image_uri when artifact_uri and local_model is given

rragundez:
Remove bug that forces `serving_container_image_uri` and `artifact_uri` to be given. While `artifact_uri` might be a local path which doesn't necessarily need `serving_container_image_uri` to be set if `local_model` is set.

At the moment there is a bug in the code where if `local_model` is given and `artifact_uri` is given as a local path it will raise an exception because of this line: `PREBUILT_IMAGE_RE = "(us|europe|asia)-docker.pkg.dev/vertex-ai/prediction/"`

https://github.com/googleapis/python-aiplatform/blob/3e7bf819d95167502be6b9f04c7b92e8415e5a68/google/cloud/aiplatform/models.py#L4989

which is doing a regex over the variable `serving_container_image_uri ` which does not have to be given since `local_model` was. 

The actual exception is:  
```python
Traceback (most recent call last):
  File "/Users/rodrigo.agundez/repos/dds-platform-utils/build/source/script.py", line 34, in <module>
    model = aiplatform.Model.upload(
  File "/Users/rodrigo.agundez/Library/Caches/pypoetry/virtualenvs/dyson-nova-vEH0SleR-py3.10/lib/python3.10/site-packages/google/cloud/aiplatform/base.py", line 863, in wrapper
    return method(*args, **kwargs)
  File "/Users/rodrigo.agundez/Library/Caches/pypoetry/virtualenvs/dyson-nova-vEH0SleR-py3.10/lib/python3.10/site-packages/google/cloud/aiplatform/models.py", line 4989, in upload
    if re.match(PREBUILT_IMAGE_RE, serving_container_image_uri):
  File "/Users/rodrigo.agundez/.pyenv/versions/3.10.12/lib/python3.10/re.py", line 190, in match
    return _compile(pattern, flags).match(string)
TypeError: expected string or bytes-like object
```

Looking at your code, I don't think this regex check is needed since there is no concept of an pre-built image, while it is useful to have the combination of `local_model`  and `artifact_uri` to use models that might be on disk while using a custom model or endpoint later on.


rragundez:
not sure why the CI is failing with this:

image

rragundez:
@parthea  @sasha-gitg @jaycee-li HI, could you have a look, I think is quite a simple one and it is currently holding up our deployments, so I would like to avoid my team to have to hack its way and solve it from the source.

google/cloud/aiplatform/models.py
```diff
@@ -4985,20 +4985,18 @@ def upload(
...
-                    )
-                if not any(
-                    (model_dir / file_name).exists()
-                    for file_name in _SUPPORTED_MODEL_FILE_NAMES
-                ):
-                    raise ValueError(
-                        "artifact_uri directory does not contain any supported model files. "
-                        f"When using a prebuilt serving image, the upload method only supports the following model files: '{_SUPPORTED_MODEL_FILE_NAMES}'"
-                    )
+            if not model_dir.is_dir():
```
jaycee-li, line 4988: Your PR changes the logic of the `model_dir` validation. If you only concern about the error raised when `serving_container_image_uri` is not provided, you can do something like this rather than remove this check.
```
if serving_container_image_uri and re.match(PREBUILT_IMAGE_RE, serving_container_image_uri):
```

jaycee-li (changes requested):

rragundez:
Hi @jaycee-li thanks for the help! I made the change as you requested and indeed this will help solve the behaviour I am seeing.

Moreover, I don't think this really fixes the core issue which I think it is that on this line
https://github.com/googleapis/python-aiplatform/blob/main/google/cloud/aiplatform/models.py#L4991
there is a check `if artifact_uri and not artifact_uri.startswith("gs://")` and the checks about being a dictionary and such happens only if `serving_container_image_uri` is given, but I do not see why it wouldn't checked anyway if in local mode or not.

Like I mentioned this PR already addressed the specific problem I am facing. So thanks for the review!

jaycee-li (approved):

jaycee-li (approved):

google/cloud/aiplatform/models.py
```diff
@@ -5020,7 +5020,7 @@ def upload(
             if not model_dir.exists():
                 raise ValueError(f"artifact_uri path does not exist: '{artifact_uri}'")
             PREBUILT_IMAGE_RE = "(us|europe|asia)-docker.pkg.dev/vertex-ai/prediction/"
-            if re.match(PREBUILT_IMAGE_RE, serving_container_image_uri):
+            if serving_container_image_uri and re.match(PREBUILT_IMAGE_RE, serving_container_image_uri):
```
jaycee-li, line 5023: Hi @rragundez , could you please fix the linting issue 

image

jaycee-li (changes requested):

google/cloud/aiplatform/models.py
rragundez: Done

jaycee-li (approved):

jaycee-li:
PR submitted https://github.com/googleapis/python-aiplatform/commit/ccd89b41229516c561aee686497ad5af95c844ed. Close this one.

jaycee-li:
@rragundez We release every Tuesday, so the next release will be 11/19 assuming everything goes smoothly.

rragundez:
OK got it thanks. Can I subscribe as a viewer to some page or so to get emails when a new release is made?

PR #4668, opened 2024-11-18
feat: expose HarmBlockMethod to external callers

telpirion:
This enum is required for the SafetySettings() class to allow users to toggle harmful content blocking.

Without exposing this type, the following code won't work:

```
generative_models.SafetySetting(
        category=generative_models.HarmCategory.HARM_CATEGORY_DANGEROUS_CONTENT,
        method=generative_models.HarmBlockMethod.SEVERITY,
        threshold=generative_models.HarmBlockThreshold.BLOCK_ONLY_HIGH,
    ),
```

Found [here](https://cloud.google.com/vertex-ai/generative-ai/docs/multimodal/configure-safety-filters#how_to_configure_safety_filters).


Ark-kun:
It should be exposed via `SafetySettings.HarmBlockMethod`. Is it not?

telpirion:
> It should be exposed via `SafetySettings.HarmBlockMethod`. Is it not?

No, it is not.

Ark-kun:
>Question: why is HarmBlockThreshold and HarmCategory exposed directly in the generative_models module but not HarmBlockThreshold?

Historical reasons. Initially we tried to use `safety_settings: dict[HarmCategory, HarmBlockThreshold]` instead of `safety_settings: list[SafetySetting]`. So `SafetySetting` was not exposed and `HarmBlockThreshold` and `HarmCategory` we exposed directly. Then `HarmBlockMethod` was added and we had to expose `SafetySetting` and switch to `safety_settings: list[SafetySetting]`. Now there are 4 classes related to safety. To avoid the `generative_models` module crowding we decided not to expose `HarmBlockMethod` and any subsequent classes, choosing to group them under `SafetySetting`.

Maybe we're wrong and we should just expose `HarmBlockMethod` the same way...

telpirion:
Okay, makes sense. If you decide not to expose `HarmBlockMethod` in this module, feel free to delete this PR. Thanks!

matthew29tang:

PR #6074, opened 2025-10-31
fix: added option to set fps in video metadata according to the documentation

### LINKED ISSUE #6073: fps option is missing in VideoMetadata
FPS option is missed in `class VideoMetadata(proto.Message)` in types/content.py

It should exist according to both docs (v1 and v1beta1):
https://cloud.google.com/vertex-ai/generative-ai/docs/reference/rest/v1/Content
https://cloud.google.com/vertex-ai/generative-ai/docs/reference/rest/v1beta1/Content

And also is existent in VideoMetadata in `content.proto`:
https://github.com/googleapis/googleapis/blob/97763d6efb2d35474e0363c254ee6994837a08fa/google/ai/generativelanguage/v1/content.proto#L107
as:
```
message VideoMetadata {
  // Optional. The start offset of the video.
  google.protobuf.Duration start_offset = 1
      [(google.api.field_behavior) = OPTIONAL];

  // Optional. The end offset of the video.
  google.protobuf.Duration end_offset = 2
      [(google.api.field_behavior) = OPTIONAL];

  // Optional. The frame rate of the video sent to the model. If not specified,
  // the default value will be 1.0. The fps range is (0.0, 24.0].
  double fps = 3 [(google.api.field_behavior) = OPTIONAL];
}
```

philipshurpik:
FPS option is missed in class VideoMetadata(proto.Message) in types/content.py
both in v1 and v1beta1

It should exist according to both docs (v1 and v1beta1):
https://cloud.google.com/vertex-ai/generative-ai/docs/reference/rest/v1/Content
https://cloud.google.com/vertex-ai/generative-ai/docs/reference/rest/v1beta1/Content

And also is existent in VideoMetadata in content.proto:
https://github.com/googleapis/googleapis/blob/97763d6efb2d35474e0363c254ee6994837a08fa/google/ai/generativelanguage/v1/content.proto#L107

I implemented it's support according to proto file and documentation.
Tested locally that library version with this changes is actually working by using less 'input_tokens'.
Also added using of fps field in few tests to the existing video_metadata object 

Fixes #6073  🦕


matthew29tang:
These files are auto-generated so we can't modify them. If you want to use the fps field, you should use the new Python GenAI SDK: https://github.com/googleapis/python-genai

PR #6119, opened 2025-11-13
Fix: Missing embedding_metadata in MatchNeighbor returned by find_neighbors in Vertex AI Matching Engine.

shubham-maindola:
## Summary
The `aiplatform.MatchingEngineIndexEndpoint.find_neighbors()` method returns results as 
`List[List[aiplatform.matching_engine.matching_engine_index_endpoint.MatchNeighbor]]`.  
However, the current `MatchNeighbor` dataclass does not expose the `embedding_metadata` field, 
even when `return_full_datapoint=True` and the API response includes the `embeddingMetadata` attribute.

## Issue
When `return_full_datapoint=True` is set, the API correctly returns the full datapoint with 
`embeddingMetadata`, but this metadata is silently dropped because the 
`MatchNeighbor` class does not have a corresponding attribute to capture it.

## Fix
The `aiplatform_v1beta1.index.IndexDatapoint` class already supports `embedding_metadata`.  
This PR adds an `embedding_metadata` field to the `MatchNeighbor` dataclass and populates it 
within the `MatchNeighbor.from_index_datapoint()` method.

## Testing
Verified the fix locally by querying an index with `return_full_datapoint=True`.  
The `embedding_metadata` field now appears correctly in the returned `MatchNeighbor` objects 
and matches the data returned directly from the API.

## Impact
This change enables users to access `embedding_metadata` when using 
`find_neighbors()` through the Python SDK, bringing it to parity with the REST API behavior.

### DIFF
--- a/google/cloud/aiplatform/matching_engine/matching_engine_index_endpoint.py
+++ b/google/cloud/aiplatform/matching_engine/matching_engine_index_endpoint.py
@@ -16,7 +16,7 @@
 #
 
 from dataclasses import dataclass, field
-from typing import Dict, List, Optional, Sequence, Tuple, Union
+from typing import Dict, List, Optional, Sequence, Tuple, Union, Any
 
 from google.auth import credentials as auth_credentials
 from google.cloud.aiplatform import base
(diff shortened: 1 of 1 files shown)


shubham-maindola:
I used a different email account for this my bad.

PR #5774, opened 2025-08-31
fix: make `async_stream_query` fully async

uncured7036:
fix: make `async_stream_query` fully async.

Fixes #5773

PR #6295, opened 2026-01-26
fix: make stream_query_reasoning_engine async in AsyncClient

### LINKED ISSUE #5773: Bug: `async_stream_query` not async
#### Environment details

  - OS type and version: Linux 6.6.87.2-microsoft-standard-WSL2
  - Python version: Python 3.11.9
  - pip version: pip 25.2
  - `google-cloud-aiplatform` version: 1.111.0

#### Steps to reproduce

  1. Deploy a simple adk agent.
  2. Run multiple async functions to call the agent's `async_stream_query` method.

#### Code example

```python
async def run_query(app, uid):
    prompt = 'some prompts have delayed responses'
    session = await app.async_create_session(user_id=uid)
    async for event in app.async_stream_query(
        user_id=uid,
        session_id=session['id'],
        message=prompt,
    ):
        print(event)
    await app.async_delete_session(user_id=uid, session_id=session['id'])

async def run():
    agent = agent_engines.get(AGENT_ID)
    await asyncio.gather(
        run_query(agent, 'u1'),
        run_query(agent, 'u2'),
        run_query(agent, 'u3'),
    )

if __name__ == '__main__':
    asyncio.run(run())
```

#### Result

The requests are not sent concurrently to the Vertex AI Agent Engine.

Image

olibook:
- Add async keyword to stream_query_reasoning_engine method
- Add await to rpc call in async client
- Use execution_async_client instead of execution_api_client
- Use async for loop to iterate over response stream

Fixes #5773

This is a rebase of https://github.com/googleapis/python-aiplatform/pull/5774

### DIFF
--- a/google/cloud/aiplatform_v1/services/reasoning_engine_execution_service/async_client.py
+++ b/google/cloud/aiplatform_v1/services/reasoning_engine_execution_service/async_client.py
@@ -408,7 +408,7 @@ class ReasoningEngineExecutionServiceAsyncClient:
         # Done; return the response.
         return response
 
-    def stream_query_reasoning_engine(
+    async def stream_query_reasoning_engine(
         self,
         request: Optional[
             Union[
@@ -541,7 +541,7 @@ class ReasoningEngineExecutionServiceAsyncClient:
         self._client._validate_universe_domain()
 
         # Send the request.
-        response = rpc(
+        response = await rpc(
             request,
             retry=retry,
             timeout=timeout,
--- a/vertexai/agent_engines/_agent_engines.py
+++ b/vertexai/agent_engines/_agent_engines.py
@@ -1643,14 +1643,14 @@ def _wrap_async_stream_query_operation(
     """
 
     async def _method(self, **kwargs) -> AsyncIterable[Any]:
-        response = self.execution_api_client.stream_query_reasoning_engine(
+        response = await self.execution_async_client.stream_query_reasoning_engine(
             request=aip_types.StreamQueryReasoningEngineRequest(
                 name=self.resource_name,
                 input=kwargs,
                 class_method=method_name,
             ),
         )
-        for chunk in response:
+        async for chunk in response:
             for parsed_json in _utils.yield_parsed_json(chunk):
                 if parsed_json is not None:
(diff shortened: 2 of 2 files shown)


olibook:
Hello @sararob  ! do you mind having a look at action workflows ? requires an approval from a maintainer.

Cheers

NameXRipple:
Hi @yeesian ,

Could you approve this PR ?

Thanks

spectakural:
Thanks for getting to this first — your change to `_wrap_async_stream_query_operation` matches what I independently arrived at while looking into #7136.

Two things that I think are keeping it from landing, offered in case they help:

**1. The existing tests break under this change.**

The three `async_stream_query` tests in `tests/unit/vertex_langchain/test_agent_engines.py` patch the synchronous `ReasoningEngineExecutionServiceClient`. Once the source uses `execution_async_client`, that mock is no longer in the path and those tests fail. I confirmed this locally by applying only the source half of your change. They need a fixture patching `ReasoningEngineExecutionServiceAsyncClient` instead.

This wouldn't have shown up here — the checks that ran on this PR are the CLA, snippet-bot and commit-lint bots, none of which execute the test suite. It's also worth noting that these particular tests live under `tests/unit/vertex_langchain`, which `nox -s unit` explicitly ignores; the session that covers them is `nox -s unit_langchain`.

**2. The `async_client.py` change may not be necessary.**

`google/cloud/aiplatform_v1/.../async_client.py` is GAPIC-generated and gets regenerated from the protos, so hand edits there tend to be reverted. As far as I can tell it also isn't needed: `api_core`'s `grpc_helpers_async._wrap_stream_errors` already wraps the RPC in a coroutine function, so the current `def` + `response = rpc(...)` returns an awaitable that resolves to an async iterable — which is exactly what the method's own annotation and docstring describe. The SDK-level `await` + `async for` works against the generated client as it stands today.

I have a branch with the test-fixture updates plus a regression test that asserts `asyncio.wait_for` can cancel a slow stream. Happy to push those onto your branch instead of opening a competing PR if you'd rather keep this one — just say the word. Otherwise I'll open one referencing #7136 and link it here.

PR #6399, opened 2026-03-09
fix: remove predictable GCS bucket names to prevent bucket squatting

KevinZhao:
## Summary

The fix for CVE-2026-2473 (GHSA-wh2j-26j7-9728) in v1.133.0 patched `metadata/_models.py` but missed two other locations in `utils/gcs_utils.py` that construct GCS bucket names from predictable inputs (project ID + region):

- `stage_local_data_in_gcs()` line 206: `"{project}-vertex-staging-{location}"`
- `generate_gcs_directory_for_pipeline_artifacts()` line 254: `"{project}-vertex-pipelines-{location}"`

An attacker who knows a victim's project ID and region can pre-register these globally-unique bucket names in their own GCP project and configure public write access. When the victim's SDK auto-generates the same predictable name, `Bucket.exists()` returns True for the attacker's bucket, and the SDK silently uploads model artifacts, training data, and pipeline outputs to attacker-controlled storage.

## Changes

Apply the same fix pattern as CVE-2026-2473:

- **`stage_local_data_in_gcs()`**: Require explicit `staging_gcs_dir` or `aiplatform.init(staging_bucket=...)`. Raise `RuntimeError` if neither is provided, instead of auto-generating a predictable bucket name.
- **`generate_gcs_directory_for_pipeline_artifacts()`**: Use `staging_bucket` from global config. Raise `RuntimeError` if not set. Add `validate_gcs_path()` to ensure proper `gs://` prefix.
- Updated docstrings with `Raises` sections and deprecated parameter notes.
- Updated unit tests to match new behavior.

## Affected entry points

- `Model.upload()` when `staging_bucket` is not provided (calls `stage_local_data_in_gcs`)
- `PipelineJob()` when `pipeline_root` is not provided (calls `generate_gcs_directory_for_pipeline_artifacts`)

## Test plan

- [x] Updated `test_generate_gcs_directory_for_pipeline_artifacts` — tests both success path (with staging_bucket) and RuntimeError path
- [x] Updated `test_create_gcs_bucket_for_pipeline_artifacts_if_it_does_not_exist` — passes explicit `output_artifacts_gcs_dir`
- [ ] Verify no regressions in `vertexai/` callers (distillation, language models)

### DIFF
--- a/google/cloud/aiplatform/utils/gcs_utils.py
+++ b/google/cloud/aiplatform/utils/gcs_utils.py
@@ -185,6 +185,8 @@ def stage_local_data_in_gcs(
 
     Raises:
         RuntimeError: When source_path does not exist.
+        RuntimeError: When staging_gcs_dir is not provided and staging_bucket
+            is not configured via aiplatform.init().
         GoogleCloudError: When the upload process fails.
     """
     data_path_obj = pathlib.Path(data_path)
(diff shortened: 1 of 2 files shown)


matthew29tang:
Bucket squatting should be mitigated from https://github.com/googleapis/python-aiplatform/pull/6582

PR #6599, opened 2026-04-16
fix: update _evals_common to be compatible with litellm >=1.83.0

quad2524:
Summary
This PR updates the litellm dependency to version 1.83.0 or higher. This upgrade is necessary to bypass known security vulnerabilities present in version 1.82.7.

Because litellm introduced changes to how models and providers are validated, I have also updated the internal utility functions and associated tests to maintain compatibility.

Changes
Dependency Update: Bumped litellm version requirement in setup files.

Core Logic (_evals_common): Updated _is_litellm_model to utilize the newer get_llm_provider pattern. This ensures we accurately validate model strings against LiteLLM’s supported provider list.

Test Suite:

Refactored mocks to account for the new return signature of litellm.get_llm_provider, which now includes additional metadata (model, provider, etc.).

Updated get_valid_models mocks to ensure consistent behavior during unit testing.

Fixes #6598 🦕

PR #6608, opened 2026-04-17
fix(deps): repin litellm above 1.83.0

### LINKED ISSUE #6598: Please repin litellm to >1.83
I'm getting security alerts for "LiteLLM: Authentication bypass via OIDC userinfo cache key collision". I'd like to ugprade however the google-cloud-aiplatform is pinning litellm to <1.82.6 (before the supply chain attack).  Now that it's resolved, please repin it to a version afterwards so we can pick up the newer fixes.

Or just support the openai api directly and remove litellm

VenkataAnilKumar:
Fixes #6598

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -181,8 +181,8 @@ evaluation_extra_require = [
     "jsonschema",
     "ruamel.yaml",
     "pyyaml",
-    "litellm>=1.75.5, <=1.82.6",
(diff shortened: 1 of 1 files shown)


VenkataAnilKumar:
Closing this as duplicate of #6599 to keep one active track for #6598.

PR #6624, opened 2026-04-21
fix: align app_name with Cloud ML Job ID in deployed environments

EaCognitive:
## Problem
When an `AdkApp` is deployed to Vertex AI Agent Engine, there is often a mismatch between the local `app.name` and the Resource ID assigned by the backend. The SDK currently uses the local name for artifact lookups, leading to 404 errors because the deployment infrastructure expects the Resource ID.

## Solution
Updated `AdkApp._app_name()` to check for the `CLOUD_ML_JOB_ID` environment variable. If present (indicating a deployed environment), the method returns this ID, ensuring consistency with the backend's expectations for artifact storage and lookup.

## Changes
- `vertexai/agent_engines/templates/adk.py`: Added environment variable check to `_app_name()`.

## Verification
- Verified that the logic correctly prioritizes the environment variable over the local name when available.


EaCognitive:
Updated with current `main` and adjusted the app-name regression test for the upstream ADK version check. Local validation: `uv run ... pytest tests/unit/vertex_adk/test_agent_engine_templates_adk.py -k 'test_default_app_name_uses_agent_engine_id or test_initialization_keeps_explicit_app_name or test_default_app_name_uses_fallback_without_agent_engine_id'` -> 3 passed. Requesting maintainer review when available.

PR #7034, opened 2026-07-29
feat: support authorizations field in async_stream_query

neville-m-exa:
## Summary

Add `authorizations` support to `async_stream_query`, matching the existing behavior in `streaming_agent_run_with_events`.

## Problem

When `authorizations` is included in the input payload of an `async_stream_query` call to Agent Engine, it flows through `**kwargs` to `Runner.run_async()` which doesn't accept it:

```
TypeError: Runner.run_async() got an unexpected keyword argument 'authorizations'
```

The `authorizations` field currently only works via `streaming_agent_run_with_events` (which parses a `request_json` string into `_StreamRunRequest`). However, `streaming_agent_run_with_events` is documented as "primarily meant for invocation from AgentSpace" and has a different input/output contract, making it unsuitable for standard API callers that need to forward user credentials to agent tools.

## Solution

- Add `authorizations` as a named parameter to `async_stream_query`
- Convert each authorization's `access_token` into a `state_delta` entry (same logic as `streaming_agent_run_with_events` uses via `_StreamRunRequest`)
- Pass `state_delta` to `Runner.run_async()` (which already supports it)

This is a minimal, backward-compatible change — callers that don't pass `authorizations` see no difference (`state_delta=None` is a no-op).

## Use Case

Agents deployed on Agent Engine that need to call authenticated external APIs on behalf of the user. The caller (e.g., a gateway service) extracts the user's JWT and forwards it via `authorizations`. The agent reads it from `session.state` and uses it as a Bearer token.

## Test Plan

- [x] Added `test_async_stream_query_with_authorizations` — verifies no error when authorizations is passed
- [x] Added `test_async_stream_query_authorizations_passed_as_state_delta` — verifies tokens are converted to state_delta
- [x] Added `test_async_stream_query_no_authorizations_passes_none_state_delta` — verifies no-op when absent

🤖 Generated with [Claude Code](https://claude.com/claude-code)


neville-m-exa:
@yeesian , @Tongzhou-Jiang can you please review.

PR #7072, opened 2026-08-14
fix: preserve zero max wait duration

### LINKED ISSUE #7067: CustomContainerTrainingJob.run drops max_wait_duration=0 instead of requesting indefinite DWS wait
Thanks for maintaining this library. `CustomContainerTrainingJob.run(max_wait_duration=0)` drops the value because the high-level training-job implementation uses a truthiness check. This prevents callers from requesting the documented indefinite wait for DWS Flex Start.

#### Environment details

- OS type and version: macOS
- Python version: 3.12
- `google-cloud-aiplatform` versions: reproduced in 1.148.1; source inspection confirms the same behavior in 1.150.0, 1.163.0, and current `main`

#### Steps to reproduce

1. Create a `CustomContainerTrainingJob` using any valid project, staging bucket, and container image.
2. Run it with Flex Start and `max_wait_duration=0`.
3. Inspect the scheduling inputs produced by `_prepare_training_task_inputs_and_output_dir` or the resulting `TrainingPipeline` request.

#### Code example

```python
job.run(
    scheduling_strategy=custom_job.Scheduling.Strategy.FLEX_START,
    max_wait_duration=0,
)
```

Observed scheduling value:

```text
max_wait_duration: None
```

Expected scheduling value:

```text
max_wait_duration: 0s
```

For comparison, a positive input such as `14400` is preserved as `14400s`.

#### Cause

The high-level training-job path converts the duration using a truthiness check, so zero follows the same branch as `None`:

https://github.com/googleapis/python-aiplatform/blob/v1.163.0/google/cloud/aiplatform/training_jobs.py#L1663-L1667

The generated API contract says that explicit zero means indefinite waiting, while omission defaults to 24 hours:

https://github.com/googleapis/python-aiplatform/blob/v1.163.0/google/cloud/aiplatform_v1/types/custom_job.py#L577-L582

`max_wait_duration` is a presence-aware `google.protobuf.Duration`, so explicit zero and an omitted field are distinct requests.

A previous fix correctly changed the lower-level `CustomJob` and hyperparameter-tuning paths to use `is not None`, explicitly noting that `0` is valid:

https://github.com/googleapis/python-aiplatform/commit/d9675fdf051233539f478187143f2833fd6e6af0

That commit did not update the `CustomContainerTrainingJob` path in `training_jobs.py`.

#### Suggested fix

Use an explicit `is not None` check in `_prepare_training_task_inputs_and_output_dir`, consistent with the lower-level implementation, and add tests covering omitted, zero, and positive durations.

#### Stack trace

No exception is raised. The problem is a silent request-semantics change: `0` is serialized as omission.

hugosmoreira:
## Summary

- preserve an explicit `max_wait_duration=0` when building custom-training scheduling inputs
- continue treating `None` as an omitted duration
- add regression coverage for omission, zero with Flex Start, zero without another scheduling option, and a positive duration

## Why

The documented value `0` is a sentinel that requests an indefinite Dynamic Workload Scheduler wait. Request preparation used truthiness checks, causing explicit zero to be converted to `None` and silently omitted from the outgoing training pipeline.

Using explicit `is not None` checks keeps the API distinction between omission and zero without changing positive-duration behavior.

Fixes #7067.

## Validation

- focused regression test: `4 passed, 236 deselected`
- complete `tests/unit/aiplatform/test_training_jobs.py` module: `240 passed`
- negative control: restoring the original truthiness checks makes both zero cases fail with `None != "0s"`
- Black 24.8.0 check passes for the changed production file
- Flake8 6.1.0 passes for both changed files
- `git diff --check` passes

A broader local Windows package run was not completed because the Python process encountered a native memory-access failure; no full-suite result is claimed. GitHub's Linux CI remains the broader verification environment.

### DIFF
--- a/google/cloud/aiplatform/training_jobs.py
+++ b/google/cloud/aiplatform/training_jobs.py
@@ -1659,10 +1659,12 @@ class _CustomTrainingJob(_TrainingJob):
             or restart_job_on_worker_restart
             or disable_retries
             or scheduling_strategy
-            or max_wait_duration
+            or max_wait_duration is not None
         ):
             timeout = f"{timeout}s" if timeout else None
-            max_wait_duration = f"{max_wait_duration}s" if max_wait_duration else None
+            max_wait_duration = (
+                f"{max_wait_duration}s" if max_wait_duration is not None else None
+            )
             scheduling = {
                 "timeout": timeout,
                 "restart_job_on_worker_restart": restart_job_on_worker_restart,
--- a/tests/unit/aiplatform/test_training_jobs.py
+++ b/tests/unit/aiplatform/test_training_jobs.py
@@ -5107,6 +5107,77 @@ class TestCustomContainerTrainingJob:
 
     @mock.patch.object(training_jobs, "_JOB_WAIT_TIME", 0.05)
     @mock.patch.object(training_jobs, "_LOG_WAIT_TIME", 0.05)
+    @pytest.mark.usefixtures("mock_pipeline_service_get_with_scheduling")
+    @pytest.mark.parametrize(
+        (
+            "scheduling_strategy",
+            "max_wait_duration",
+            "expected_max_wait_duration",
+        ),
+        [
+            pytest.param(
+                gca_custom_job.Scheduling.Strategy.FLEX_START,
+                None,
+                None,
+                id="omitted",
+            ),
+            pytest.param(
+                gca_custom_job.Scheduling.Strategy.FLEX_START,
+                0,
+                "0s",
+                id="zero-with-flex-start",
+            ),
+            pytest.param(
+                gca_custom_job.Scheduling.Strategy.FLEX_START,
+                _TEST_MAX_WAIT_DURATION,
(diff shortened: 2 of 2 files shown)


hugosmoreira:
I revalidated this on Python 3.13 in an isolated, offline WSL-backed Docker environment: all four regression cases and all 240 tests in the related module pass, both at the PR head and with the change applied to current main. Restoring the original code makes exactly the two zero cases fail. The installed-wheel regression, Black, changed-file lint and package builds also pass.

I am keeping the PR in draft while resolving the wider gates. The current `[testing]` dependency resolution selects Protobuf 5.29.0rc3 because of the Protobuf cap and exclusions in the resolved databricks-sdk; I did not install that set or override the constraints. Full Flake8 under Python 3.13 also reports the same five findings on the original code, PR head and current main.

Could you point me to the expected stable CI dependency set and lint environment, and help approve the pending workflows? I would like to finish those checks without broadening this fix into unrelated changes.

